PR #24, opened 2017-11-27
Modinv 26 marc to folio

aclseelen:
*DON'T MERGE!* _I pushed the feature branch to folio-org and will create a new pull request when the ticket is ready enough to be implemented. This PR is only for the purpose of reviewing._

Initial commit for ticket MODINV-26. Includes some minimal MarcJson to Folio instance POJO mapping with corresponding tests.

NOTE: merging this pull request shouldn't harm (only new classes involved and tests that pass and should pass SonarLint), but the ticket is not yet ready. In case someone else instead of me will continue working on this ticket, I'll write a list down here that I will also put in the ticket in Jira.

Still to be done:

1) after retrieving exact specifications of all Marc structures/fields/data to be supported the test class and sample data should be extended accordingly and code should be written to make the test pass.

2) the Instance class (`org.folio.inventory.domain.Instance`) probably has to be extended to be able to handle all the Marc data (fields et cetera) that is considered to be supported.

3) a method that converts a Marc file to a Marc JsonObject (`io.vertx.core.json.JsonObject`) has to be added to complete the ticket.


src/main/java/org/folio/inventory/support/MarcParser.java
```diff
@@ -0,0 +1,87 @@
...
+    String title = "";
+    String source = "";
+    String instanceTypeId = "";
+
+    if (marcEntry.containsKey(FIELDS) && marcEntry.getValue(FIELDS).getClass() == JsonArray.class) {
+      JsonArray fieldArray = marcEntry.getJsonArray(FIELDS);
+      List<JsonObject> fields = this.extractFromJsonArray(fieldArray);
+      title = this.extractTitle(fields);
+    } else {
+      LOGGER.debug("JsonObject [marcEntry]: no JsonArray found at key 'fields'");
```
julianladisch: throw exception, don't create Instance with empty strings

PR #39, opened 2018-01-18
MODINV-44: Extend Instances to include MARC payload.

julianladisch:


julianladisch:
We do not merge this code because we want to come up with a better solution in https://issues.folio.org/browse/MODINVSTOR-26

PR #93, opened 2019-01-08
MODINV-100: Use Alpine docker container.

julianladisch:


julianladisch:
Embedded postgres doesn't run under Alpine. However mod-inventory doesn't use embedded postgres (no RMB PostgresClient, no RMB postgres-runner).

PR #104, opened 2019-03-22
Prohibit updating item status

dmytrotur:
## Purpose
Resolves: MODINV-109
Restrict updating item from UI to avoid status change
## Approach
Set old status when item is updated
## Links
https://issues.folio.org/browse/MODINV-109


marcjohnson-kint:
I'm closing this based upon the conversation on the issue around complications with this approach

PR #118, opened 2019-06-11
MODINV-121 Add configuration with names of blocked fields for Instance editing

Igor-Gorchakov:
* Created a new endpoint GET /inventory/config/instances/blockedFields
* Response returns 200 code and JSON object with an array of fields
* Covered with API tests


Igor-Gorchakov:
Hi @marc-johnson , please take a look, I'm waiting for your review

ramls/blocked_fields.json
```diff
@@ -0,0 +1,18 @@
+{
```
marcjohnson-kint: Why does this filename use an underscore rather than a hyphen? 

Maybe it is because two very old schema (that I created) use underscores. These should really be renamed (though they have both been deprecated for well over a year)

descriptors/ModuleDescriptor-template.json
```diff
@@ -2,6 +2,18 @@
   "id": "${artifactId}-${version}",
   "name": "Inventory Module",
   "provides": [
+    {
+      "id": "inventory-config",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/inventory/config/instances/blockedFields",
```
marcjohnson-kint: Personally, I prefer hyphenating rather than camel casing in resource paths

```suggestion
          "pathPattern": "/inventory/config/instances/blocked-fields",
```

src/main/java/org/folio/inventory/resources/InventoryConfiguration.java
```diff
@@ -0,0 +1,21 @@
+package org.folio.inventory.resources;
+
+import com.google.common.collect.Sets;
+import java.util.Set;
+
+/**
+ * Stores configuration properties for Inventory entities: Instances, Items, Holdings
+ */
+public class InventoryConfiguration {
+  protected static final Set<String> BLOCKED_FIELDS = Sets.newHashSet(
```
marcjohnson-kint: Why is this protected, there does not appear to be any inheritance?

src/test/java/api/InstancesApiExamples.java
```diff
@@ -308,6 +308,17 @@ public void shouldReturnServerErrorIfOneOfInstancesPostedWithoutTitle() throws M
...
 
+  @Test
+  public void shouldReturnBlockedFieldsConfig() throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException {
+    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
+    okapiClient.get(ApiRoot.blockedFieldsConfig(), ResponseHandler.json(getCompleted));
+    Response getResponse = getCompleted.get(5, TimeUnit.SECONDS);
+
+    JsonObject response = getResponse.getJson();
+    assertThat(response.containsKey("blockedFields"), is(true));
+    assertThat(response.getJsonArray("blockedFields"), notNullValue());
```
marcjohnson-kint: Why is this only checking that the array is present, rather than it's contents?

src/test/java/api/InstancesApiExamples.java
```diff
@@ -308,6 +308,17 @@ public void shouldReturnServerErrorIfOneOfInstancesPostedWithoutTitle() throws M
     assertEquals(postResponse.getJson().getInteger("totalRecords"), Integer.valueOf(2));
   }
 
+  @Test
+  public void shouldReturnBlockedFieldsConfig() throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException {
+    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
```
marcjohnson-kint: Personally, I find it easier to read code when variable declaration and usage are separated by whitespace.

src/test/java/api/InstancesApiExamples.java
```diff
@@ -308,6 +308,17 @@ public void shouldReturnServerErrorIfOneOfInstancesPostedWithoutTitle() throws M
...
+    okapiClient.get(ApiRoot.blockedFieldsConfig(), ResponseHandler.json(getCompleted));
+    Response getResponse = getCompleted.get(5, TimeUnit.SECONDS);
+
+    JsonObject response = getResponse.getJson();
```
marcjohnson-kint: What happens if the response is a failure, or responds with a different status code?

src/test/java/api/support/ApiRoot.java
```diff
@@ -10,6 +10,10 @@ public static String inventory() {
     return String.format("%s/inventory", ApiTestSuite.apiRoot());
   }
 
+  public static String inventoryConfig() {return String.format("%s/inventory/config", ApiTestSuite.apiRoot());}
```
marcjohnson-kint: Why are these two paths separate? Is there another endpoint provided under `/inventory/config`?

ramls/blocked_fields.json
Igor-Gorchakov: Sure, I could rename that schemas as well

src/main/java/org/folio/inventory/resources/InventoryConfiguration.java
Igor-Gorchakov: Sonar advised me to use 'protected' access instead of 'public' , because of this field is used within the class in the same package.

src/test/java/api/InstancesApiExamples.java
Igor-Gorchakov: I added checking for contents

src/test/java/api/InstancesApiExamples.java
Igor-Gorchakov: Agree

src/test/java/api/InstancesApiExamples.java
Igor-Gorchakov: I added checking for proper status code 200

src/main/java/org/folio/inventory/resources/InventoryConfiguration.java
marcjohnson-kint: It seems Sonar gets very unhappy with this being public. I suggest we put this back to protected, apologies :-/

descriptors/ModuleDescriptor-template.json
```diff
@@ -2,6 +2,18 @@
...
   "name": "Inventory Module",
   "provides": [
+    {
+      "id": "inventory-config",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/inventory/config/instances/blocked-fields",
+          "permissionsRequired": ["inventory.config.instances.blocked-fields"],
```
marcjohnson-kint:
```suggestion
          "permissionsRequired": ["inventory.config.instances.blocked-fields.get"],
```

Igor-Gorchakov:
@marc-johnson , thanks for your code review. I could not resolve that warnings from lint-raml-cop. Do you have ideas why raml-cop warns my changes ?

Igor-Gorchakov:
Hi @dcrossleyau, I could not resolve that warnings from lint-raml-cop. I guess raml warnings are about the documentation generation process. Could you please help us to understand how to fix it. I'm not clear of resolution procedure, maybe those warnings would be fixed by somebody else after merge is done ?

dcrossleyau:
@Igor-Gorchakov, The WARNING is not a problem. Both the lint-raml and generate-api-docs scripts will both still discover all RAML files, including new ones, and process them. Usually this set of notifications would only be presented if there is an actual ERROR.

Yes later there eventually needs to be followup to adjust the configuration. Do not worry about that.

In this case there is something drastically wrong with the new inventory-config.raml file. It crashes the outer script that calls raml-cop. That is why there are no further messages from the processing of that file.

I am investigating today.

dcrossleyau:
It was indentation errors ([pull/121](https://github.com/folio-org/mod-inventory/pull/121)) in inventory-config.raml which cause YAML parser errors.

Also there was a deficiency with our 'lint-raml' script not detecting that.

PR #123, opened 2019-07-01
MODINV-122 ISBN normalization API

pautau-s:
## Purpose
[MODINV-122](https://issues.folio.org/browse/MODINV-122) implement API for ISBN validation and conversion

## Approach
- defined isbn-utils interface in moduleDescriptor
- created raml file for new API
- added implementation and unit tests for defined API


src/main/java/org/folio/inventory/resources/IsbnUtilsApi.java
```diff
@@ -0,0 +1,157 @@
...
+      case "isbn10":
+        isValid = IsbnUtil.isValid10DigitNumber(isbnCode);
+        break;
+      case "isbn13":
+        isValid = IsbnUtil.isValid13DigitNumber(isbnCode);
+        break;
+      case ISBN_PARAM:
+        isValid = IsbnUtil.isValid10DigitNumber(isbnCode) || IsbnUtil.isValid13DigitNumber(isbnCode);
+        break;
+      default: ClientErrorResponse.badRequest(routingContext.response(),
```
piotr-kalashuk: Instead of plain text, please return error json

ramls/isbn.raml
```diff
@@ -0,0 +1,91 @@
...
+version: v1
+protocols: [ HTTP, HTTPS ]
+baseUri: http://localhost
+
+documentation:
+  - title: "ISBN API"
+    content: <b>API for validation and conversion of ISBN-10 and ISBN-13 numbers </b>
+
+types:
+  errors: !include raml-util/schemas/errors.schema
```
piotr-kalashuk: As far as I can see the logic returns one error maximum. In this case `error.schema` can be used instead

descriptors/ModuleDescriptor-template.json
```diff
@@ -157,6 +157,30 @@
           "permissionsRequired": ["inventory.ingest.mods.status.get"]
         }
       ]
+    },
+    {
+      "id": "isbn-utils",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["get"],
```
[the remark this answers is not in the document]
pautau-s: Done

descriptors/ModuleDescriptor-template.json
```diff
@@ -300,6 +324,31 @@
         "inventory.ingest.mods.status.get",
         "inventory.config.instances.blocked-fields.get"
       ]
+    },
+    {
+      "permissionName": "isbn-utils.convert-to-13.get",
+      "displayName": "Isbn utils - convert to isbn13 format",
```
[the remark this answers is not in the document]
pautau-s: Updated

src/main/java/org/folio/inventory/resources/IsbnUtilsApi.java
pautau-s: Done

ramls/isbn.raml
pautau-s: Done

marcjohnson-kint: I might be too late with this comment.

Personally I think that it is better to use a structure which allows for multiple errors. 

It allows for future expansion and is more uniform / consistent with other APIs (the shared RAML defines a [trait](https://github.com/folio-org/raml/blob/raml1.0/traits/validation.raml) for this purpose), and clients can use the same parsing and interpretation logic across APIs.

pautau-s: Returned `errors` back.

ramls/isbn.raml
```diff
@@ -0,0 +1,104 @@
...
+          default: false
+          required: false
+      responses:
+        200:
+          description: "The response body is a json object containing the converted isbn"
+          body:
+            application/json:
+              type: isbn
+              example: !include examples/isbn13.json
+        400:
```
marcjohnson-kint: The standard API design uses [400 with plain text](https://github.com/folio-org/raml/blob/69f6074f5f48fd58e379e9ae76c892aa68c115ea/rtypes/item-collection.raml#L71), and [422 with JSON errors](https://github.com/folio-org/raml/blob/69f6074f5f48fd58e379e9ae76c892aa68c115ea/traits/validation.raml#L3).

What is different about the ISBN normalization API to mean that 400 with JSON is the chosen response?

ramls/isbn.raml
```diff
@@ -0,0 +1,104 @@
...
+            application/json:
+              type: errors
+              example:
+                strict: false
+                value: !include raml-util/examples/errors.sample
+  /validator:
+    get:
+      description: Checks the code is a valid ISBN code.
+      queryParameters:
+        /^(isbn|isbn10|isbn13)$/: string
```
marcjohnson-kint: Wha does this mean? That a single query string parameter should be provided, and it could be either `isbn`, `isbn10` or `isbn13`?

ramls/isbn.raml
```diff
@@ -0,0 +1,104 @@
...
+protocols: [ HTTP, HTTPS ]
+baseUri: http://localhost
+
+documentation:
+  - title: "ISBN API"
+    content: <b>API for validation and conversion of ISBN-10 and ISBN-13 numbers </b>
+
+types:
+  errors: !include raml-util/schemas/errors.schema
+  isbn:
```
marcjohnson-kint: Why are these types defined inline, rather than as JSON schema?

@dcrossleyau Do our RAML validation tools support definitions like this?

dcrossleyau: The tools should handle it. And yes, looking at the CI job report from the last run, there were no complaints.

If it is going to be re-used in other places, then better to be an "include".

ramls/isbn.raml
piotr-kalashuk: Yes, this is kind of choice. The logic validates that only one parameter is specified and returns an error if not.

ramls/isbn.raml
pautau-s: The main idea is that it is easier for the client to process the response in JSON format than to parse the text.

craigmcnally:
ping @marc-johnson do you see issues with this or have suggestions?  If not, can we please get an approval on this?  It's blocking other work that we're trying to get into the Q3.1 release.

craigmcnally:
@pautau-s please bring your fork up-to-date with master

craigmcnally:
This has now been open for 11 days...  Can we please get this approved/merged by someone with appropriate permissions?

PR #125, opened 2019-07-09
[MODINV-129] Generate source and date information for circulation notes

opochepam:
## Purpose
Resolves: MODINV-129
Generate source and date information for circulation notes when item is created/updated
## Links
https://issues.folio.org/browse/MODINV-129
https://github.com/folio-org/mod-inventory-storage/pull/289


src/test/java/support/fakes/FakeOkapi.java
```diff
@@ -178,4 +179,12 @@ private void registerFakeContributorNameTypesModule(Router router) {
       .withCollectionPropertyName("contributorNameTypes")
       .create().register(router);
   }
+
+  private void registerFakeUsersModule(Router router) {
```
marcjohnson-kint: The new dependency on the user’s interface needs expressing in the Module Descriptor requires section.

opochepam: done

ramls/item.json
```diff
@@ -176,6 +176,34 @@
             "type": "boolean",
             "description": "Records the fact that the note should only be displayed for staff",
             "default": false
+          },
+          "source": {
```
marcjohnson-kint: If these properties are intended to be stored, then the required version of the instance-storage interface should be changed to reflect the expanded storage properties.

ramls/item.json
```diff
@@ -176,6 +176,34 @@
...
+            "type": "object",
+            "description": "The user who added the note",
+            "properties": {
+              "id": {
+                "type": "string",
+                "description": "The id of the user who added the note"
+              },
+              "personal": {
```
marcjohnson-kint: What is the purpose of storing the personal information of the user who created the note? Or are these derived properties, fetched on demand when the record is read?

ramls/item.json
```diff
@@ -176,6 +176,34 @@
...
+                  },
+                  "firstName": {
+                    "description": "The user's first name",
+                    "type": "string"
+                  }
+                }
+              }
+            }
+          },
+          "date": {
```
marcjohnson-kint: Is this the date when the note was added?

If so, could it have a more explicit name like dateAdded?

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -937,6 +957,158 @@ public void canRemoveBarcodeFromAnExistingItem()
...
+    assertThat(permanentLoanType.getString("name"), is("Can Circulate"));
+
+    assertThat("Item should not have permanent location",
+      updatedItem.containsKey("permanentLocation"), is(false));
+
+    assertThat(updatedItem.getJsonObject("temporaryLocation").getString("name"), is("Reading Room"));
+
+    JsonObject checkInNote = updatedItem.getJsonArray(CIRCULATION_NOTES_KEY).getJsonObject(0);
+
+    assertThat(checkInNote.getString(DATE_KEY), not(createdItemCirculationNoteDate));
```
marcjohnson-kint: Does this check that the date the note was added does not change on update?

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -592,12 +610,27 @@ private void addItem(
     RoutingContext routingContext,
     WebContext webContext,
     Item newItem,
-    ItemCollection itemCollection) {
+    ItemCollection itemCollection,
+    UserCollection userCollection) {
+
+    String userId = routingContext.request().getHeader("X-Okapi-User-Id");
```
marcjohnson-kint: Why does this not use the method added to the context?

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -592,12 +610,27 @@ private void addItem(
...
-    itemCollection.add(newItem, success -> {
-      Item item = success.getResult();
-      respondWithItemRepresentation(item, STATUS_CREATED, routingContext, webContext);
-    }, FailureResponseConsumer.serverError(routingContext.response()));
+        itemCollection.add(newItem.withCirculationNotes(notes), success -> {
+          Item item = success.getResult();
+          respondWithItemRepresentation(item, STATUS_CREATED, routingContext, webContext);
+        }, FailureResponseConsumer.serverError(routingContext.response()));
+      },
+      FailureResponseConsumer.serverError(routingContext.response()));
```
marcjohnson-kint: Does this mean that if the user cannot be found, the whole request to create the instance will fail?

ramls/item.json
opochepam: It is a requirement from the following story https://issues.folio.org/browse/MODINVSTOR-310.

The properties is fetched when the note is saved or updated. For example if the user created the note and another user updated, the information will change.

ramls/item.json
opochepam: Actually it is the date when the note is added or updated. As I understood from the Emma's comment from the following story it should work in this way. https://issues.folio.org/browse/MODINV-129

src/test/java/api/items/ItemApiExamples.java
opochepam: The check reflects that the date is changed when the note is updated.

src/main/java/org/folio/inventory/resources/Items.java
opochepam: Replaced by fetching the `userId` from the `WebContext`.

ramls/item.json
opochepam: done

src/main/java/org/folio/inventory/resources/Items.java
opochepam: Yes, it does.

ramls/item.json
marcjohnson-kint: As these properties are stored copies of information from the user (presumably to support sorting) they should be marked read only as this behaviour is controlled by the server (and this usage reflected in the description).

src/main/java/org/folio/inventory/resources/Items.java
marcjohnson-kint: This means that instances cannot be created without a user header (and is dependent upon the availability of the module providing the `users` interface)

Does the data import create circulation notes? If so, does it run in the context of a user (I imagine it is a background process)?

ramls/item.json
```diff
@@ -176,6 +176,34 @@
...
             "description": "Records the fact that the note should only be displayed for staff",
             "default": false
+          },
+          "source": {
+            "type": "object",
+            "description": "The user who added the note",
+            "properties": {
+              "id": {
+                "type": "string",
+                "description": "The id of the user who added the note"
```
marcjohnson-kint: Is this always the user who added the note, or does it change when the note is updated (to the user who last updated the note)?

ramls/item.json
opochepam: Made source and date `readOnly` and updated the description.

ramls/item.json
opochepam: It changes when the note is updated as well. Updated the description.

src/main/java/org/folio/inventory/resources/Items.java
opochepam: What do you mean by data import?

ramls/item.json
marcjohnson-kint: Thank you for updating this. Typically we mark all of the individual properties as read only, not only the object itself.

ramls/item.json
marcjohnson-kint: Then I suggest the property is named explicitly for that purpose.

```suggestion
          "dateAddedOrUpdated": {
```
```suggestion
          "createdOrUpdatedDate": {
```

src/main/java/org/folio/inventory/common/MessagingContext.java
```diff
@@ -24,6 +24,11 @@ public String getOkapiLocation() {
     return getHeader("okapiLocation");
   }
 
+  @Override
+  public String getUserId() {
+    return getHeader("userId");
```
marcjohnson-kint: There appears to be no mapping to this header when messages are created (see `IngestMessages`). Meaning this value will always be null.

If this is deliberate, I think it could be easier to follow by returning null explicitly.

Does this work because the MODS Ingestion does not create any circulation notes?

opochepam: Yes, MODS Ingestion does not create any circulation notes. Changed to return null explicitly.

ramls/item.json
opochepam: Marked the indivudal properties as read only as well.

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -771,38 +804,83 @@ private boolean hasSameBarcode(Item updatedItem, Item foundItem) {
...
 
+  private CirculationNote updateCirculationNoteIfChanged(CirculationNote newNote,
+                                                         User user, Map<String, CirculationNote> oldNotes) {
+    String noteId = newNote.getId();
+
+    if (noteId == null) {
+      return newNote.withId(UUID.randomUUID().toString())
+        .withSource(user)
+        .withDate(dateTimeFormatter.format(ZonedDateTime.now()));
+    } else if (circulationNoteChanged(newNote, oldNotes.get(noteId))){
```
kostyantyn-kh:
```suggestion
    } else if (circulationNoteChanged(newNote, oldNotes.get(noteId))) {
```

opochepam: done

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -150,17 +166,19 @@ private void update(RoutingContext routingContext) {
...
 
     ItemCollection itemCollection = storage.getItemCollection(context);
+    UserCollection userCollection = storage.getUserCollection(context);
 
     itemCollection.findById(routingContext.request().getParam("id"), getItemResult -> {
-      if(getItemResult.getResult() != null) {
-        if(hasSameBarcode(updatedItem, getItemResult.getResult())) {
-          updateItem(routingContext, updatedItem, itemCollection);
+      Item oldItem = getItemResult.getResult();
+      if(oldItem != null) {
```
kostyantyn-kh:
```suggestion
      if (oldItem != null) {
```

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -150,17 +166,19 @@ private void update(RoutingContext routingContext) {
...
+        if(hasSameBarcode(newItem, oldItem)) {
```
kostyantyn-kh:
```suggestion
        if (hasSameBarcode(newItem, oldItem)) {
```

src/main/java/org/folio/inventory/common/WebContext.java
```diff
@@ -25,6 +25,11 @@ public String getOkapiLocation() {
     return getHeader("X-Okapi-Url", "");
   }
 
+  @Override
+  public String getUserId() {
+    return getHeader("X-Okapi-User-Id", "");
```
OleksiiKuzminov: Maybe it will be a good idea to move all strings with headers names to the constants.

opochepam: Moved to the constants.

PR #129, opened 2019-07-11
MODINV-124 Protect Instance fields from being updated

Igor-Gorchakov:
MODINV-124 - Protect Instance fields from being updated.

If someone is trying to load Instances directly into the mod-inventory (bypassing the frontend) and the Instance's data is being controlled by a linked SRS MARC record, then the mod-inventory should stop Instance's fields from update by the load.


src/test/java/api/InstancesApiExamples.java
```diff
@@ -419,6 +417,141 @@ public void cannotUpdateAnInstanceThatDoesNotExist()
...
+
+    for (String field : config.getInstanceBlockedFields()) {
+      URL instanceLocation = new URL(String.format("%s/%s", ApiRoot.instances(), id));
+      // Put Instance for update
+      CompletableFuture<Response> putCompleted = new CompletableFuture<>();
+      okapiClient.put(instanceLocation, instanceForUpdate, ResponseHandler.any(putCompleted));
+      Response putResponse = putCompleted.get(5, TimeUnit.SECONDS);
+
+      assertThat(putResponse.getStatusCode(), is(422));
+      assertThat(putResponse.getJson().getJsonArray("errors").size(), is(1));
```
Igor-Gorchakov: Here is how a response body looks if the statusCode=422 :
```

{  
   "errors":[  
      "Unprocessable entity: given Instance is controlled by MARC record, it's blocked fields can not be updated: discoverySuppress,previouslyHeld,statusId,hrid,staffSuppress,source,clickable-add-statistical-code"
   ]
}
```

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -66,10 +72,14 @@
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   private final Storage storage;
   private final HttpClient client;
+  private final InventoryConfiguration config;
+  private static final String BLOCKED_FIELDS_UPDATE_ERROR_MESSAGE = "Unprocessable entity: given Instance is controlled by MARC record, "
+    + "it's blocked fields can not be updated: ";
```
Anatolii-Starkov: I think it should be "These blocked fields can not be updated: " or "it's blocked fields, they can not be updated: "

Igor-Gorchakov:
Never create git branches with underscore symbols, such naming breaks docker build :)

PR #130, opened 2019-07-11
MODINV-124 Protect Instance fields from being updated

Igor-Gorchakov:
MODINV-124 - Protect Instance fields from being updated.

If someone is trying to load Instances directly into the mod-inventory (bypassing the frontend) and the Instance's data is being controlled by a linked SRS MARC record, then the mod-inventory should stop Instance's fields from update by the load.


src/test/java/api/InstancesApiExamples.java
```diff
@@ -419,6 +417,141 @@ public void cannotUpdateAnInstanceThatDoesNotExist()
...
+
+    for (String field : config.getInstanceBlockedFields()) {
+      URL instanceLocation = new URL(String.format("%s/%s", ApiRoot.instances(), id));
+      // Put Instance for update
+      CompletableFuture<Response> putCompleted = new CompletableFuture<>();
+      okapiClient.put(instanceLocation, instanceForUpdate, ResponseHandler.any(putCompleted));
+      Response putResponse = putCompleted.get(5, TimeUnit.SECONDS);
+
+      assertThat(putResponse.getStatusCode(), is(422));
+      assertThat(putResponse.getJson().getJsonArray("errors").size(), is(1));
```
marcjohnson-kint: Could it be worthwhile to add checks that the validation error is the one that is expected?

Does the implementation accumulate errors for each blocked property that is different?

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -66,10 +72,14 @@
   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
   private final Storage storage;
   private final HttpClient client;
+  private final InventoryConfiguration config;
+  private static final String BLOCKED_FIELDS_UPDATE_ERROR_MESSAGE = "Unprocessable entity: given Instance is controlled by MARC record, "
```
marcjohnson-kint: The unprocessable entity statement at the beginning of this message seems redundant to me, given it is a validation error being sent with a 422 status code.

What do you think?

Has this message been agreed with the PO for this work?

src/test/java/api/InstancesApiExamples.java
Igor-Gorchakov: I think -yes, It is worth to add a check for the whole error message.
The implementation does not accumulate errors and differences - this implementation may be called only if the client sends request bypassing the frontend. The work to keep fields in a proper unchanged values lies on a client is this case.

src/main/java/org/folio/inventory/resources/Instances.java
Igor-Gorchakov: I agree, I will skip this prefix. Message has not been agreed with PO, It's up to us to choose the error message. The main idea of this Jira ticket is to stop Instance from being updated.

src/test/java/api/InstancesApiExamples.java
Igor-Gorchakov: I added a check for the error message in this test method `canNotUpdateAnExistingMARCInstanceIfBlockedFieldsAreChangedToNulls`

src/main/java/org/folio/inventory/resources/Instances.java
Igor-Gorchakov: I removed that redundant prefix

Igor-Gorchakov:
@marc-johnson , waiting for you review

marcjohnson-kint:
> @marc-johnson , waiting for your review

This pull request has already been approved by two other folks.

Do you need a review from me specifically? 

If so, it will likely need to wait until Monday. If not, this can be merged based upon other approvals.

PR #133, opened 2019-07-28
MODINV-135 - Inventory App: Assign tags to instance record

natali-zaitseva:
## Purpose 
Due to https://issues.folio.org/browse/MODINV-135 we would like to introduce tags for instance entities.

## Approach
* updated instance with tags
* updated instances implementation to include tags
* updated instance tests accordingly

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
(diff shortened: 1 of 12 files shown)


natali-zaitseva:
should be merged after https://github.com/folio-org/mod-inventory-storage/pull/301

natali-zaitseva:
Hi, @marcjohnson-kint thanks for the note regarding interfaces versions. Updated.

PR #134, opened 2019-07-29
MODINV-146 - Inventory App: Assign tags to instance record

natali-zaitseva:
## Purpose 
Due to https://issues.folio.org/browse/MODINV-135 we would like to introduce tags for instance entities.

## Approach
* updated instance with tags
* updated interfaces versions
* updated instances implementation to include tags
* updated instance tests accordingly


natali-zaitseva:
should be merged after https://github.com/folio-org/mod-inventory-storage/pull/302

src/main/java/org/folio/inventory/domain/instances/Instance.java
```diff
@@ -76,6 +78,7 @@
   private String statusId;
   private String statusUpdatedDate;
   private Metadata metadata = null;
+  private List<String> tag;
```
marcjohnson-kint: Is this a single tag, or is it intended to represent multiple tags?
```suggestion
  private List<String> tags;
```

natali-zaitseva: @marcjohnson-kint it is meant to be multiple, will fix it. Thanks

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -1,5 +1,16 @@
 package org.folio.inventory.resources;
 
+import io.vertx.core.CompositeFuture;
```
marcjohnson-kint: We typically follow the Eclipse style ordering of imports, see https://issues.folio.org/browse/FOLIO-1667?focusedCommentId=41677&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-41677 for more details

Please put these import statements back to the end of the list

natali-zaitseva: @marcjohnson-kint fixed imports

PR #136, opened 2019-08-01
MODINV-139- 400/422 error reporting

KVupp:
## PURPOSE
https://issues.folio.org/browse/MODINV-139

## APPROACH
- raml: Modified 400 Error code to return text/plain response instead of application/json
- Modified Unit tests accordingly

### DIFF
--- a/ramls/isbn.raml
+++ b/ramls/isbn.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Inventory API
-version: v1
+version: v2
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost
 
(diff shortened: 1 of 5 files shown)


ramls/isbn.raml
```diff
@@ -47,11 +46,8 @@ types:
         400:
           description: "Bad request, an invalid ISBNs is provided"
           body:
-            application/json:
-              type: errors
-              example:
-                strict: false
-                value: !include raml-util/examples/errors.sample
+            text/plain:
+              example: "ISBN value is invalid"
```
kostyantyn-kh: Looks like changing response from json to plane text is breaking change and should be reflected in raml and module descriptor versions.

KVupp: Done, increased the major version for "isbn-utils" interface

PR #137, opened 2019-08-05
MODINV-134 - Update batch post instances to send batch save to inventory-storage

RuslanLavrov:
## PURPOSE
https://issues.folio.org/browse/MODINV-134

## APPROACH
- changed logic of post /inventory/instances/batch using new batch save endpoint in inventory-storage to create Instances
- updated tests

### DIFF
--- a/src/main/java/org/folio/inventory/domain/AsynchronousCollection.java
+++ b/src/main/java/org/folio/inventory/domain/AsynchronousCollection.java
@@ -5,6 +5,7 @@ import org.folio.inventory.common.domain.Failure;
 import org.folio.inventory.common.domain.MultipleRecords;
 import org.folio.inventory.common.domain.Success;
 
+import java.util.List;
 import java.util.function.Consumer;
 
 public interface AsynchronousCollection<T> {
@@ -30,4 +31,10 @@ public interface AsynchronousCollection<T> {
   void update(T item,
               Consumer<Success<Void>> completionCallback,
               Consumer<Failure> failureCallback);
+
+  default void addBatch(List<T> items,
+                        Consumer<Success<BatchResult<T>>> resultCallback,
+                        Consumer<Failure> failureCallback) {
+    throw new UnsupportedOperationException();
+  }
 }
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/inventory/domain/AsynchronousCollection.java
```diff
@@ -30,4 +31,10 @@ void delete(String id,
   void update(T item,
               Consumer<Success<Void>> completionCallback,
               Consumer<Failure> failureCallback);
+
+  default void addBatch(List<T> items,
+                        Consumer<Success<BatchResult<T>>> resultCallback,
+                        Consumer<Failure> failureCallback) {
+    throw new UnsupportedOperationException();
```
marcjohnson-kint: Why has a default implementation been added? I think this should be an active decision by each implementor of the interface. If some of the implementors aren't going to implement this, then maybe it should be a separate interface?

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -0,0 +1,414 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+
+public abstract class AbstractInstances {
```
marcjohnson-kint: Personally, I prefer composition over inheritance. Could the shared methods have been moved to separate classes rather than an abstract base class?

src/main/java/org/folio/inventory/domain/AsynchronousCollection.java
RuslanLavrov: thank you for review, we will take note of this. I created the [task](https://issues.folio.org/browse/MODINV-154) to make refactoring in this way.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
RuslanLavrov: shared methods was moved to abstract class cause some of the methods closely related with instance creation flow, so they was moved to super class to keep endpoints logic concentrated in the classes that define endpoints. Thank you for review, I added this point to [task](https://issues.folio.org/browse/MODINV-154) for refactoring.

PR #138, opened 2019-08-06
MODINV-153 Fix InstancesBatchResponse body after addition of "tags"

KaterynaSenchenko:


marcjohnson-kint:
I'm aware this has already been merged.

There were no test changes in this pull request, suggesting that there isn't a test which demonstrates either this bug or the successful processing of a batch with tags. Could tests be added separately?

KaterynaSenchenko:
Hi @marcjohnson-kint, it's a quick fix and the whole implementation of batch endpoint will be changed  after https://github.com/folio-org/mod-inventory/pull/137 is merged. We can add all the necessary tests with that PR.

marcjohnson-kint:
> Hi @marcjohnson-kint, it's a quick fix and the whole implementation of batch endpoint will be changed after #137 is merged. We can add all the necessary tests with that PR.

Sure, thanks

PR #139, opened 2019-08-09
update suppress from discovery flag in SRS on instance update (MODINV-144)

OleksiiKuzminov:
Libraries sometimes want to suppress records from discovery by their patrons. Much of the discovery work involves harvesting SRS MARC Bib records via OAI-PMH. When a user updates this flag in the FOLIO instance, the corresponding change should be made to the related SRS MARC Bib record, which then in turn updates the flag in the MARCcat MARC Bib record

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,6 +110,12 @@
       <artifactId>joda-time</artifactId>
       <version>2.10.3</version>
     </dependency>
+    <dependency>
+      <groupId>org.folio</groupId>
+      <artifactId>mod-source-record-storage-client</artifactId>
+      <version>2.6.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -110,6 +110,12 @@
       <artifactId>joda-time</artifactId>
       <version>2.10.3</version>
     </dependency>
+    <dependency>
+      <groupId>org.folio</groupId>
+      <artifactId>mod-source-record-storage-client</artifactId>
+      <version>2.6.0-SNAPSHOT</version>
```
marcjohnson-kint: This will need changing prior to the next formal release, as SNAPSHOT dependencies are not intended to be used.

PR #141, opened 2019-08-13
MODINV-149 UIIN-470: Backend for nature of content on Instance

Anatolii-Starkov:
-added the natureOfContentTermIds field to the Instance

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
-      "version": "9.3",
+      "version": "9.4",
       "handlers": [
         {
           "methods": ["GET"],
@@ -196,11 +196,11 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.1"
+      "version": "7.2"
     },
     {
       "id": "instance-storage-batch",
-      "version": "0.1"
+      "version": "0.4"
     },
     {
       "id": "holdings-storage",
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-inventory</artifactId>
   <groupId>org.folio</groupId>
-  <version>12.1.0-SNAPSHOT</version>
+  <version>12.2.0-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache License 2.0</name>
--- a/ramls/inventory-batch.raml
+++ b/ramls/inventory-batch.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Inventory API
-version: v0.3
+version: v0.4
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost
 
--- a/src/test/java/api/InstancesApiExamples.java
+++ b/src/test/java/api/InstancesApiExamples.java
@@ -41,6 +41,7 @@ import static api.support.InstanceSamples.taoOfPooh;
 import static api.support.InstanceSamples.temeraire;
 import static api.support.InstanceSamples.treasureIslandInstance;
 import static api.support.InstanceSamples.uprooted;
+import static java.util.Arrays.asList;
 import static org.folio.inventory.domain.instances.Instance.TAGS_KEY;
 import static org.folio.inventory.domain.instances.Instance.TAG_LIST_KEY;
 import static org.hamcrest.CoreMatchers.hasItem;
@@ -81,7 +82,8 @@ public class InstancesApiExamples extends ApiTests {
         .put("name", "Chambers, Becky")))
       .put("source", "Local")
       .put("instanceTypeId", ApiTestSuite.getTextInstanceType())
-      .put(TAGS_KEY, new JsonObject().put(TAG_LIST_KEY, new JsonArray().add(tagNameOne)));
+      .put(TAGS_KEY, new JsonObject().put(TAG_LIST_KEY, new JsonArray().add(tagNameOne)))
+      .put("natureOfContentTermIds", new JsonArray(asList("96879b60-098b-453b-bf9a-c47866f1ab2a", "f5908d05-b16a-49cf-b192-96d55a94a0d1")));
 
     CompletableFuture<Response> postCompleted = new CompletableFuture<>();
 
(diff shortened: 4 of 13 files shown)


marcjohnson-kint:
There does not appear to be a corresponding change in mod-inventory-storage for this new property, until that is present, and the relevant interface dependencies updated, this should not be merged

marcjohnson-kint:
What is expected to happen when an instance is created with a nature of content term ID that does not refer to an existing record?

src/test/java/api/InstancesApiExamples.java
```diff
@@ -81,7 +82,8 @@ public void canCreateAnInstance()
         .put("name", "Chambers, Becky")))
       .put("source", "Local")
       .put("instanceTypeId", ApiTestSuite.getTextInstanceType())
-      .put(TAGS_KEY, new JsonObject().put(TAG_LIST_KEY, new JsonArray().add(tagNameOne)));
+      .put(TAGS_KEY, new JsonObject().put(TAG_LIST_KEY, new JsonArray().add(tagNameOne)))
+      .put("natureOfContentTermIds", new JsonArray(asList("96879b60-098b-453b-bf9a-c47866f1ab2a", "f5908d05-b16a-49cf-b192-96d55a94a0d1")));
```
marcjohnson-kint: I imagine these IDs do not refer to nature of content term records, meaning that if these tests are used with a real storage module they will likely fail.

I believe the other other tests deliberately use a fake module and create the records they need. Could this be done for these tests as well?

Anatolii-Starkov: Done

marcjohnson-kint: Thanks

marcjohnson-kint:
> The `instance-storage` dependency needs to be updated to version 7.2 as this is dependent upon [these changes](https://github.com/folio-org/mod-inventory-storage/pull/311)

Ignore that completely, this has already been done, apologies

Anatolii-Starkov:
Marc, could you approve it if you don't have any objections?

Anatolii-Starkov:
Right. I think it's already figured out in another PR from @bohdan-suprun that is related to this PR. 
Anyway. I think validation is out of scope this ticket, because we need to develop a mechanism of validation of a list of UUIDs as it is common issue.

I don't think it's a good idea to make a call for each id to check either it exists or not.
The validation should be on storage side in terms of performance it will cost us less than http call per ID.
As I mentioned it above. It's common issue, the Instance has many lists of IDs and we should perform that validation for all of them.

marcjohnson-kint:
> Right. I think it's already figured out in another PR from @bohdan-suprun that is related to this PR.
> Anyway. I think validation is out of scope this ticket, because we need to develop a mechanism of validation of a list of UUIDs as it is common issue.

Indeed, reference validation of array properties is an outstanding area across FOLIO. As far as I'm aware there isn't a standard way to do it.

I ask the question to understand the thought process the developer has gone through, and if they've attempted to implement this.

descriptors/ModuleDescriptor-template.json
```diff
@@ -196,11 +196,11 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.1"
+      "version": "7.2"
     },
     {
       "id": "instance-storage-batch",
-      "version": "0.1"
+      "version": "0.4"
```
marcjohnson-kint: Why has this been moved on 3 versions, is that because it wasn't in previous versions and should've been?

pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-inventory</artifactId>
   <groupId>org.folio</groupId>
-  <version>12.1.0-SNAPSHOT</version>
+  <version>12.2.0-SNAPSHOT</version>
```
marcjohnson-kint: Why has this version been changed?

Version 12.1.0 has not been released yet, and these changes are minor, meaning they fit within the bounds of the current 12.0.0 -> 12.1.0 increment.

Is this because of the conversation we had about changing the major implementation version in the other pull request?

descriptors/ModuleDescriptor-template.json
Anatolii-Starkov: In raml file the version is 0.3, so i decided that somebody forgot to change the version here

src/test/java/api/InstancesApiExamples.java
```diff
@@ -53,6 +21,41 @@
...
+import com.github.jsonldjava.core.JsonLdProcessor;
+
+import api.support.ApiRoot;
+import api.support.ApiTests;
+import api.support.InstanceApiClient;
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
 public class InstancesApiExamples extends ApiTests {
```
marcjohnson-kint: There doesn't appear to be any test for changing the nature of contents of an existing instance, please can a test for this be added e.g. removing and adding a different nature of content

It can likely form part of the existing test for replacing / editing an instance

pom.xml
Anatolii-Starkov: From the https://dev.folio.org/guidelines/contributing/#version-numbers

> For modules the middle part should be incremented if you implemented an addition to the API(s), (the minor version of the particular API has been changed). For non-modules it may also mean any additional functionality. For both, the change must be backwards compatible with respect to any client code or agents.

according to this " (the minor version of the particular API has been changed)" I decided to change the version of impl.

marcjohnson-kint: > From the https://dev.folio.org/guidelines/contributing/#version-numbers
> 
> > For modules the middle part should be incremented if you implemented an addition to the API(s), (the minor version of the particular API has been changed). For non-modules it may also mean any additional functionality. For both, the change must be backwards compatible with respect to any client code or agents.
> 
> according to this " (the minor version of the particular API has been changed)" I decided to change the version of impl.

Alas, that guidance is incorrect in the context of pull requests (and probably out of date in other ways), the minor version number should have already been incremented during the last formal release.

Otherwise pretty much every pull request would require the implementation version to change and that would just unnecessarily waste version numbers.

Please reset the implementation version number. Apologies for the confusion.

Does that make sense?

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Ok, thanks for spotting this

pom.xml
Anatolii-Starkov: It's clear. Thanks

src/test/java/api/InstancesApiExamples.java
Anatolii-Starkov: Ok. I will do it

ramls/inventory-batch.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Inventory API
-version: v0.3
+version: v0.4
```
Igor-Gorchakov: @marcjohnson-kint  why to change interface version if nothing added in API declaration ?
We change minor number if things only added to the interface – e.g. a new resource or method on existing resources

src/test/java/api/InstancesApiExamples.java
Anatolii-Starkov: Done

Anatolii-Starkov:
the implementation version is reset

PR #146, opened 2019-09-04
Modinvstor 343 add preceding succeeding properties

bohdan-suprun:
* Add precedingTitle and succeedingTitle properties to instance
* Implement logic to handle those properties in get/getMany/create/update Instances resource + InstanceBatch resource
* Add unit tests

Related PR https://github.com/folio-org/mod-inventory-storage/pull/322


marcjohnson-kint:
Closed because it is out of scope of the original work. It might be revisited if the UI implementation encounters challenges as described in MODINVSTOR-343

PR #149, opened 2019-09-13
Pass X-Okapi-Request-Id MODINV-156

adamdickmeiss:
I can't believe how many classes I had to touch for this. There should
be an "OkapiClient" that just takes context from outer request
and relay headers as necessary.. Now the Okapi headers definitions
are duplicated.

### DIFF
--- a/src/main/java/org/folio/inventory/common/WebContext.java
+++ b/src/main/java/org/folio/inventory/common/WebContext.java
@@ -11,6 +11,7 @@ public class WebContext implements Context {
   private static final String OKAPI_TOKEN_HEADER = "X-Okapi-Token";
   private static final String OKAPI_URL_HEADER = "X-Okapi-Url";
   private static final String OKAPI_USER_ID_HEADER = "X-Okapi-User-Id";
+  private static final String OKAPI_REQUEST_ID = "X-Okapi-Request-Id";
 
   public WebContext(RoutingContext routingContext) {
     this.routingContext = routingContext;
@@ -34,6 +35,11 @@ public class WebContext implements Context {
   @Override
   public String getUserId() {
     return getHeader(OKAPI_USER_ID_HEADER, "");
+  }
+
+  @Override
+  public String getRequestId() {
+    return getHeader(OKAPI_REQUEST_ID);
   }
 
   @Override
--- a/src/main/java/org/folio/inventory/common/WebContext.java
+++ b/src/main/java/org/folio/inventory/common/WebContext.java
@@ -37,23 +37,19 @@ public class WebContext implements Context {
     return getHeader(OKAPI_USER_ID_HEADER, "");
   }
 
-  @Override
   public String getRequestId() {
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/inventory/support/http/client/OkapiHttpClient.java
```diff
@@ -150,6 +154,10 @@ private void okapiHeaders(HttpClientRequest request) {
       request.headers().add(TOKEN_HEADER, this.token);
     }
 
+    if (this.requestId != null) {
```
julianladisch: This is inconsistent, it should use the same logic as the token above:

    if (StringUtils.isNotBlank(this.requestId)) {

src/main/java/org/folio/inventory/common/WebContext.java
```diff
@@ -36,6 +37,11 @@ public String getUserId() {
     return getHeader(OKAPI_USER_ID_HEADER, "");
   }
 
+  @Override
+  public String getRequestId() {
+    return getHeader(OKAPI_REQUEST_ID);
```
julianladisch: This is inconsistent. All other methods return "" if undefined. Can you change to

    return getHeader(OKAPI_REQUEST_ID, "");

src/main/java/org/folio/inventory/support/http/client/OkapiHttpClient.java
adamdickmeiss: I don't agree. This code is faster and uses a very common idiom.. Null value for absent header.

src/main/java/org/folio/inventory/support/http/client/OkapiHttpClient.java
```diff
@@ -34,12 +35,21 @@
...
+    throws MalformedURLException {
+
+    this(httpClient, new URL(context.getOkapiLocation()),
+      context.getTenantId(), context.getToken(), context.getUserId(),
+      context.getRequestId(), exceptionHandler);
+  }
+
+  public OkapiHttpClient(HttpClient httpClient,
+    URL okapiUrl,
+    String tenantId,
```
julianladisch: Can you add javadoc that null, an empty ("") String or a whitespace only means no tenantId header?

src/main/java/org/folio/inventory/support/http/client/OkapiHttpClient.java
```diff
@@ -34,12 +35,21 @@
...
+    String token,
```
julianladisch: Can you add javadoc that null, an empty ("") String or a whitespace only means no token header?

src/main/java/org/folio/inventory/support/http/client/OkapiHttpClient.java
```diff
@@ -34,12 +35,21 @@
...
+    String userId,
+    String requestId,
```
julianladisch: Can you add javadoc null means no requestId header and that this is different from the tenantId and token parameters?

PR #150, opened 2019-09-13
MODINV-157: Transfer item effective location property from storage in…

bohdan-suprun:
Remove old item effectiveLocation property calculation logic and use new effectiveLocationId property from storage.


src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1144,6 +1152,84 @@ public void canUpdateAnItemWithExistingCirculationNote()
...
+    UUID itemId = UUID.randomUUID();
+
+    JsonObject newItemRequest = new JsonObject()
+      .put("id", itemId.toString())
+      .put("holdingsRecordId", createInstanceAndHolding().toString())
+      .put("materialTypeId", getDvdMaterialType())
+      .put("permanentLoanTypeId", getCanCirculateLoanType())
+      .put("permanentLocationId", getMainLibraryLocation())
+      .put("temporaryLocationId", getReadingRoomLocation())
+      .put("effectiveLocationId", getReadingRoomLocation());
```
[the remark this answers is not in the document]
bohdan-suprun: Hi @marcjohnson-kint, since we are using fake storage modules there is no ability to emulate triggers behavior. And this test verifies that effectiveLocationId is transferred from storage response and converted to `effectiveLocation` property (it is a object with `id` and `name` properties)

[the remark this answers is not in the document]
marcjohnson-kint: @bohdan-suprun 

From my perspective, this is one of the challenges with moving business logic to the storage modules. It increases the coupling between the business logic and storage modules, and creates a reliance on that behaviour.

My concern is that this approach is not representative of what will actually happen. I think that impedes it's ability to verify the expected behaviour.

To me, what this test demonstrates is that the business logic provides a property to storage and gets it's back, which I think is quite different to relying on the storage determining that property.

I agree that this increases the complexity needed within the fakes (in many ways, they represent the degree of coupling between the modules). 

Personally, I think that complexity needs to be represented in the tests, which might mean expanding the capabilities of the fakes or using a custom one.

src/test/java/api/ApiTestSuite.java
```diff
@@ -114,6 +116,7 @@ public static void after()
     stopInventoryVerticle();
     stopFakeModules();
     stopVertx();
+    getStorage().dropStorage();
```
marcjohnson-kint: What does `dropStorage` do?

marcjohnson-kint:
@bohdan-suprun Quite a lot of code has been added in order to mimic the behaviour of the storage system in deriving the effective location.

Can you help me understand the design better by explaining how they fit together?

I'll admit, I hadn't intended for the implementation details of the storage modules to be implemented within these fakes, rather only the outcome. I had in mind a way of configuring a function for deriving a property that when specified the fake would know when to use it.

src/test/java/api/ApiTestSuite.java
bohdan-suprun: It just cleans-up in memory storage.

bohdan-suprun:
@marcjohnson-kint, Right now we have set of FakeStorageModules and each module encapsulates it's own Map which represents a storage. 

There are actually two options to introduce this triggers-like logic:
1. We can provide a function to a fake, and it will call it on create/update.
1. Extract the storage as a separate component and execute the function on storage level.

The 1st  option sounds easier, but it becomes a challenge when we need a cross-module (cross-table) update, i.e. what we need  for `holdings_record` table - when a holding is updated we need to update items.

The main problem is that we can not call another fake module within current (e.g. when we're serving update for a holding we can not fetch all items in this method) this is because Vert.x event bus thread is currently busy. 

In order to get rid of this issue we will have to modify fake modules to use async style (use CompletableFutures, and the trigger function will become much complicated) and this is pretty-much the same effort that extracting storage to a separate component.

So now storage is a singleton, shared across all the fake modules. I guess this approach will give us a bit more flexibility in the future.  

What do you think?

src/test/java/api/ApiTestSuite.java
marcjohnson-kint: What do you mean by clean up?

marcjohnson-kint:
> @marcjohnson-kint, Right now we have set of FakeStorageModules and each module encapsulates it's own Map which represents a storage.
> 
> There are actually two options to introduce this triggers-like logic:
> 
> 1. We can provide a function to a fake, and it will call it on create/update.
> 2. Extract the storage as a separate component and execute the function on storage level.
> 
> The 1st option sounds easier, but it becomes a challenge when we need a cross-module (cross-table) update, i.e. what we need for `holdings_record` table - when a holding is updated we need to update items.

I can understand that challenge of different endpoints need access to records from other APIs. Ad how that leads to the need to share storage.

I think there is a distinction between sharing storage and how the logic itself is expressed. 

> The main problem is that we can not call another fake module within current (e.g. when we're serving update for a holding we can not fetch all items in this method) this is because Vert.x event bus thread is currently busy.

Please can you expand upon what you mean by the vert.x event bus is busy.

> In order to get rid of this issue we will have to modify fake modules to use async style (use CompletableFutures, and the trigger function will become much complicated) and this is pretty-much the same effort that extracting storage to a separate component.

I don't think I follow, why do the fake modules need to change to an asynchronous style?
 
> So now storage is a singleton, shared across all the fake modules. I guess this approach will give us a bit more flexibility in the future.

I can understand the need to share storage. I think I'm struggling to understand the design for it.

I'd be tempted to define storage within the `FakeOkapi` rather than as a global singleton. This would remove the need for static operations and hide it within the fakes rather than it leaking into more of the test setup.

I don't necessarily understand the other classes involved. I find the use of a singleton storage makes it harder to follow how it is used for coordination, especially it being hidden deep in the triggers.

Could a simpler alternative be that the function defined to execute when a record is created or updated could take a reference to the shared storage?

bohdan-suprun:
> Please can you expand upon what you mean by the vert.x event bus is busy.

So, as per Vert.x documentation there is an event loop thread which dispatches and serves every incoming event (request). Moreover, only one event loop thread is attached to a Verticle (FakeOkapi in our case). Since we're not using async style (read as we don't use CompletableFutures callbacks, but use get() with a timeout instead) in unit tests. And then following happens:
* We send holding update request;
* FakeOkapi event loop thread accepts request and dispatches it to Holdings update handler;
* Holdings handler try to call the pre-process function:
  * Send request to the FakeOkapi to get Items that have `holdingsRecordId` == `holdings.id`
  * And here we get the error - the event loop thread is currently busy with holdings update, so it can not serve the item get request, and this request is put into the event queue and will be processed when the holdings update will be finished - what is useless for us.

> I don't think I follow, why do the fake modules need to change to an asynchronous style?

In order to solve the issue I've described above, we have to build create/update handlers and the 'triggers' to use CompletableFutures and instead of using blocking style - set-up callbacks.

>  I find the use of a singleton storage makes it harder to follow how it is used for coordination, especially it being hidden deep in the triggers.

Well, it is exactly the same approach as we used in the storage module (i.e. the business logic in triggers).

mreno-EBSCO:
> In order to solve the issue I've described above, we have to build create/update handlers and the 'triggers' to use CompletableFutures and instead of using blocking style - set-up callbacks.

Do we though? I think reproducing another modules logic for managing values in this module is prone to error. This is better suited as an integration test.

marcjohnson-kint:
> > In order to solve the issue I've described above, we have to build create/update handlers and the 'triggers' to use CompletableFutures and instead of using blocking style - set-up callbacks.
> 
> Do we though? I think reproducing another modules logic for managing values in this module is prone to error. This is better suited as an integration test.

@mreno-EBSCO Can you expand upon the approach you are describing?

Given that mod-inventory will be entirely dependent upon the implementation of the `holding-storage` and `item-storage` interfaces to determine the effective location, how would you propose we check that this is used appropriately within mod-inventory?

marcjohnson-kint:
@bohdan-suprun 

> Well, it is exactly the same approach as we used in the storage module (i.e. the business logic in triggers).

I would prefer we were coupled to this as little as possible in the downstream modules. From that perspective, I think we want to depend upon the behaviour without necessarily taking on the complexities of how that is implemented in the reference storage module itself.

Does that make sense?

mreno-EBSCO:
> @mreno-EBSCO Can you expand upon the approach you are describing?
> 
> Given that mod-inventory will be entirely dependent upon the implementation of the `holding-storage` and `item-storage` interfaces to determine the effective location, how would you propose we check that this is used appropriately within mod-inventory?

Does it matter what it is in mod-inventory? It is just a UUID. Why do we need all the business logic in the tests to validate something that cannot/should not be determined here? I suppose, if we pass in various UUIDs for locations in the item/holding, we would expect one of those UUIDs to be returned as the effective location. In the future, perhaps that may not be true. Does it matter to mod-inventory what the effective location ID is or that it is just set to some location ID? I am all for trying to ensure that we are doing the right thing, but, unless I am missing something, this module doesn't know what the right thing is.

During integration testing, we know what the expectations for effective ID are, so we can write a test that will expect the right thing returned via mod-inventory.

I am probably making too big of a deal of this. I am just concerned about the maintainability of the test support code as it grows and that we'll be debugging the test support code more than the actual code when tests break.

It'd be nice if we could solve this with simple unit tests and service mocks, but I think I may be oversimplifying this.

bohdan-suprun:
@mreno-EBSCO, I agree with your assumptions. I don't really know why we have to mimic the `effectiveLocationId` logic in the mod-inventory. The real purpose of the test I wrote is that: if there is `effectiveLocationId` property in the mod-inventory-storage's response that it must be properly carried through all the mod-inventory layers and then transformed to the:
```
{
  "id": "UUID",
  "name": "Location-name" 
}
```
structure and that's it. 

I also agree that in this case integration tests will suite better (from early errors detecting point of view).

I also fine with encapsulation storage in the FakeOkapi instead of making it a singleton.

@marcjohnson-kint what do you think?

marcjohnson-kint:
> > @mreno-EBSCO Can you expand upon the approach you are describing?
> > Given that mod-inventory will be entirely dependent upon the implementation of the `holding-storage` and `item-storage` interfaces to determine the effective location, how would you propose we check that this is used appropriately within mod-inventory?
> 
> Does it matter what it is in mod-inventory? It is just a UUID. Why do we need all the business logic in the tests to validate something that cannot/should not be determined here? 

> I suppose, if we pass in various UUIDs for locations in the item/holding, we would expect one of those UUIDs to be returned as the effective location. In the future, perhaps that may not be true. 

That's true, there is a challenge about incidental coupling here.

This is a general challenge with integration, any integration tests have to bind to implementations rather than only interfaces.

> Does it matter to mod-inventory what the effective location ID is or that it is just set to some location ID? I am all for trying to ensure that we are doing the right thing, but, unless I am missing something, this module doesn't know what the right thing is.

For mod-inventory, I don't think there is a need for tests that check the fake creates the correct calculation. Only that demonstrate that the property is properly propagated to the client.

I am partly thinking ahead to how this could be done in mod-circulation, where we might care about this more. At the very least, we care that the decision is deterministic.

Thinking about it, I could be happy with the effective location being a simpler derivation than the one implemented in storage. A trade-off with that is further diminishment of using these tests for verifying integration, and therefore the need for separate tests to do that. Maybe I need to abandon that notion for the time being :-(

> During integration testing, we know what the expectations for effective ID are, so we can write a test that will expect the right thing returned via mod-inventory.

What are you referring to as integration tests? As far as I'm aware the project does not currently have any that are actively maintained (I'm aware EBSCO and maybe EPAM used to maintain a suite of separate tests).

> I am probably making too big of a deal of this. I am just concerned about the maintainability of the test support code as it grows and that we'll be debugging the test support code more than the actual code when tests break.

I think this is an important topic. I'm not aware of an approach that doesn't have trade offs. Increased implicit coupling is inherent in the approach taken to derived properties.

> It'd be nice if we could solve this with simple unit tests and service mocks, but I think I may be oversimplifying this.

I agree. That would be nice, both mod-circulation and mod-inventory need significant investment in improving the ability to test these kinds of aspects in smaller units. 

The fakes started out simply, even though they were at the API level. They have steadily grown in complexity as we increase the sophistication of storage modules and encounter more integration issues.

marcjohnson-kint:
> @mreno-EBSCO, I agree with your assumptions. I don't really understand why we have to mimic the `effectiveLocationId` logic in the mod-inventory. The real purpose of the test I wrote is that: if there is `effectiveLocationId` property in the mod-inventory-storage's response that it must be properly carried through all the mod-inventory layers and then transformed to the:
> 
> ```
> {
>   "id": "UUID",
>   "name": "Location-name" 
> }
> ```
> 
> structure and that's it.
> 
> I also agree that in this case integration tests will suite better (from early errors detecting point of view).
> 
> I also fine with encapsulation storage in the FakeOkapi instead of making it a singleton.
> 
> @marcjohnson-kint what do you think?

The primary reason for mimicking the behaviour is because I'm hanging on to the idea that these API tests can be used for integration as well :-( Maybe I need to abandon that notion and accept that any deterministic fake behaviour is ok.

If we do that, I think we really need to think about how FOLIO can sustainable do integration testing.

mreno-EBSCO:
> What are you referring to as integration tests? As far as I'm aware the project does not currently have any that are actively maintained (I'm aware EBSCO and maybe EPAM used to maintain a suite of separate tests).

Yeah, maybe integration test is a bit too generic here. I was thinking more of a workflow test, an overall test of the inventory functionality where the tests have general knowledge of all domains. I was not thinking specifically of the integration tests that EBSCO created and EPAM uses, which are simple postman tests, but something like that could be used. Ultimately, they'd need to be created, which is the biggest hurdle at this point.

marcjohnson-kint:
> > What are you referring to as integration tests? As far as I'm aware the project does not currently have any that are actively maintained (I'm aware EBSCO and maybe EPAM used to maintain a suite of separate tests).
> 
> Yeah, maybe integration test is a bit too generic here. I was thinking more of a workflow test, an overall test of the inventory functionality where the tests have general knowledge of all domains. I was not thinking specifically of the integration tests that EBSCO created and EPAM uses, which are simple postman tests, but something like that could be used. Ultimately, they'd need to be created, which is the biggest hurdle at this point.

I agree that there is a need for integration tests that demonstrate that parts of the system operate together. Beyond the UI regression tests, I don't believe we really have any.

Given that, what would we like to do for this change?

Should I abandon the notion that the API tests can also act as integration tests? In which case, the effective location can use a simpler derivation in the fakes than in the reference storage module implementation.

I think this is separate to how we manage the storage in the fakes itself. Though, if we choose the simpler derivation, we can potentially (a little cheekily) use the item level location properties only, and avoid the need to connect the two records in the fakes.

bohdan-suprun:
@marcjohnson-kint, sounds good for me. We can do following then:
* add a functions to the fake module that will be executed on entity update/create.
* For items update/create the functions will just pick either temp or perm locations.

Is it fine?

marcjohnson-kint:
> @marcjohnson-kint, sounds good for me. We can do following then:
> 
> * add a functions to the fake module that will be executed on entity update/create.
> * For items update/create the functions will just pick either temp or perm locations.

> Is it fine?

This could work for me. I think if we make sure that all of the examples in mod-inventory define an item level location, and so this would match up with the real behaviour, it might be possible to stay consistent without the need for the added complexity of mimicking the reference behaviour (and therefore the need for accessing records across endpoints).

I think it is important that the decision about which property is chosen is deterministic.

@mreno-EBSCO What do you think?

I think we might need to revisit this in mod-circulation, where decisions are made based upon location.

mreno-EBSCO:
Sounds good to me.

src/test/java/api/ApiTestSuite.java
bohdan-suprun: outdated

bohdan-suprun:
@mreno-EBSCO, @marcjohnson-kint, it is done now.

src/test/java/support/fakes/FakeOkapi.java
```diff
@@ -97,6 +98,8 @@ private void registerFakeItemsStorageModule(Router router) {
       .withRootPath("/item-storage/items")
       .withRequiredProperties("materialTypeId", "permanentLoanTypeId")
       .withDefault("status", new JsonObject().put("name", "Available"))
+      .withCreateProcessors(new ItemCreateUpdateEffectiveLocationProcessor())
```
marcjohnson-kint: Why does this code need to support multiple processors when the only use case we have has a single processor?

src/test/java/support/fakes/processor/item/ItemCreateUpdateEffectiveLocationProcessor.java
```diff
@@ -0,0 +1,26 @@
...
+
+
+public class ItemCreateUpdateEffectiveLocationProcessor implements RecordCreateProcessor, RecordUpdateProcessor {
+  @Override
+  public void onCreate(JsonObject newEntity) {
+    onUpdate(null, newEntity);
+  }
+
+  @Override
+  public void onUpdate(JsonObject oldEntity, JsonObject newEntity) {
```
marcjohnson-kint: Why is `oldEntity` needed here, it doesn't appear to be used?

src/test/java/support/fakes/FakeOkapi.java
```diff
@@ -97,6 +98,8 @@ private void registerFakeItemsStorageModule(Router router) {
...
+      .withUpdateProcessors(new ItemCreateUpdateEffectiveLocationProcessor())
```
marcjohnson-kint: I think I might be missing something with the processors model.

I understand the intent of this code is to take a record representation and process it in some way to produce a new record representation, in this case, add an `effectiveLocationId` property.

This could be achieved by using a Function<JsonObject, JsonObject> where the input is the current representation and the return value is the representation after processing.

I like the idea of describing this in an interface definition rather than a raw functional type. IIRC it is possible to define an instance of a single method interface implicitly using a method reference or similar. Could this be simplified to a single interface with an implicit implementation?

PR #158, opened 2019-11-12
MODINV-173: Add last check in properties to item representation

mrdpopov:
## Purpose
Show most recent check in on item record.
Resolves: MODINV-173
See MODINVSTOR-386 and CIRC-522 for reference.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
-      "version": "9.4",
+      "version": "9.5",
       "handlers": [
         {
           "methods": ["GET"],
@@ -192,7 +192,7 @@
   "requires": [
     {
       "id": "item-storage",
-      "version": "7.5"
+      "version": "7.7"
     },
     {
       "id": "instance-storage",
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -26,7 +26,7 @@
     "discoverySuppress": {
       "type": "boolean",
       "description": "Records the fact that the record should not be displayed in a discovery system"
-    },    
+    },
     "title": {
       "type": "string",
       "description": "Resouce title (read only, inherited from associated instance record)",
@@ -395,6 +395,28 @@
       "id": "tags",
       "type": "object",
       "$ref": "raml-util/schemas/tags.schema"
+    },
+    "lastCheckIn": {
+      "type": "object",
+      "additionalProperties": false,
+      "description": "Information about when an item was last scanned in the Inventory app.",
+      "properties": {
+        "dateTime": {
+          "type": "string",
+          "description": "Date and time of the last check in of the item.",
+          "format": "date-time"
+        },
+        "servicePointId": {
+          "type": "string",
+          "description": "Service point ID being used by a staff member when item was scanned in Check in app.",
+          "$ref": "uuid.json"
+        },
+        "staffMemberId": {
+          "type": "string",
+          "description": "ID a staff member who scanned the item",
+          "$ref": "uuid.json"
+        }
+      }
     }
   },
   "additionalProperties": false,
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -399,22 +399,26 @@
     "lastCheckIn": {
       "type": "object",
       "additionalProperties": false,
-      "description": "Information about when an item was last scanned in the Inventory app.",
+      "description": "Information about when an item was last checked in",
+      "readonly": true,
       "properties": {
         "dateTime": {
           "type": "string",
           "description": "Date and time of the last check in of the item.",
-          "format": "date-time"
+          "format": "date-time",
+          "readonly": true
         },
         "servicePointId": {
           "type": "string",
           "description": "Service point ID being used by a staff member when item was scanned in Check in app.",
-          "$ref": "uuid.json"
+          "$ref": "uuid.json",
+          "readonly": true
         },
         "staffMemberId": {
           "type": "string",
           "description": "ID a staff member who scanned the item",
-          "$ref": "uuid.json"
+          "$ref": "uuid.json",
+          "readonly": true
         }
       }
     }
--- a/src/main/java/org/folio/inventory/domain/ingest/IngestMessageProcessor.java
+++ b/src/main/java/org/folio/inventory/domain/ingest/IngestMessageProcessor.java
@@ -14,7 +14,6 @@ import org.folio.inventory.domain.instances.Instance;
 import org.folio.inventory.domain.instances.InstanceCollection;
 import org.folio.inventory.domain.items.Item;
 import org.folio.inventory.domain.items.ItemCollection;
-import org.folio.inventory.domain.items.LastCheckIn;
 import org.folio.inventory.domain.items.Status;
 import org.folio.inventory.resources.ingest.IngestJob;
 import org.folio.inventory.resources.ingest.IngestJobState;
@@ -147,9 +146,7 @@ public class IngestMessageProcessor {
                 ? loanTypes.getString("Can Circulate")
                 : loanTypes.getString("Can circulate"),
                null)
-                    .withBarcode(record.getString("barcode"))
-                .withLastCheckIn(
-                    LastCheckIn.from(record.getJsonObject("lastCheckIn")));
+                    .withBarcode(record.getString("barcode"));
         })
         .forEach(item -> itemCollection.add(item, allItems.receive(),
           failure -> log.error("Item processing failed: " + failure.getReason()))));
--- a/src/main/java/org/folio/inventory/domain/ingest/IngestMessageProcessor.java
+++ b/src/main/java/org/folio/inventory/domain/ingest/IngestMessageProcessor.java
@@ -14,6 +14,7 @@ import org.folio.inventory.domain.instances.Instance;
 import org.folio.inventory.domain.instances.InstanceCollection;
 import org.folio.inventory.domain.items.Item;
 import org.folio.inventory.domain.items.ItemCollection;
+import org.folio.inventory.domain.items.LastCheckIn;
 import org.folio.inventory.domain.items.Status;
 import org.folio.inventory.resources.ingest.IngestJob;
 import org.folio.inventory.resources.ingest.IngestJobState;
@@ -146,7 +147,9 @@ public class IngestMessageProcessor {
                 ? loanTypes.getString("Can Circulate")
                 : loanTypes.getString("Can circulate"),
                null)
-                    .withBarcode(record.getString("barcode"));
+                    .withBarcode(record.getString("barcode"))
+                .withLastCheckIn(
+                    LastCheckIn.from(record.getJsonObject("lastCheckIn")));
         })
         .forEach(item -> itemCollection.add(item, allItems.receive(),
           failure -> log.error("Item processing failed: " + failure.getReason()))));
--- a/src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
+++ b/src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
@@ -8,8 +8,6 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
 import java.util.Date;
-import javax.validation.constraints.Pattern;
-import org.joda.time.DateTime;
 import org.joda.time.format.ISODateTimeFormat;
 
 /**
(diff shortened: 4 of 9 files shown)


src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
```diff
@@ -0,0 +1,137 @@
...
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder( {
+    "dateTime",
+    "servicePointId",
+    "staffMemberId"
+})
+public class LastCheckIn {
+
+    static final LastCheckIn EMPTY = new LastCheckIn();
```
mreno-EBSCO: Can we just use `null` instead of an `EMPTY` object? I don't see what this `EMPTY` object gives us over `null` here. Also, this class is mutable, so the `EMPTY` object could be modified, which doesn't seem right if this is to be a constant.

mreno-EBSCO:
This PR needs unit tests. There is also a sonar reported code smell for an unused import and security problems with UUID regexs. Please increase code coverage and fix the smell. Before submitting a PR, take a look at the sonar analysis for your branch build to see if the branch passes the quality gates. If it doesn't, fix the problems before submitting a PR.

For example: https://sonarcloud.io/dashboard?id=org.folio%3Amod-inventory&branch=MODINV-173&resolved=false

src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
mrdpopov: Sure.

mrdpopov:
@mreno-EBSCO Thank you for your review. This will be definately taken into account in future. As far as this PR goes, the test coverage has been increased.

src/main/java/org/folio/inventory/domain/ingest/IngestMessageProcessor.java
```diff
@@ -146,7 +147,9 @@ private void processRecordsMessage(Message<JsonObject> message, final EventBus e
                 ? loanTypes.getString("Can Circulate")
                 : loanTypes.getString("Can circulate"),
                null)
-                    .withBarcode(record.getString("barcode"));
+                    .withBarcode(record.getString("barcode"))
+                .withLastCheckIn(
```
marcjohnson-kint: Why would an ingested record ever have a last check in?

src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
```diff
@@ -0,0 +1,128 @@
...
+ * Information about when an item was last scanned in the Inventory app.
+ *
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder( {
+    "dateTime",
+    "servicePointId",
+    "staffMemberId"
+})
+public class LastCheckIn {
```
marcjohnson-kint: Is this generated code?

If not, why does this class need to mutable? Why does it need JSON property attributes? Is all of that added only to avoid implementing the `toJson` method manually?

mrdpopov: It is generated by RMB and copied here.

marcjohnson-kint: I would prefer we didn't copy and paste generated code from other modules without changing it.

Please can this be amended to better fit within the style of mod-inventory / mod-circulation.

Personally, I would do that by making it immutable (with constructor parameters) and manually implementing the `toJson` method.

src/main/java/org/folio/inventory/domain/ingest/IngestMessageProcessor.java
mrdpopov: It was added as a `safe` measure... It will not break/change anything. Do you think it should be removed?

src/main/java/org/folio/inventory/resources/ItemRepresentation.java
```diff
@@ -166,6 +165,10 @@ private JsonObject toJson(
       log.warn(String.format("Failed to create self link for item: %s", e.toString()));
     }
 
+    if (item.getLastCheckIn() != null) {
```
VRohach: I think that it will better to use "Objects.nonNull()" here. What do you think?

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1226,6 +1230,55 @@ public void canSearchItemsByLocation() throws Exception {
...
+    itemsStorageClient.create(readingRoomItem);
+
+    JsonObject readingRoomItems = findItems("effectiveLocationId=" + getReadingRoomLocation());
+
+    JsonObject actualItem = readingRoomItems.getJsonArray("items").getJsonObject(0);
+    assertThat(actualItem, is(notNullValue()));
+    assertThat(actualItem.getJsonObject(LAST_CHECK_IN_FIELD), is(nullValue()));
+  }
+
+
```
VRohach: Could you please remove this extra line?

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1226,6 +1230,55 @@ public void canSearchItemsByLocation() throws Exception {
       is(mainLibraryItem.getString("id")));
   }
 
+  @Test
+  public void itemHasLastCheckInPropertiesWhenTheyAreSet() throws Exception {
```
VRohach: Can we set here a more specific exception? Not common 'Exception'? (And in the next test)

src/main/java/org/folio/inventory/resources/ItemRepresentation.java
mrdpopov: Why?

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1226,6 +1230,55 @@ public void canSearchItemsByLocation() throws Exception {
...
+    JsonObject readingRoomItem = new JsonObject()
+      .put("id", UUID.randomUUID().toString())
+      .put("holdingsRecordId", createInstanceAndHolding().toString())
+      .put("materialTypeId", getDvdMaterialType())
+      .put("permanentLoanTypeId", getCanCirculateLoanType())
+      .put("permanentLocationId", getMainLibraryLocation())
+      .put("temporaryLocationId", getReadingRoomLocation());
+
+    JsonObject lastCheckInObj = new JsonObject();
+    lastCheckInObj.put(USER_ID_FIELD, UUID.randomUUID().toString());
```
VRohach: I think it will be better to use 'write' method here. Does it make sense?

src/main/java/org/folio/inventory/resources/ItemRepresentation.java
mrdpopov: API Note: This method exists to be used as a Predicate, filter(Objects::isNull)

VRohach: I think that we can avoid here NullPointerException if 'item' will be null in this way.

VRohach: Oh, really. Thank you!

src/test/java/api/items/ItemApiExamples.java
mrdpopov: No as there is no 'write' method in this module.

src/test/java/api/items/ItemApiExamples.java
mrdpopov: Sure.

src/test/java/api/items/ItemApiExamples.java
mrdpopov: Yep

src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
mrdpopov: Done.

src/main/java/org/folio/inventory/domain/ingest/IngestMessageProcessor.java
mrdpopov: Removed.

src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
marcjohnson-kint: Thank you for changing the implementation of this class. I think that means that the comments and the annotations can also be removed?

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1226,6 +1230,59 @@ public void canSearchItemsByLocation() throws Exception {
...
+    itemsStorageClient.create(readingRoomItem);
+
+    JsonObject readingRoomItems = findItems("effectiveLocationId=" + getReadingRoomLocation());
+    JsonObject actualItem = readingRoomItems.getJsonArray("items").getJsonObject(0);
+    JsonObject actualLastCheckIn = actualItem.getJsonObject(LAST_CHECK_IN_FIELD);
+
+    assertThat(actualLastCheckIn.getString(DATETIME_FIELD), is(notNullValue()));
+    assertThat(actualLastCheckIn.getString(SERVICE_POINT_FIELD), is(notNullValue()));
+    assertThat(actualLastCheckIn.getString(USER_ID_FIELD), is(notNullValue()));
+
```
marcjohnson-kint: I think this blank line can be removed

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1226,6 +1230,59 @@ public void canSearchItemsByLocation() throws Exception {
...
+    itemsStorageClient.create(readingRoomItem);
+
+    JsonObject readingRoomItems = findItems("effectiveLocationId=" + getReadingRoomLocation());
+    JsonObject actualItem = readingRoomItems.getJsonArray("items").getJsonObject(0);
+    JsonObject actualLastCheckIn = actualItem.getJsonObject(LAST_CHECK_IN_FIELD);
+
+    assertThat(actualLastCheckIn.getString(DATETIME_FIELD), is(notNullValue()));
+    assertThat(actualLastCheckIn.getString(SERVICE_POINT_FIELD), is(notNullValue()));
+    assertThat(actualLastCheckIn.getString(USER_ID_FIELD), is(notNullValue()));
```
marcjohnson-kint: Given that the values are explicitly set in storage, can they be explicitly checked here?

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
-      "version": "9.4",
+      "version": "9.5",
```
marcjohnson-kint: There doesn't appear to be any changes to the JSON schema, should these new properties be added?

src/test/java/api/items/ItemApiExamples.java
mrdpopov: Sure.

src/test/java/api/items/ItemApiExamples.java
mrdpopov: Sure.

descriptors/ModuleDescriptor-template.json
mrdpopov: Sure.

src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
mrdpopov: Sure

ramls/item.json
```diff
@@ -395,6 +395,28 @@
       "id": "tags",
       "type": "object",
       "$ref": "raml-util/schemas/tags.schema"
+    },
+    "lastCheckIn": {
+      "type": "object",
+      "additionalProperties": false,
+      "description": "Information about when an item was last scanned in the Inventory app.",
```
marcjohnson-kint:
```suggestion
      "description": "Information about when an item was last checked in",
```

ramls/item.json
```diff
@@ -395,6 +395,28 @@
       "id": "tags",
       "type": "object",
       "$ref": "raml-util/schemas/tags.schema"
+    },
+    "lastCheckIn": {
```
marcjohnson-kint: Are these properties writable using this API?

src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
```diff
@@ -0,0 +1,48 @@
...
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+import io.vertx.core.json.JsonObject;
+import java.util.UUID;
+import org.joda.time.DateTime;
+import org.joda.time.format.ISODateTimeFormat;
+
+/**
+ * Information about when an item was last scanned in the Inventory app.
+ *
+ */
```
marcjohnson-kint: I think these comments can be removed

ramls/item.json
mrdpopov: Done.

ramls/item.json
mrdpopov: I guess not, thank you for catching.

src/main/java/org/folio/inventory/domain/items/LastCheckIn.java
mrdpopov: Done.

mrdpopov:
@marc-johnson  is this ready for merging?

PR #160, opened 2019-11-25
Modinv 174 effective call number components

rladdusaw:
Resolves the implementation portion of  MODINV-176.
Updated versions for inventory and item storage.
Implemented EffectiveCallNumberComponents as an immutable class.

Make fake storage record pre-processors to be async (consistent with circulation ones).


src/test/java/api/items/ItemApiExamples.java
```diff
@@ -403,6 +415,9 @@ public void canUpdateExistingItem()
       .forHolding(holdingId)
       .withBarcode("645398607547")
       .canCirculate()
+      .withItemLevelCallNumber(CALL_NUMBER)
+      .withItemLevelCallNumberSuffix(CALL_NUMBER_SUFFIX)
+      .withItemLevelCallNumberPrefix(CALL_NUMBER_PREFIX)
```
marcjohnson-kint: If these properties are not being changed, why do they need to be included in this test?

bohdan-suprun: I just realized that this test is redundant, because PUT does not return any entity, so we just need to tests POST and GET.

PR #161, opened 2019-11-29
MODINV-162, MODINV-160: Handle HRID for instance and item.

bohdan-suprun:
https://issues.folio.org/browse/MODINV-160 - HRID handling in Inventory for Instance records.
https://issues.folio.org/browse/MODINV-162 - HRID handling in Inventory for Item records.

### Changes:
* Added Item/Instance fake pre-processors to generate a HRID if nothing has specified;
* Disallow changing of hrid for item and instance;
* Added unit tests.

### DIFF
--- a/src/test/java/api/InstancesApiExamples.java
+++ b/src/test/java/api/InstancesApiExamples.java
@@ -20,6 +20,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
+import static support.matchers.ResponseMatchers.hasValidationError;
 
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -150,10 +151,12 @@ public class InstancesApiExamples extends ApiTests {
     dublinCoreContextLinkRespectsWayResourceWasReached(createdInstance);
     selfLinkRespectsWayResourceWasReached(createdInstance);
     selfLinkShouldBeReachable(createdInstance);
+
+    assertThat(createdInstance.getString("hrid"), notNullValue());
   }
 
   @Test
-  public void canCreateAnInstanceWithAnID()
+  public void canCreateAnInstanceWithAnIDAndHrid()
     throws InterruptedException,
     MalformedURLException,
     TimeoutException,
@@ -161,9 +164,11 @@ public class InstancesApiExamples extends ApiTests {
     JsonLdError {
 
     String instanceId = UUID.randomUUID().toString();
+    final String hrid = "in777";
 
     JsonObject newInstanceRequest = new JsonObject()
       .put("id", instanceId)
+      .put("hrid", hrid)
       .put("title", "Long Way to a Small Angry Planet")
       .put("identifiers", new JsonArray().add(new JsonObject()
       .put("identifierTypeId", ApiTestSuite.getIsbnIdentifierType())
@@ -221,6 +226,7 @@ public class InstancesApiExamples extends ApiTests {
     dublinCoreContextLinkRespectsWayResourceWasReached(createdInstance);
     selfLinkRespectsWayResourceWasReached(createdInstance);
     selfLinkShouldBeReachable(createdInstance);
+    assertThat(createdInstance.getString("hrid"), is(hrid));
   }
 
   @Test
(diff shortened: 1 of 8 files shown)


src/test/java/api/InstancesApiExamples.java
```diff
@@ -150,20 +151,24 @@ public void canCreateAnInstance()
     dublinCoreContextLinkRespectsWayResourceWasReached(createdInstance);
     selfLinkRespectsWayResourceWasReached(createdInstance);
     selfLinkShouldBeReachable(createdInstance);
+
+    assertThat(createdInstance.getString("hrid"), notNullValue());
```
marcjohnson-kint: I think this test should be renamed to `canCreateInstanceWithoutAnIDAndHRID`

src/test/java/api/InstancesApiExamples.java
```diff
@@ -150,20 +151,24 @@ public void canCreateAnInstance()
...
   }
 
   @Test
-  public void canCreateAnInstanceWithAnID()
+  public void canCreateAnInstanceWithAnIDAndHrid()
```
marcjohnson-kint:
```suggestion
  public void canCreateAnInstanceWithAnIDAndHRID()
```

src/test/java/api/InstancesApiExamples.java
```diff
@@ -826,6 +832,31 @@ public void cannotFindAnUnknownInstance()
     assertThat(getResponse.getStatusCode(), is(404));
   }
 
+  @Test
+  public void cannotUpdateHrid() throws Exception {
```
marcjohnson-kint: What happens if the HRID is not specified in the update?

marcjohnson-kint:
```suggestion
  public void cannotUpdateHRID() throws Exception {
```

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -144,16 +145,18 @@ public void canCreateAnItem()
 
     selfLinkRespectsWayResourceWasReached(createdItem);
     selfLinkShouldBeReachable(createdItem);
+    assertThat(createdItem.getString("hrid"), notNullValue());
   }
 
   @Test
-  public void canCreateItemWithAnID()
+  public void canCreateItemWithAnIDAndHrid()
```
marcjohnson-kint:
```suggestion
  public void canCreateItemWithAnIDAndHRID()
```

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1311,6 +1317,40 @@ public void itemHasNoLastCheckInPropertiesWhenNotSet() throws Exception {
     assertThat(actualItem.getJsonObject(LAST_CHECK_IN_FIELD), is(nullValue()));
   }
 
+  @Test
+  public void cannotUpdateHrid() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void cannotChangeHRID() throws Exception {
```

src/test/java/api/InstancesApiExamples.java
bohdan-suprun: The update will be rejected.

marcjohnson-kint: Please can a test be added to demonstrate that a HRID cannot be removed (for both instance and item).

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -78,7 +79,7 @@ public ItemApiExamples() throws MalformedURLException {
   }
 
   @Test
-  public void canCreateAnItem()
+  public void canCreateAnItemWithoutIdAndHRID()
```
marcjohnson-kint:
```suggestion
  public void canCreateAnItemWithoutIDAndHRID()
```

marcjohnson-kint:
Does the instance batch API allow for instances to be updated?

bohdan-suprun:
> Does the instance batch API allow for instances to be updated?

It seems not, there is only create batch method. It executes the storage batch method + saves instance relationships.

src/test/java/api/items/ItemApiExamples.java
marcjohnson-kint: The `Id` needs capitalising to `ID`

PR #164, opened 2019-12-11
MODINV-177: Forward call number type ID from item storage.

bohdan-suprun:
https://issues.folio.org/browse/MODINV-177: Forward effective call number typeId property.

* Forward the `effectiveCallNumberComponents.typeId` property from item;
* Update the inventory API version.

**Have to be merged after** https://github.com/folio-org/mod-inventory-storage/pull/372

### DIFF
--- a/src/main/java/org/folio/inventory/domain/items/EffectiveCallNumberComponents.java
+++ b/src/main/java/org/folio/inventory/domain/items/EffectiveCallNumberComponents.java
@@ -6,11 +6,14 @@ public class EffectiveCallNumberComponents {
     private String callNumber;
     private String prefix;
     private String suffix;
+    private String typeId;
 
-    public EffectiveCallNumberComponents(String callNumber, String prefix, String suffix) {
+    public EffectiveCallNumberComponents(String callNumber, String prefix,
+                                         String suffix, String typeId) {
       this.callNumber = callNumber;
       this.prefix = prefix;
       this.suffix = suffix;
+      this.typeId = typeId;
     }
 
     public static EffectiveCallNumberComponents from(JsonObject representation) {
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/inventory/domain/items/EffectiveCallNumberComponents.java
```diff
@@ -6,11 +6,14 @@
     private String callNumber;
     private String prefix;
     private String suffix;
+    private String typeId;
 
-    public EffectiveCallNumberComponents(String callNumber, String prefix, String suffix) {
+    public EffectiveCallNumberComponents(String callNumber, String prefix,
+                                         String suffix, String typeId) {
```
marcjohnson-kint: I personally dislike this alignment of the second line of parameters, because it sacrifices so much horizontal space (which is what wrapping cares about) and is harder to maintain.

I’m aware IntelliJ does this by default, and I haven’t figured out how to change that.

What are other folks thoughts?

(I’m not saying you need to change this specific case, it’s just another vagary in code formatting)

mreno-EBSCO: I don't like horizontal alignment as well. If we change the name, now we need to realign. I prefer 4 space indentation on the next line.

This is the approach Google style uses: [4.5.2 Indent continuation lines at least +4 spaces](https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent). They do have a section on horizontal alignment: [4.6.3 Horizontal alignment: never required](https://google.github.io/styleguide/javaguide.html#s4.6.3-horizontal-alignment). Note their "tip".

bohdan-suprun: Yes, I don't like it too.

PR #165, opened 2019-12-26
[BREAKING] MODINV-189: item-storage 8.0 updates

bohdan-suprun:
https://issues.folio.org/browse/MODINV-189: Item-storage 8.0 breaking changes

This story includes changes required by breaking stories:

* MODINVSTOR-283 Restrict item states
* MODINVSTOR-332 Item: Change array copyNumbers to simple property copyNumber (breaking)
* MODINVSTOR-392 Make item.status.date read only field.

**Must be merged after:** https://github.com/folio-org/mod-inventory-storage/pull/382

### DIFF
--- a/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
+++ b/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
@@ -15,7 +15,6 @@ public enum ItemStatusName {
   MISSING("Missing"),
   ON_ORDER("On order"),
   PAGED("Paged"),
-  WITHDRAWN("Withdrawn"),
   DECLARED_LOST("Declared lost"),
   RECEIVED("Received"),
   ORDER_CLOSED("Order closed");
--- a/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
+++ b/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
@@ -8,6 +8,7 @@ import java.util.Map;
 public enum ItemStatusName {
   AVAILABLE("Available"),
   AWAITING_PICKUP("Awaiting pickup"),
+  AWAITING_DELIVERY("Awaiting delivery"),
   CHECKED_OUT("Checked out"),
   IN_PROCESS("In process"),
   IN_TRANSIT("In transit"),
--- a/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
+++ b/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
@@ -1,0 +1,52 @@
+package org.folio.inventory.domain.items;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public enum ItemStatusName {
+  AVAILABLE("Available"),
+  AWAITING_PICKUP("Awaiting pickup"),
+  CHECKED_OUT("Checked out"),
+  IN_PROCESS("In process"),
+  IN_TRANSIT("In transit"),
+  MISSING("Missing"),
+  ON_ORDER("On order"),
+  PAGED("Paged"),
+  WITHDRAWN("Withdrawn"),
+  DECLARED_LOST("Declared lost");
+
+  private static final Map<String, ItemStatusName> VALUE_TO_INSTANCE_MAP = initValueToInstanceMap();
+  private final String value;
+
+  ItemStatusName(String value) {
+    this.value = value;
+  }
+
+  @Override
+  public String toString() {
+    return this.value;
+  }
+
+  public String value() {
+    return this.value;
+  }
+
+  public static ItemStatusName forName(String value) {
+    return VALUE_TO_INSTANCE_MAP.get(value);
+  }
+
+  public static boolean hasKnownValue(String value) {
+    return VALUE_TO_INSTANCE_MAP.containsKey(value);
+  }
+
+  private static Map<String, ItemStatusName> initValueToInstanceMap() {
+    final Map<String, ItemStatusName> instances = new HashMap<>();
+
+    Arrays.stream(values())
+      .forEach(itemStatusName -> instances.put(itemStatusName.value(), itemStatusName));
+
+    return Collections.unmodifiableMap(instances);
+  }
+}
--- a/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
+++ b/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
@@ -16,7 +16,6 @@ public enum ItemStatusName {
   ON_ORDER("On order"),
   PAGED("Paged"),
   DECLARED_LOST("Declared lost"),
-  RECEIVED("Received"),
   ORDER_CLOSED("Order closed");
 
   private static final Map<String, ItemStatusName> VALUE_TO_INSTANCE_MAP = initValueToInstanceMap();
--- a/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
+++ b/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
@@ -37,7 +37,7 @@ public enum ItemStatusName {
     return VALUE_TO_INSTANCE_MAP.get(value);
   }
 
-  public static boolean hasKnownValue(String value) {
+  public static boolean isStatusCorrect(String value) {
     return VALUE_TO_INSTANCE_MAP.containsKey(value);
   }
 
--- a/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
+++ b/src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
@@ -16,7 +16,9 @@ public enum ItemStatusName {
   ON_ORDER("On order"),
   PAGED("Paged"),
   WITHDRAWN("Withdrawn"),
-  DECLARED_LOST("Declared lost");
+  DECLARED_LOST("Declared lost"),
+  RECEIVED("Received"),
+  ORDER_CLOSED("Order closed");
 
   private static final Map<String, ItemStatusName> VALUE_TO_INSTANCE_MAP = initValueToInstanceMap();
   private final String value;
--- a/src/main/java/org/folio/inventory/resources/Items.java
+++ b/src/main/java/org/folio/inventory/resources/Items.java
@@ -4,6 +4,7 @@ import static org.folio.inventory.common.FutureAssistance.allOf;
 import static org.folio.inventory.support.CqlHelper.multipleRecordsCqlQuery;
 import static org.folio.inventory.support.JsonArrayHelper.toListOfStrings;
 import static org.folio.inventory.support.JsonHelper.getNestedProperty;
+import static org.folio.inventory.validation.ItemStatusValidator.itemHasNoOrKnownStatus;
 
 import java.io.UnsupportedEncodingException;
 import java.lang.invoke.MethodHandles;
@@ -16,6 +17,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
@@ -46,6 +48,7 @@ import org.folio.inventory.support.http.server.ForwardResponse;
 import org.folio.inventory.support.http.server.JsonResponse;
 import org.folio.inventory.support.http.server.ServerErrorResponse;
 import org.folio.inventory.support.http.server.SuccessResponse;
+import org.folio.inventory.support.http.server.ValidationError;
 
 import io.vertx.core.http.HttpClient;
 import io.vertx.core.json.JsonObject;
@@ -132,6 +135,12 @@ public class Items {
 
     JsonObject item = routingContext.getBodyAsJson();
 
+    Optional<ValidationError> validationError = itemHasNoOrKnownStatus(item);
+    if (validationError.isPresent()) {
+      JsonResponse.unprocessableEntity(routingContext.response(), validationError.get());
+      return;
+    }
+
     Item newItem = requestToItem(item);
 
     ItemCollection itemCollection = storage.getItemCollection(context);
@@ -164,6 +173,12 @@ public class Items {
     WebContext context = new WebContext(routingContext);
 
     JsonObject itemRequest = routingContext.getBodyAsJson();
+
+    Optional<ValidationError> validationError = itemHasNoOrKnownStatus(itemRequest);
+    if (validationError.isPresent()) {
+      JsonResponse.unprocessableEntity(routingContext.response(), validationError.get());
+      return;
+    }
 
     Item newItem = requestToItem(itemRequest);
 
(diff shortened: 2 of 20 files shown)


src/main/java/org/folio/inventory/domain/items/ItemStatusName.java
```diff
@@ -0,0 +1,52 @@
+package org.folio.inventory.domain.items;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+public enum ItemStatusName {
```
marcjohnson-kint: Why is this specifically the item status’ name only, rather than the status itself?

Personally, I think this represents the status rather than the name of the status.

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -314,7 +326,7 @@ private Item requestToItem(JsonObject itemRequest) {
             .withPermanentLocationId(permanentLocationId)
             .withTemporaryLocationId(temporaryLocationId)
             .withTemporaryLoanTypeId(temporaryLoanTypeId)
-            .withCopyNumbers(copyNumbers)
+            .withCopyNumber(itemRequest.getString("copyNumber"))
```
marcjohnson-kint: Folks seem to likely constants for the names of the JSON properties. Given this is used in multiple places, I think it might be Worthing following suite with this

src/main/java/org/folio/inventory/support/JsonHelper.java
```diff
@@ -17,6 +20,20 @@ public static String getNestedProperty(
...
 
+  public static String getString(JsonObject representation, String propertyName) {
+    if (representation != null) {
+      return representation.getString(propertyName);
+    }
+
+    return null;
+  }
+
+  public static void includeIfPresent(JsonObject target, String propertyName, String value) {
```
marcjohnson-kint: Why does this method use the name target for the JSON object and the one above use representation?

src/main/java/org/folio/inventory/validation/ItemStatusValidator.java
```diff
@@ -0,0 +1,38 @@
...
+
+public final class ItemStatusValidator {
+
+  private ItemStatusValidator() {}
+
+  public static Optional<ValidationError> itemHasCorrectStatus(JsonObject itemRequest) {
+    final String statusName = JsonHelper.getNestedProperty(itemRequest, "status", "name");
+
+    if (StringUtils.isBlank(statusName)) {
+      return Optional.of(new ValidationError(
```
marcjohnson-kint: Is this formatting mimicking code I’d previously written, or do you find the parameters easier to interpret with this formatting?

src/test/java/api/support/builders/ItemRequestBuilder.java
```diff
@@ -29,13 +29,14 @@
...
   private final String hrid;
+  private final String copyNumber;
 
   public ItemRequestBuilder() {
     this(UUID.randomUUID(), null, null, null, "645398607547",
       AVAILABLE_STATUS, bookMaterialType(), null, null, null,
       canCirculateLoanType(), null, null, null,
       null, null, null,
-      null
+      null, null
```
marcjohnson-kint: Given that multiple values are present on the same line, these null values can be included on the previous line.

src/test/java/support/fakes/FakeStorageModule.java
```diff
@@ -294,22 +294,34 @@ private void checkTokenHeader(RoutingContext routingContext) {
...
       routingContext.next();
-    }
-    else {
+    } else {
       JsonResponse.unprocessableEntity(routingContext.response(), errors);
     }
   }
 
+  private Object getPropertyValue(JsonObject body, String requiredProperty) {
+    String[] pathElements = requiredProperty.split("\\.");
```
marcjohnson-kint: I think we should consider using something like JSON path for checks like this rather than our own format. This is likely beyond the scope of this work

src/test/java/support/fakes/processors/StorageRecordPreProcessors.java
```diff
@@ -90,7 +94,40 @@ private StorageRecordPreProcessors() {
...
+
+      if (!Objects.equals(oldStatus, newStatus)) {
+        JsonObject newStatusObject = newItem.containsKey("status")
+          ? newItem.getJsonObject("status")
+          : new JsonObject();
+
+        newStatusObject = newStatusObject.put("date",
+          DateTime.now(DateTimeZone.UTC)
+            // RMB format
+            .toString("yyyy-MM-dd'T'HH:mm:ss.SSS+0000")
```
marcjohnson-kint: Does RAML Module Builder always use this format?

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1383,6 +1421,208 @@ public void cannotRemoveHRID() throws Exception {
     assertThat(existingItem, is(createdItem));
   }
 
+  @Test
+  public void cannotSetWrongStatus() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void cannotProvideUnrecognisedStatusName() throws Exception {
```

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1383,6 +1421,208 @@ public void cannotRemoveHRID() throws Exception {
...
+    Response response = postCompleted.get(5, TimeUnit.SECONDS);
+    assertThat(response, hasValidationError(
+      "Undefined status specified",
+      "status.name",
+      "Wrong name"
+    ));
+  }
+
+  @Test
+  public void cannotUpdateItemWithWrongStatus() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void cannotUpdateItemWithUnrecognisedStatusName() throws Exception {
```

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1383,6 +1421,208 @@ public void cannotRemoveHRID() throws Exception {
...
+    Response updateResponse = updateItem(updatedItem);
+    assertThat(updateResponse, hasValidationError(
+      "Undefined status specified",
+      "status.name",
+      "Wrong name"
+    ));
+  }
+
+  @Test
+  public void cannotRemoveEntireStatus() throws Exception {
```
marcjohnson-kint:
```suggestion
  public void cannotUpdateItemWithNoStatus() throws Exception {
```

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1383,6 +1421,208 @@ public void cannotRemoveHRID() throws Exception {
...
+    updatedItem.remove("status");
+
+    Response updateResponse = updateItem(updatedItem);
+    assertThat(updateResponse,
+      hasValidationError("Status is a required field", "status", null)
+    );
+  }
+
+  @Test
+  public void canRemoveStatusName() throws Exception {
```
marcjohnson-kint: Should this be cannot?

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1383,6 +1421,208 @@ public void cannotRemoveHRID() throws Exception {
...
+      .put("status", new JsonObject());
+
+    Response updateResponse = updateItem(updatedItem);
+    assertThat(updateResponse,
+      hasValidationError("Status is a required field", "status", null)
+    );
+  }
+
+  @Test
+  public void canSetCopyNumberProperty() throws Exception {
```
marcjohnson-kint: Similarly to the storage module, are there tests for general creating or updating an item?

src/test/java/support/fakes/processors/StorageRecordPreProcessors.java
bohdan-suprun: Jackson uses this pattern by default

src/test/java/api/items/ItemApiExamples.java
bohdan-suprun: This test and `canUpdateCopyNumberProperty` have additional statements to verify that old `copyNumbers` property is not present in item-storage (see line # 1552).

I think it make sense to have them separately.

src/test/java/support/fakes/processors/StorageRecordPreProcessors.java
marcjohnson-kint: Does that mean that RAML Module Builder also does this by default?

src/test/java/api/items/ItemApiExamples.java
marcjohnson-kint: Ok. As API tests are expensive to maintain, when should this separate test be removed?

src/test/java/support/fakes/processors/StorageRecordPreProcessors.java
bohdan-suprun: Yes, RMB uses Jackson for serialization/deserialization.

marcjohnson-kint: Does it definitely use the default formatting?

If so, that is a shame, because that format isn't compliant with RFC-3339. 

I think we should use a format that is compliant.

src/test/java/api/items/ItemApiExamples.java
bohdan-suprun: hmm... You're right, I think thees checks only important for refactoring step. Will remove them.

src/test/java/support/fakes/processors/StorageRecordPreProcessors.java
bohdan-suprun: Here is `90000` item from snapshot:

image

The `lastCheckIn.dateTime` and `metadata.updatedDate/createdDate` have format: date-time in json schema.

As you can see, they uses the `yyyy-MM-dd'T'HH:mm:ss.SSS+0000` format.

marcjohnson-kint: Ok, that's really a bug. For the purposes of this change, can you extract a named constant with comments that explain this issue.

src/test/java/api/items/ItemApiExamples.java
bohdan-suprun: yes, thanks, fixed.

src/test/java/support/fakes/processors/StorageRecordPreProcessors.java
bohdan-suprun: Extracted, please take a look.

src/main/java/org/folio/inventory/validation/ItemStatusValidator.java
bohdan-suprun: I've changed formatting, now all the parameters on the single line.

Is it something you meant?

src/test/java/support/fakes/FakeStorageModule.java
bohdan-suprun: Agreed.

src/main/java/org/folio/inventory/validation/ItemStatusValidator.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import org.folio.inventory.domain.items.ItemStatusName;
+import org.folio.inventory.support.JsonHelper;
+import org.folio.inventory.support.http.server.ValidationError;
+
+import io.vertx.core.json.JsonObject;
+
+public final class ItemStatusValidator {
+
```
marcjohnson-kint: I would personally remove this blank line

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1383,6 +1429,145 @@ public void cannotRemoveHRID() throws Exception {
...
+      .put("status", new JsonObject());
+
+    Response updateResponse = updateItem(updatedItem);
+    assertThat(updateResponse,
+      hasValidationError("Status is a required field", "status", null)
+    );
+  }
+
+  @Test
+  public void canHandleStatusDateProperty() throws Exception {
```
marcjohnson-kint: What does handle mean in this situation?

marcjohnson-kint: Please do not use `throws Exception` it masked all of the relevant checked exceptions and makes refactoring the tests harder

src/main/java/org/folio/inventory/support/JsonHelper.java
```diff
@@ -28,9 +28,11 @@ public static String getString(JsonObject representation, String propertyName) {
     return null;
   }
 
-  public static void includeIfPresent(JsonObject target, String propertyName, String value) {
-    if (target != null && StringUtils.isNotBlank(propertyName) && value != null) {
-      target.put(propertyName, value);
+  public static void includeIfPresent(
+    JsonObject representation, String propertyName, String value) {
```
marcjohnson-kint: Why has the formatting changed to include all of the parameters on the following line?

bohdan-suprun: yes

src/test/java/api/items/ItemApiExamples.java
bohdan-suprun: Yes, they mask, but there is another side, when we specify all the "standard" test checked exceptions, which in a fact do not give me any useful info (since we don't execute test methods in code), but instead have to maintain (e.g. test were refactored to use client instead of okapiClient, and now `MalformedURLException` have to be removed) and makes tests cluttered a bit.

But, it is just my opinion.

marcjohnson-kint: I think your point about useful information is interesting. It is annoying that these exceptions propagate. In general I think this is a flaw of checked exceptions in general.

I'm going through a refactoring of mod-circulation's tests to use Rest Assured, which affects a lot of the `standard` exceptions thrown by the tests. The use of `Exception` has made that harder.

Personally, I like that we need to remove the exceptions when they are no longer thrown, it is a clear indicator of what exceptions are thrown.

Do you use exception because it means that test maintainers don't have to think about checked exceptions?

marcjohnson-kint: @bohdan-suprun You are correct that this is a more involved topic than for this pull request, please don't make any changes to the checked exceptions

bohdan-suprun: Yes, I prefer single `Exception` here because we never call the test method within another method, so we don't need to definitely know what may be thrown, and JUnit takes care about handling it.

But if it makes harder to introduce the new approach, then I will expand the throws clause.

bohdan-suprun: I've renamed it to `statusDatePropertyPresentOnStatusUpdated` does it work better?

marcjohnson-kint: I was trying to better understand. What does the test demonstrate, that after an item status is changed that the `status changed date` is present in the response?

marcjohnson-kint: @bohdan-suprun It does make refactoring harder because it impedes some of the automated tools from detecting the correct exceptions.

This can be addressed during a future change if necessary.

src/main/java/org/folio/inventory/support/JsonHelper.java
marcjohnson-kint: Apologies, I was asking why it has been formatted that way, rather than splitting the line part way through the parameters?

bohdan-suprun: It is because of the horizontal alignment.

src/test/java/api/items/ItemApiExamples.java
bohdan-suprun: The test demonstrates that the status.date is forwarded from item-storage and has expected format and is within 2 seconds from _now()_.

PR #166, opened 2020-01-02
MODINV-190: Most recent check in information clears when item record is edited.

VRohach:
JIRA-ticket: https://issues.folio.org/browse/MODINV-190

The main purpose of this PR is to fix the problem with "lastCheckIn" field in response to the item.

### DIFF
--- a/src/main/java/org/folio/inventory/resources/Items.java
+++ b/src/main/java/org/folio/inventory/resources/Items.java
@@ -324,7 +324,7 @@ public class Items {
             .withElectronicAccess(electronicAccess)
             .withStatisticalCodeIds(statisticalCodeIds)
             .withPurchaseOrderLineidentifier(itemRequest.getString(Item.PURCHASE_ORDER_LINE_IDENTIFIER))
-            .withLastCheckIn(LastCheckIn.from(itemRequest.getJsonObject("lastCheckIn")))
+            .withLastCheckIn(LastCheckIn.from(itemRequest.getJsonObject(Item.LAST_CHECK_IN)))
             .withTags(tags);
   }
 
--- a/src/main/java/org/folio/inventory/resources/Items.java
+++ b/src/main/java/org/folio/inventory/resources/Items.java
@@ -29,6 +29,7 @@ import org.folio.inventory.common.domain.Success;
 import org.folio.inventory.domain.items.CirculationNote;
 import org.folio.inventory.domain.items.Item;
 import org.folio.inventory.domain.items.ItemCollection;
+import org.folio.inventory.domain.items.LastCheckIn;
 import org.folio.inventory.domain.items.Note;
 import org.folio.inventory.domain.items.Status;
 import org.folio.inventory.domain.sharedproperties.ElectronicAccess;
@@ -323,6 +324,7 @@ public class Items {
             .withElectronicAccess(electronicAccess)
             .withStatisticalCodeIds(statisticalCodeIds)
             .withPurchaseOrderLineidentifier(itemRequest.getString(Item.PURCHASE_ORDER_LINE_IDENTIFIER))
+            .withLastCheckIn(LastCheckIn.from(itemRequest.getJsonObject("lastCheckIn")))
             .withTags(tags);
   }
 
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
-      "version": "9.7",
+      "version": "9.8",
       "handlers": [
         {
           "methods": ["GET"],
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -421,7 +421,6 @@
       "type": "object",
       "additionalProperties": false,
       "description": "Information about when an item was last checked in",
-      "readonly": true,
       "properties": {
         "dateTime": {
           "type": "string",
--- a/src/main/java/org/folio/inventory/domain/items/Item.java
+++ b/src/main/java/org/folio/inventory/domain/items/Item.java
@@ -37,6 +37,7 @@ public class Item {
   public static final String NOTES_KEY = "notes";
   public static final String CIRCULATION_NOTES_KEY = "circulationNotes";
   public static final String PURCHASE_ORDER_LINE_IDENTIFIER = "purchaseOrderLineIdentifier";
+  public static final String LAST_CHECK_IN = "lastCheckIn";
 
   public final String id;
   private String hrid;
--- a/src/main/java/org/folio/inventory/domain/items/Item.java
+++ b/src/main/java/org/folio/inventory/domain/items/Item.java
@@ -459,6 +459,7 @@ public class Item {
             .withYearCaption(this.yearCaption)
             .withElectronicAccess(this.electronicAccess)
             .withStatisticalCodeIds(this.statisticalCodeIds)
+            .withLastCheckIn(this.lastCheckIn)
             .withPurchaseOrderLineidentifier(this.purchaseOrderLineidentifier);
   }
 
--- a/src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
+++ b/src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
@@ -136,6 +136,10 @@ class ExternalStorageModuleItemCollection
       itemToSend.put("status", item.getStatus());
     }
 
+    if(item.getLastCheckIn() != null) {
+      itemToSend.put(Item.LAST_CHECK_IN, item.getLastCheckIn().toJson());
+    }
+
     includeIfPresent(itemToSend, Item.HRID_KEY, item.getHrid());
     itemToSend.put(Item.FORMER_IDS_KEY, item.getFormerIds());
     itemToSend.put(Item.DISCOVERY_SUPPRESS_KEY, item.getDiscoverySuppress());
@@ -170,7 +174,6 @@ class ExternalStorageModuleItemCollection
     itemToSend.put(Item.STATISTICAL_CODE_IDS_KEY, item.getStatisticalCodeIds());
     itemToSend.put(Item.PURCHASE_ORDER_LINE_IDENTIFIER, item.getPurchaseOrderLineidentifier());
     itemToSend.put(Item.TAGS_KEY, new JsonObject().put(Item.TAG_LIST_KEY, new JsonArray(item.getTags())));
-    itemToSend.put(Item.LAST_CHECK_IN, item.getLastCheckIn().toJson());
 
     return itemToSend;
   }
--- a/src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
+++ b/src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
@@ -170,6 +170,7 @@ class ExternalStorageModuleItemCollection
     itemToSend.put(Item.STATISTICAL_CODE_IDS_KEY, item.getStatisticalCodeIds());
     itemToSend.put(Item.PURCHASE_ORDER_LINE_IDENTIFIER, item.getPurchaseOrderLineidentifier());
     itemToSend.put(Item.TAGS_KEY, new JsonObject().put(Item.TAG_LIST_KEY, new JsonArray(item.getTags())));
+    itemToSend.put(Item.LAST_CHECK_IN, item.getLastCheckIn().toJson());
 
     return itemToSend;
   }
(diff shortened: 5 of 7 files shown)


src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -323,6 +324,7 @@ private Item requestToItem(JsonObject itemRequest) {
             .withElectronicAccess(electronicAccess)
             .withStatisticalCodeIds(statisticalCodeIds)
             .withPurchaseOrderLineidentifier(itemRequest.getString(Item.PURCHASE_ORDER_LINE_IDENTIFIER))
+            .withLastCheckIn(LastCheckIn.from(itemRequest.getJsonObject(Item.LAST_CHECK_IN)))
```
bohdan-suprun: Could you please remove `readonly: true` for the `lastCheckin` property in the item schema.

mrdpopov: I think it should be readonly in inventory.

bohdan-suprun: Ok, but readonly means that the property is not intended to be updated (RMB module even removes this property before dispatching the request to the API resource).

But as far as I can see the property is being transferred from inventory to inventory-storage (see `src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java`) it means that the property can be added/updated.

mrdpopov: My understanding is that the readonly properties are NOT writable using this API. Anyway that was the intention: Lastcheckin not to be not updatable using /inventory/items API. If there are other notions to the readonly property, could you please point to the documentation?

bohdan-suprun: >  Anyway that was the intention: Lastcheckin not to be updatable via /inventory/items API.

But it will be, I think, since you're passing it to the `item-storage`.  Could you please add a test case then to verify it?

mrdpopov: That was never part of the requirements. I don't think  it is needed.

marcjohnson-kint: @mrdpopov 

> That was never part of the requirements. I don't think it is needed.

What do you mean that it wasn't part of the requirements?

If a property is marked as read only in the API then any client submitted value should be ignored, and there should be a way to demonstrate that.

mrdpopov: @marc-johnson Do you  think that it should be readonly?

marcjohnson-kint: > Do you think that it should be readonly?

I don't know if it should be read-only. It is currently marked as read only.

I think what is important is that the API definition and the module behaviour are consistent, if it is read only, then any client supplied value should be ignored.

(Also, please use my work github account @marcjohnson-kint when referring to me on pull requests, otherwise I get duplicate notifications)

mrdpopov: Let me share some intuition behind the reasons the last checkin was **meant** to be read-only. According to the business requirements, the last checkin information should be saved when, well, item is checked in. If we allow it to be modified anywhere else, it seems to me as a potential problem and source of bugs in future.
It looks like, however, the "read-only" is perceived as something else in this context. I may be wrong, but the whole item is recreated from the storage, not just the last checkin field. And it also looks like the read-only attribute has no effect in this case whatsoever.

And all of that lead to me to think about why the functionality of updating items is even present in mod-inventory, which is in essence a roundtrip to the storage. Whereas one could just go directly to item storage. There are at least two ways of updating items that behave differently. Just food for thought.

bohdan-suprun: Thanks @mrdpopov for explaining this. I think then you have to add some validation restrictions (it seems to me that `readonly` attribute is not add any effect for a non RMB modules, like mod-inventory).

I think you should add following validation rules:
* Restrict lastCheckin to be added during item create;
* Maybe allow lastCheckin to be updated/added when old and new item statuses is a some combination, (I believe old status should be `Checked out`, but maybe it is not for all the cases), hopefully we already retrieve previous item to validate barcode and hrid). I'm not sure, if it is possible.

Does it sound fine?

mrdpopov: I think it makes sense to go for postponing the test creation and go with read-only false for now.

marcjohnson-kint: As I understand it, it has already been marked as read only. Are you proposing changing the API to make it writable?

Changing it back to read only later will be a breaking compatibility change.

VRohach: Ok. After discussion, we decide to remove "readonly" field.

bohdan-suprun: Hi @marcjohnson-kint, no it is readonly now. My concern was that since we will be able to pass the lastCheckin property to item-storage then it have to not be readonly, because it can be changed in DB then.

We had a talk offline with @mrdpopov and @VRohach, so decided to make it writable for now.

mrdpopov: I already  explained the reasoning behind making it not writable and I still think it should not be. But due to implementation details(the item dto is first created from the request in inventory, then it's sent to item-storage), the item is fully sent to the storage to update it, it probably makes more sense to go with making the last check in writable for now.

Alternatively, and this makes the most sence to me, we could change UI to send PUTS directly to items-storage.

marcjohnson-kint: @bohdan-suprun  That was my question. It is already read only (from a prior change), so we are intending to change it to writable, with a corresponding API change and tests?

bohdan-suprun: Yes, we are. Now it is going to be writable.

marcjohnson-kint: > I already explained the reasoning behind making it not writable and I still think it should not be. But due to implementation details (the item dto is first created from the request in inventory, then it's sent to item-storage), the item is fully sent to the storage to update it, it probably makes more sense to go with making the last check in writable for now.

What I don't understand is why not change the implementation in mod-inventory to make the property read only?

> Alternatively, and this makes the most sense to me, we could change UI to send PUTS directly to items-storage.

Why does this make the most sense to you. That makes it writable from the UI, I thought that is what you wanted to avoid?

> According to the business requirements, the last checkin information should be saved when, well, item is checked in. If we allow it to be modified anywhere else, it seems to me as a potential problem and source of bugs in future.

Any client to `item-storage` can change this information, not only circulation, so this source of problems is already present.

> It looks like, however, the "read-only" is perceived as something else in this context. I may be wrong, but the whole item is recreated from the storage, not just the last checkin field. And it also looks like the read-only attribute has no effect in this case whatsoever.

I don't think this is about how `read-only` properties are perceived. I think it is due to this property being stated as `read only` in the interface definition and yet writable in the implementation.

Read only is not automatically implemented because this is not a RAML Module Builder based module, that does not change it's intended meaning, only how it is implemented.

> And all of that lead to me to think about why the functionality of updating items is even present in mod-inventory, which is in essence a roundtrip to the storage. Whereas one could just go directly to item storage. There are at least two ways of updating items that behave differently. Just food for thought.

That is the nature of the business logic module and storage module split in FOLIO. Much like it is possible to create loans directly in circulation storage without doing a check out and the two behave very differently.

Would you expect clients to use storage directly most of the time?

mrdpopov: > That is the nature of the business logic module and storage module split in FOLIO. Much like it is possible to create loans directly in circulation storage without doing a check out and the two behave very differently.

But it's a bit different with update in mod-inventory, isn't it? Let's consider a scenario when a users edits an item in UI:

1. request goes to PUT /inventory/items/id along with full item's json, which includes the last checkin object(non editable in UI). 


2. In the server-side, the item from request is manually copied to another item dto and sent to the item-storage.
Do I get it correctly that the read-only implies removing it from the incoming requests?
**If the last check in weren't in the request, the updated item wouldn't contain it either.**

3. Browser goes back to  GET /inventory/items/id, under the hood the server does to item-storage.

As you can see, In order to edit an item, the item's JSON is fully replaced with the new one. That is why we need the last check in object from UI, at least with current implementation.

@marc-johnson 
As I get your point, the interface and the implementation should be in line with each other. Since making it read-only with the current implementation entails a degree of designing and reworking, do you think we can go for updating the interface to declare it writable?

bohdan-suprun: @mrdpopov. It is more clear now, thanks. So the `lastCheckin` is not intended to be updated  by UI in this case, am I right? 

How we going to init the value of this property, via item-storage, right?

If so, then I think we're ok to leave the property as readonly, but I'd recommend to add following logic to an item update - for every item update set `newItem.lastCheckin` = `oldItem.lastCheckin`. It is going to be more safe, because as for now, you can easily replace the property using API.

marcjohnson-kint: > > That is the nature of the business logic module and storage module split in FOLIO. Much like it is possible to create loans directly in circulation storage without doing a check out and the two behave very differently.
> 
> But it's a bit different with update in mod-inventory, isn't it? Let's consider a scenario when a users edits an item in UI:
> 
> 1. request goes to PUT /inventory/items/id along with full item's json, which includes the last checkin object(non editable in UI).

This is because PUT semantics are for a complete replacement of a resource. And the UI developers don't want to have to remove read only properties themselves. This means the server is expected to manage these properties differently.

> 2. In the server-side, the item from request is manually copied to another item dto and sent to the item-storage.

Yes, the business logic representation of the item is converted into the storage representation. 

They are often very similar (in some cases the same) so it can be strange to map between them.

>    Do I get it correctly that the read-only implies removing it from the incoming requests?
>    **If the last check in weren't in the request, the updated item wouldn't contain it either.**

Removing the value of the property from the incoming representation is only part of what the server needs to do. This is done by RAML Module Builder.

It also needs to make sure that the updated storage representation contains the previous information. This is *not* done by RAML Module Builder (except for change metadata). Modules have to implement this themselves, whether they are storage or business logic modules.

> 3. Browser goes back to  GET /inventory/items/id, under the hood the server does to item-storage.

> As you can see, In order to edit an item, the item's JSON is fully replaced with the new one. That is why we need the last check in object from UI, at least with current implementation.

FOLIO has struggled with the complete replacement aspect of PUT and how it designs resources.

A combination of read only and writable properties in the same resource has made this more complicated. The server needs to reconstruct a full representation, based upon a combination of incoming request and existing stored representation. 

This is the case for both storage and business logic modules. I believe many of the storage modules have used triggers to do this.

> @marc-johnson
> As I get your point, the interface and the implementation should be in line with each other. Since making it read-only with the current implementation entails a degree of designing and reworking, do you think we can go for updating the interface to declare it writable?

Making it writable will simplify the immediate situation. Making it read only in future will be a breaking compatibility change.

I think the decision depends upon how important it is to reduce the chances that a client changes this information (possibly accidentally).

marcjohnson-kint: > How we are going to init the value of this property, via item-storage, right?

The value is set by circulation, meaning it must be writable in storage for the moment (at least until we use events as a different door for this kind of change)

> If so, then I think we're ok to leave the property as readonly, but I'd recommend to add following logic to an item update - for every item update set `newItem.lastCheckin` = `oldItem.lastCheckin`. 

What is `oldItem` in this context?

bohdan-suprun: It is the item representation before update (i.e. from DB).

marcjohnson-kint: > It is the item representation before update (i.e. from DB).

In your example, is that in the storage module (where the value is writable) or in the business logic module (where it is readonly)?

bohdan-suprun: In business module. We can not make this field readonly in storage I think.

I think it can be similar to how we restrict HRID to be updated for an item/instance, but except failing request, we just transfer old value of the field.

marcjohnson-kint: > In business module. We can not make this field readonly in storage I think.

Old item refers to the storage representation of the item then?

> I think it can be similar to how we restrict HRID to be updated for an item/instance, but except failing request, we just transfer old value of the field.

Yeah, a similar process could be used here. This is simpler if the existing item is already being fetched as part of updating an item.

mrdpopov: @marcjohnson-kint  Can we go with read-only = false for now as the code is already pushed, and accept the risks of someone inadvertidely modifying the last checkin?

marcjohnson-kint: @mrdpopov I'm most concerned about the API documentation being consistent with the behaviour. If that means removing the read only property, then that is ok, as long as a minor version update is also applied.

I think we might be taking the expedient way out of this, and if we actually want the property to be read only, this is deferring the pain for the future.

VRohach:
@marcjohnson-kint  I bumped up minor version. Could you please approve this PR ?

PR #168, opened 2020-01-21
MODINV-181 - Integrate mod-inventory with mod-pubsub

RuslanLavrov:


src/main/java/org/folio/inventory/InventoryVerticle.java
```diff
@@ -12,10 +12,7 @@
 import io.vertx.ext.web.Router;
 import org.folio.inventory.common.WebRequestDiagnostics;
 import org.folio.inventory.domain.ingest.IngestMessageProcessor;
-import org.folio.inventory.resources.Instances;
-import org.folio.inventory.resources.InstancesBatch;
-import org.folio.inventory.resources.IsbnUtilsApi;
-import org.folio.inventory.resources.Items;
+import org.folio.inventory.resources.*;
```
OleksiiKuzminov: imports

src/main/java/org/folio/inventory/resources/TenantApi.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  private void postTenant(RoutingContext routingContext) {
+    MultiMap headersMap = routingContext.request().headers();
+    String tenantId = TenantTool.calculateTenantId(headersMap.get(OKAPI_TENANT_HEADER));
+    LOG.info("sending postTenant for " + tenantId);
+
+    Map<String, String> okapiHeaders = getOkapiHeaders(routingContext);
+    registerModuleToPubsub(okapiHeaders, routingContext.vertx())
+      .setHandler(ar ->
```
bohdan-suprun: Do we need exception handling here? `ar` can be either succeeded or failed future here.

RuslanLavrov: for now, we ignore the result of the module registration to mod-pubsub until mod-pubsub is used as a transport for the data import process

src/main/resources/MessagingDescriptor.json
```diff
@@ -0,0 +1,41 @@
+{
```
marcjohnson-kint: Where is there a definition of a what a messaging description is and what it is used for?

src/main/java/org/folio/inventory/resources/TenantApi.java
```diff
@@ -0,0 +1,67 @@
...
+      if (headerKey.startsWith("x-okapi")) {
+        okapiHeaders.put(headerKey, headerEntry.getValue());
+      }
+    });
+    return okapiHeaders;
+  }
+
+  private Future<Void> registerModuleToPubsub(Map<String, String> headers, Vertx vertx) {
+    Future<Void> future = Future.future();
+    PubSubClientUtils.registerModule(new OkapiConnectionParams(headers, vertx))
```
marcjohnson-kint: What does this method do?

src/main/java/org/folio/inventory/resources/TenantApi.java
bohdan-suprun: Ok, thanks for explaining it.

src/test/java/api/tenant/TenantApiExamples.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.concurrent.TimeoutException;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class TenantApiExamples extends ApiTests {
+
+  public TenantApiExamples() throws MalformedURLException {
+    super();
+  }
```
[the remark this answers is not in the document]
RuslanLavrov: we need to explicitly specify the checked exception MalformedURLException to the constructor signature

src/main/resources/MessagingDescriptor.json
RuslanLavrov: Messaging descriptor is a file that describes what events the module publishes and to which events module is subscribed. This file is used for registration of module to mod-pubsub by the method PubSubClientUtils#registerModule(). Please, find a definition of MessagingDescriptor file in the MessagingDescriptor section of [mod-pubsub documentation](https://github.com/folio-org/mod-pubsub#messaging-descriptor).

src/main/java/org/folio/inventory/resources/TenantApi.java
RuslanLavrov: this method registers module as publisher/subscriber of events to mod-pubsub using information from MessagingDescriptor file. Please, find information about module registration in  [documentation of mod-pubsub](https://github.com/folio-org/mod-pubsub#module-registration-in-pub-sub)

src/main/resources/MessagingDescriptor.json
marcjohnson-kint: Ok, thanks. I don't really understand why a file is required for this.

src/main/java/org/folio/inventory/resources/TenantApi.java
marcjohnson-kint: That section appears to describe using the client library, it doesn't seem to describe what the client library does. Have I missed something?

OleksiiKuzminov: > PubSub Client provides utility methods for registering modules as Publishers and/or Subscribers in pub-sub and sending Events.

OleksiiKuzminov:
@marcjohnson-kint Why do you added label "do not merge"? There are no breaking changes in this PR

marcjohnson-kint:
> @marcjohnson-kint Why do you added label "do not merge"? There are no breaking changes in this PR

I added do not merge in order to allow code owners an opportunity to review this pull request, given that it has been approved very quickly.

src/main/java/org/folio/inventory/resources/TenantApi.java
marcjohnson-kint: @OleksiiKuzminov Thanks. That doesn't really tell me anything beyond the method names of the client library.

I have questions like:
* How do I know which permissions to grant for the methods in the client library?
* How does a client of pub-sub know what a method does in respect to pub sub configuration?

I doubt these questions are going to be answered in the context of this pull request.

OleksiiKuzminov:
> This change does not introduce any behaviour in terms of publishing or receiving events. As far as I can tell it only registers it's intentions to publish or receive events and it deliberately does not care about whether that registration is successful, which I think significantly reduces the value of this change.

These changes prepare mod-inventory to further integration with data import. After this PR mod-inventory is ready to work with pub-sub and not only for data-import needs. It's a main idea of these changes. Data-import integration and handlers for events will be added in the next stories and PRs.

> Many of the details are hidden behind the client library and it is difficult to tell what should or should not be happening.

And it's good from my point of view. We have a lib with easy to understand methods, and manual. As you can see for start using pub-sub in any module you need just call a couple of methods and declare MessageDescriptor. 

> My interpretation of the use of CREATED_SRS_MARC_BIB_RECORD suggests that responsibility for converting a MARC record is transferred from data import to inventory. Is that the intention? If so, I think this is a fairly significant architectural change.

Yes, mod inventory is responsible for work with instances\holdings\items, so all actions like save\update will be appearing there. All these actions will be processed in a particular event handler for data-import and don't change any existing architecture. What about data-import architecture, so there are no changes, cause it what we designed, approved and planned to do. Current data-import is just mapping and saving instances without using all data-import flows and profiles

marcjohnson-kint:
> > This change does not introduce any behaviour in terms of publishing or receiving events. As far as I can tell it only registers it's intentions to publish or receive events and it deliberately does not care about whether that registration is successful, which I think significantly reduces the value of this change.
> 
> These changes prepare mod-inventory to further integration with data import. After this PR mod-inventory is ready to work with pub-sub and not only for data-import needs. It's a main idea of these changes. Data-import integration and handlers for events will be added in the next stories and PRs.

It's difficult for me to understand this without understand the rest of the work, which I am not familiar with.

> > Many of the details are hidden behind the client library and it is difficult to tell what should or should not be happening.
> 
> And it's good from my point of view. We have a lib with easy to understand methods, and manual. As you can see for start using pub-sub in any module you need just call a couple of methods and declare MessageDescriptor.

This is likely a topic for a separate conversation, for example, how does a developer know what permissions to grant when using this client library.

> > My interpretation of the use of CREATED_SRS_MARC_BIB_RECORD suggests that responsibility for converting a MARC record is transferred from data import to inventory. Is that the intention? If so, I think this is a fairly significant architectural change.
> 
> Yes, mod inventory is responsible for work with instances\holdings\items, so all actions like save\update will be appearing there. All these actions will be processed in a particular event handler for data-import and don't change any existing architecture. What about data-import architecture, so there are no changes, cause it what we designed, approved and planned to do. Current data-import is just mapping and saving instances without using all data-import flows and profiles

I don't understand. What is mod-inventory going to do when a CREATED_SRS_MARC_BIB_RECORD event is received?

As I understand it currently, data import is responsible for converting MARC records to inventory instances. If inventory is going to convert MARC into instances instead, that is a significant architectural design change IMO.

I think I need to ask this question to a broader audience on the JIRA issue.

> If you have not any concerns about this PR and changes, plz let us know.

I've tried to express my concerns above. My primary concern with this pull request is that it does not provide any observable change in behaviour. 

The most meaningful change is the registration with mod-pubsub, and the result of that is intentionally ignored. If I were to suggest a single change to this pull request, it would be to check this response and fail the tenant API if this failed. That way, the system would fail fast if this failed, and this pull request would provide some validation of the integration.

I don't understand the decomposition of work here enough to be able to evaluate a small non-vertical slice of the work.

Given that, I'm not sure I am capable of adequately reviewing this work.

PR #169, opened 2020-01-29
MODINV-194: Add 'Claimed returned' to allowed item statuses list. 

bohdan-suprun:
https://issues.folio.org/browse/MODINV-194: Backend: Claim returned: mark an item claim returned.

* Add `Claimed returned` to allowed item statuses list;
* Add test to verify schema enum is consistent to `org.folio.inventory.domain.items.ItemStatusName` enum.

### DIFF
--- a/src/test/java/api/items/ItemAllowedStatusesSchemaTest.java
+++ b/src/test/java/api/items/ItemAllowedStatusesSchemaTest.java
@@ -1,0 +1,103 @@
+package api.items;
+
+import static api.support.InstanceSamples.smallAngryPlanet;
+import static java.nio.file.Files.readAllBytes;
+import static java.nio.file.Paths.get;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+
+import org.folio.inventory.domain.items.ItemStatusName;
+import org.folio.inventory.support.http.client.IndividualResource;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import api.support.ApiTests;
+import api.support.builders.HoldingRequestBuilder;
+import api.support.builders.ItemRequestBuilder;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+
+/**
+ * This test verifies that ramls/item.json and {@link ItemStatusName} is consistent.
+ * <p>
+ * If you're introducing a new status for an item you have to update both item.json
+ * and the {@link ItemStatusName} enum, otherwise item won't pass validation
+ * and will be rejected.
+ */
+@RunWith(JUnitParamsRunner.class)
+public class ItemAllowedStatusesSchemaTest extends ApiTests {
+
+  @Test
+  public void schemaAndEnumIsConsistent() throws IOException {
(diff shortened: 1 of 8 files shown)


src/test/java/api/items/ItemAllowedStatusesSchemaTest.java
```diff
@@ -0,0 +1,103 @@
...
+  @Parameters(method = "getItemStatusNameEnumAllowedItemStatuses")
+  public void canCreateItemsWithAllStatuses(String itemStatus) throws Exception {
+    final UUID holdingsId = createInstanceAndHolding();
+
+    final IndividualResource createResponse = itemsClient.create(
+      new ItemRequestBuilder()
+        .forHolding(holdingsId)
+        .canCirculate()
+        .withStatus(itemStatus)
+    );
```
marcjohnson-kint: Personally I'd close the method call on the same line as `withStatus`

src/test/java/api/items/ItemAllowedStatusesSchemaTest.java
```diff
@@ -0,0 +1,103 @@
...
+ * <p>
+ * If you're introducing a new status for an item you have to update both item.json
+ * and the {@link ItemStatusName} enum, otherwise item won't pass validation
+ * and will be rejected.
+ */
+@RunWith(JUnitParamsRunner.class)
+public class ItemAllowedStatusesSchemaTest extends ApiTests {
+
+  @Test
+  public void schemaAndEnumIsConsistent() throws IOException {
```
marcjohnson-kint: Does this demonstrate that the API allows for items to be created with each of the statuses in the interface schema?

src/test/java/api/support/ApiTests.java
```diff
@@ -13,7 +13,7 @@
 
 public abstract class ApiTests {
   private static boolean runningOnOwn;
-  protected final OkapiHttpClient okapiClient;
+  protected static OkapiHttpClient okapiClient;
```
marcjohnson-kint: Why does this need to change to be static?

src/test/java/api/items/ItemAllowedStatusesSchemaTest.java
bohdan-suprun: This tests verifies that item allowed statuses from the `item.json` are the same as in `org.folio.inventory.domain.items.ItemStatusName` enum. We need to keep them sync, otherwise validation will reject item create request, if there is a status in item.json but missed in the enum class.

bohdan-suprun: `canCreateItemsWithAllStatuses` in this file demonstrates that behavior.

src/test/java/api/support/ApiTests.java
bohdan-suprun: I need a to use a parameterized test, but it doesn't work when initializing `okapiClient` in constructor (NPE on class init step). Have to initialize it in @BeforeClass method, which operates with static context only.

src/test/java/api/items/ItemAllowedStatusesSchemaTest.java
marcjohnson-kint: Why do we need both `schemaAndEnumIsConsistent` and `canCreateItemsWithAllStatuses`, it seems like they do nearly the same thing?

bohdan-suprun: The `canCreateItemsWithAllStatuses` uses statuses from enum only, but `schemaAndEnumIsConsistent` verifies that the statuses are the same.

marcjohnson-kint: What do you think to the alternative of creating an item with each status from the schema in one test?

bohdan-suprun: Hmm... well the main goal of the test is to prevent a case when someone wanted to add a new status for item but forgot/missed to update either schema or enum.

I don't think if it works.

marcjohnson-kint: Ok, if the two separate tests mean that one will fail if either of them are missed out, then lets keep both of them

PR #170, opened 2020-02-10
MODINV-200: Prevent claimed returned item to mark as missing

bohdan-suprun:
https://issues.folio.org/browse/MODINV-200: BE: Claimed returned: item record actions

* Disallow `Claimed returned` item to be marked as `Missing`;


src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -186,17 +188,14 @@ private void update(RoutingContext routingContext) {
...
-        if (!Objects.equals(newItem.getHrid(), oldItem.getHrid())) {
-          log.warn("The HRID property can not be updated, old value is '{}' but new is '{}'",
-            oldItem.getHrid(), newItem.getHrid()
-          );
-
-          JsonResponse.unprocessableEntity(routingContext.response(),
-            "HRID can not be updated", "hrid", newItem.getHrid()
-          );
+        Optional<ValidationError> updateValidationErrors =
+          validateItemUpdate(oldItem, newItem);
```
marcjohnson-kint: Has `Optional` been used in other places to detect the presence of errors or not?

src/main/java/org/folio/inventory/validation/ItemUpdateValidator.java
```diff
@@ -0,0 +1,45 @@
...
+    if (!Objects.equals(newItem.getHrid(), oldItem.getHrid())) {
+      validationError = new ValidationError(
+        "HRID can not be updated", "hrid", newItem.getHrid());
+    }
+
+    if (claimedReturnedItemMarkedMissing(oldItem, newItem)) {
+      validationError = new ValidationError(
+        "Claimed returned item can not be marked as missing",
+        "status.name", ItemStatusName.MISSING.value());
+    }
```
marcjohnson-kint: I think it makes sense to try to separate these two cases into separate methods

src/main/java/org/folio/inventory/validation/ItemUpdateValidator.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+import org.folio.inventory.domain.items.Item;
+import org.folio.inventory.domain.items.ItemStatusName;
+import org.folio.inventory.support.http.server.ValidationError;
+
+import io.vertx.ext.web.RoutingContext;
+
+public final class ItemUpdateValidator {
```
marcjohnson-kint: I think we need a better name for this, maybe this could be changed to describe detecting disallowed changes?

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
```diff
@@ -130,8 +130,8 @@ protected JsonObject mapToRequest(Item item) {
       ? item.id
       : UUID.randomUUID().toString());
 
-    if(item.getStatus().getString(Status.NAME_KEY) != null) {
-      itemToSend.put("status", item.getStatus());
+    if(item.getStatus().getName() != null) {
+      itemToSend.put("status", item.getStatus().getJson());
```
marcjohnson-kint: Rather than using `getJson` here, maybe we could do the serialization in this class instead?

src/test/java/api/items/ItemApiExamples.java
```diff
@@ -1589,6 +1585,36 @@ public void statusDatePropertyPresentOnStatusUpdated()
...
+    String itemStatus = createdItem.getJson().getJsonObject("status")
+      .getString("name");
+    assertThat(itemStatus, is("Claimed returned"));
+
+    JsonObject updatedItem = createdItem.getJson().copy()
+      .put("status", new JsonObject().put("name", "Missing"));
+
+    Response updateResponse = updateItem(updatedItem);
+    assertThat(updateResponse,
+      hasValidationError("Claimed returned item can not be marked as missing",
```
marcjohnson-kint: Should this say `cannot` instead of `can not`?

```suggestion
      hasValidationError("Claimed returned item cannot be marked as missing",
```

src/test/java/org/folio/inventory/storage/external/ExternalItemCollectionExamples.java
```diff
@@ -173,7 +173,7 @@ public void anItemCanBeUpdated()
     assertThat(updated.getTemporaryLocationId(), is(added.getTemporaryLocationId()));
     assertThat(updated.getMaterialTypeId(), is(added.getMaterialTypeId()));
     assertThat(updated.getPermanentLoanTypeId(), is(added.getPermanentLoanTypeId()));
-    assertThat(updated.getStatus().getString(Status.NAME_KEY), is(ItemStatusName.CHECKED_OUT.value()));
+    assertThat(updated.getStatus().getName().value(), is("Checked out"));
```
marcjohnson-kint: Can this compare the status names' instead of the text value of them?

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
```diff
@@ -130,8 +130,8 @@ protected JsonObject mapToRequest(Item item) {
       ? item.id
       : UUID.randomUUID().toString());
 
-    if(item.getStatus().getString(Status.NAME_KEY) != null) {
-      itemToSend.put("status", item.getStatus());
+    if(item.getStatus().getName() != null) {
```
marcjohnson-kint: Should this stop the status being written out?

What happens if the item has no status when it reaches storage?

src/main/java/org/folio/inventory/validation/ItemUpdateValidator.java
```diff
@@ -0,0 +1,45 @@
...
+        "Claimed returned item can not be marked as missing",
+        "status.name", ItemStatusName.MISSING.value());
+    }
+
+    return Optional.ofNullable(validationError);
+  }
+
+  private static boolean claimedReturnedItemMarkedMissing(Item oldItem, Item newItem) {
+    return oldItem.getStatus().getName() == ItemStatusName.CLAIMED_RETURNED
+      && newItem.getStatus().getName() == ItemStatusName.MISSING;
```
marcjohnson-kint: Sonar is suggesting partial coverage on this line, could it be updated to get complete coverage?

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
bohdan-suprun: It is a redundant check I think, API should require status to be set.

src/main/java/org/folio/inventory/resources/Items.java
bohdan-suprun: No, I have not seen similar approach. Actually, there is no established approach for validation within mod-inventory.

marcjohnson-kint: Are you aware of one in other modules that EPAM folks are working on?

(The `Result` based approach in mod-circulation appears to be relatively unusual)

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
marcjohnson-kint: Can there be a check in the item that enforces that?

marcjohnson-kint: (This is why I like languages with non-nullable types)

src/main/java/org/folio/inventory/validation/ItemUpdateValidator.java
bohdan-suprun: Do you mean to separate and execute them in this method, or in the resource's one?

My intent was to move the logic to separate class, because it grows and the `Items` class has already had ~1k lines.

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
bohdan-suprun: Serialization is performed in two places: here and in `ItemRepresentation`. Should I create a representation class instead and move serialization to this class?

src/main/java/org/folio/inventory/validation/ItemUpdateValidator.java
marcjohnson-kint: I was thinking separate methods within this class, to make that method read well at a higher level of abstraction.

I'm happy for it to be in this class (I think it needs a better name). Thank you for extracting this from the `Items` class.

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
marcjohnson-kint: I think I made a mistake trying to combine those too.

I think we want to head in the direction of serialization to storage and serialization back to the client as two separate activities.

I think the serlialization to storage can be included here, or use it's own class (I likely prefer the latter)

src/main/java/org/folio/inventory/resources/Items.java
bohdan-suprun: Some of modules uses Future-based validation, it is similar to `Result` approach but validation failures/record not found/etc is a sub-class of an exception.

marcjohnson-kint: Yeah, I was trying to avoid `exceptions` with `Result`, it didn't quite work out that way. I was also trying to keep asynchronicity separate as well, with mixed results.

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
bohdan-suprun: Do you mean to add a check to Item constructor that will throw an exception in case no status provided?

bohdan-suprun:
Hi @marcjohnson-kint, I've pushed validation for the update endpoint could you please take a look.

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
marcjohnson-kint: Yes, effectively, I mean stop construction of items without a status. 

(If we had a language that supported it, I'd make it a closed or non-nullable type)

bohdan-suprun: I have added `requireNonNull` to the constructor of Item and Status. It is not possible to create both item or status without status and status.name now.

Is it fine?

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
bohdan-suprun: I've extracting serialization and deserialization to separate class.

src/main/java/org/folio/inventory/validation/ItemUpdateValidator.java
bohdan-suprun: I've renamed it to `ItemsValidator` and added methods that describes validation rule being asserted.

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleItemCollection.java
marcjohnson-kint: I think that's fine

PR #174, opened 2020-03-06
MODINV-204 Implemented InstanceLoader and handler for matching of Instances

KaterynaSenchenko:
* Created an endpoint that receives events related to data-import process
* Implemented handler that performs matching of Instances
* Implemented MatchValueLoader to support lnstance loading. Loader receives LoadQuery constructed by data-import-processing-core library and queries the mod-inventory-storage. Depending on the search result, a particular eventType is set to eventPayload (DI_INVENTORY_INSTANCE_MATCHED, DI_INVENTORY_INSTANCE_NOT_MATCHED, or DI_ERROR in case there were errors)
* A new event is sent as a result of handling (performed by data-import-processing-core)
* New functionality covered with tests


dcrossleyau:
The RAML error is being reported for `inventory-event-handlers.raml` file
because the line 23 is missing:
```
        body:
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -189,6 +189,24 @@
...
+      "id": "inventory-event-handlers",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/inventory/handlers/created-srs-marc-bib-record",
+          "permissionsRequired": [
+            "inventory.events.post"
```
marcjohnson-kint: My understanding is that this permission is required for any client invoking this endpoint. How will mod-pubsub, which I believe will be invoking it, be granted that permission?

OleksiiKuzminov: https://github.com/folio-org/mod-pubsub#permissions

marcjohnson-kint: @OleksiiKuzminov Thank you for that information.

Permissions are usually specific to an endpoint, why is this generalised to any events handling endpoint?

marcjohnson-kint: That information suggests that a special pub sub user needs to be set up with the appropriate permissions, for each tenant. Meaning that someone needs to know what about all of the event subscribers in the system.

Have I followed that documentation correctly?

OleksiiKuzminov: For the first version of the securing mechanism, you are right. But we have an umbrella and spike to investigate the best way of securing HTTP calls through pubsub. For now, we create a special user for pubsub and register it TenantApi. After it, we assign all pub-sub handlers permissions for this user

src/main/java/org/folio/inventory/dataimport/handlers/matching/InstanceLoader.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+    return future.join();
+  }
+
+  @Override
+  public boolean isEligibleForEntityType(EntityType entityType) {
+    return EntityType.INSTANCE == entityType;
+  }
+
+  private Context constructContext(String tenantId, String token, String okapiUrl) {
```
VRohach: @KaterynaSenchenko I think it would be better to extract this method to the Util-class. Because it will use in a lot of places in the future with the same implementation.

KaterynaSenchenko:
Hi @adamdickmeiss,  @marcjohnson-kint, thank you for your review. We increased the coverage on the new code and would like to merge this PR in order to proceed with other remaining work in mod-inventory that should be included in the follow-on release. If you still have any questions or concerns, please contact me or @OleksiiKuzminov in slack or in subsequent PRs to mod-inventory (as we will add other handlers).

marcjohnson-kint:
> Hi @adamdickmeiss, @marcjohnson-kint, thank you for your review. 

As agreed in the first meeting, I won't be participating in this work. Apologies, I did let my curiosity get the better of me a couple of times with some questions :-/

PR #177, opened 2020-03-11
MODINV-212: Use partitions to fetch instance relationships/preceding-succeeding titles

bohdan-suprun:
https://issues.folio.org/browse/MODINV-212 - Searches returning more than 50 instances hangs.

### Approach:
* Use partition of 30 instances to retrieve relationships and preceding-succeeding titles.
* Add /emulate-failure endpoint to fake storage modules to test that fetch failure is propagated to client.


src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -326,57 +336,28 @@ private void getById(RoutingContext routingContext) {
    *
    * @param instancesResponse Multi record Instances result
    * @param routingContext
-   * @param context
    */
-  private CompletableFuture<InstancesResponse> fetchRelationships(
+  private Future<InstancesResponse> fetchRelationships(
```
marcjohnson-kint: Why has this code switched from using `CompletableFuture` to Vert.x `Future`?

src/test/java/support/fakes/EndpointFailureDescriptor.java
```diff
@@ -0,0 +1,46 @@
+package support.fakes;
+
+import java.util.Date;
+
+public class EndpointFailureDescriptor {
```
marcjohnson-kint: What is an endpoint failure descriptor?

src/main/java/org/folio/inventory/resources/Instances.java
bohdan-suprun: The reason is to be able to propagate failure occurred during fetching relationships/preceding-succeeding titles to client. I'm not sure if it is possible with `CompletableFuture`.

src/test/java/support/fakes/FakeStorageModule.java
```diff
@@ -87,6 +91,31 @@ void register(Router router) {
 
     router.get(rootPath + "/:id").handler(this::get);
     router.delete(rootPath + "/:id").handler(this::delete);
+    router.post(rootPath + "/emulate-failure").handler(this::emulateFailure);
+  }
+
+  private void emulateFailureIfNeeded(RoutingContext routingContext) {
```
marcjohnson-kint: Why is it necessary to emulate a failure with a special endpoint?

src/main/java/org/folio/inventory/resources/Instances.java
marcjohnson-kint: What do you mean by propagate failure? Do you mean allow the Vert.x error handler to pick up exception?

src/test/java/support/fakes/EndpointFailureDescriptor.java
bohdan-suprun: I have introduced new endpoint for fakes, it is used to emulate resource failures during some period. This feature is used to test that relationships fetch failures are forwarded to client.

src/test/java/support/fakes/FakeStorageModule.java
bohdan-suprun: It is used to test that when we searching for instances and an error occurred with sub-calls (e.g. to retrieve relationships or preceding-succeeding titles) then it have to be handled in some way.

Unfortunately, it can not be tested just with unit tests, because the logic is quite encapsulated, so no way to test just a piece of functionality. 

(My opinion, is that we need a refactoring for the instances endpoint).

src/main/java/org/folio/inventory/storage/external/MultipleRecordsFetchClient.java
```diff
@@ -0,0 +1,98 @@
...
+
+import org.folio.inventory.storage.external.exceptions.ExternalResourceFetchException;
+import org.folio.inventory.support.http.client.Response;
+
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+public class MultipleRecordsFetchClient {
+  private static final int DEFAULT_PARTITION_SIZE = 30;
```
Svitlana-Zmiivska: Why does DEFAULT_PARTITION_SIZE equal 30?

src/test/java/support/fakes/FakeStorageModule.java
marcjohnson-kint: > It is used to test that when we searching for instances and an error occurred with sub-calls (e.g. to retrieve relationships or preceding-succeeding titles) then it have to be handled in some way.

Given that this is being tested from the outside, how does the system know to use the new endpoint rather than the relevant preceding / succeeding titles endpoint?

> Unfortunately, it can not be tested just with unit tests, because the logic is quite encapsulated, so no way to test just a piece of functionality.
> 
> (My opinion, is that we need a refactoring for the instances endpoint).

Please raise a tech debt issue for this

src/main/java/org/folio/inventory/resources/Instances.java
bohdan-suprun: I have reverted it to use `CompletableFuture` it also supports the error propagation.

Yes, I mean to not execute all further calls in the chain if on some step an exception has been raised and also have a common method which will be executed regardless failure or success, to generate response to client.

marcjohnson-kint: > I have reverted it to use `CompletableFuture` it also supports the error propagation.

Thank you.

> Yes, I mean to not execute all further calls in the chain if on some step an exception has been raised and also have a common method which will be executed regardless failure or success, to generate response to client.

Yes, if you haven't already (I haven't reviewed the changes yet), please add an `exceptionally` on to the end of the chain to forward any uncaught exception.

src/test/java/support/fakes/FakeStorageModule.java
bohdan-suprun: > Given that this is being tested from the outside, how does the system know to use the new endpoint rather than the relevant preceding / succeeding titles endpoint?

It is not a new endpoint, it is just a sub-path to existing endpoint, so the whole path will be for example `/preceding-succeeding-titles/emulate-failure`. And steps to test a failure case might be following:
* Do some preparation (e.g. create instances/relationships);
* Enable failure emulation for the API for some time period, by sending request to the mentioned URI;
* Attempt call the API that needs to be tested;
* Verify response;
* In an `@After` method do disable of failure emulation.

src/main/java/org/folio/inventory/storage/external/MultipleRecordsFetchClient.java
bohdan-suprun: The jira says that the failure occurred when there are more than 50 instances, so I decided to chose a middle value.

Do you think it should be changed?

src/main/java/org/folio/inventory/services/InstanceRelationshipsService.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+
+import org.folio.inventory.storage.external.CollectionResourceClient;
+import org.folio.inventory.storage.external.CqlQuery;
+import org.folio.inventory.storage.external.MultipleRecordsFetchClient;
+
+import io.vertx.core.Future;
+import io.vertx.core.json.JsonObject;
+
+public class InstanceRelationshipsService {
```
Svitlana-Zmiivska: How about removing duplication? Work with only one fetchClient and pass relationshipsFetchClient'/precedingSucceedingFetchClient as a parameter to the  constructor?

src/test/java/support/fakes/FakeCQLToJSONInterpreter.java
```diff
@@ -76,10 +79,12 @@ private boolean containsSort(ImmutablePair<String, String> queryAndSort) {
 
     String splitRegex;
     BinaryOperator<Predicate<JsonObject>> accumulator;
-    //TODO Need to implement operator 'or' for query which contains search by different fields.
-    //This is just a temporary solution.
-    if (query.matches(PRECEDING_SUCCEEDING_URL_REGEX)) {
-      splitRegex = " or ";
+
+    final Matcher matcher = Pattern.compile(OR_REGEX)
```
Svitlana-Zmiivska: How about moving Pattern object to a field of the class? It seems to me we don't need to create pattern object every time when this method is called

src/test/java/support/fakes/FakeStorageModule.java
marcjohnson-kint: To me, that is a new endpoint.

Oh, this endpoint is used to control when other endpoints in that fake report failures. Is that a sensible understanding?

src/main/java/org/folio/inventory/services/InstanceRelationshipsService.java
bohdan-suprun: All these fetch client requires different CQL generation methods. So it is not possible to re-use code. Also, the intent of this class to hide the fetching logic.

Is it fine?

src/test/java/support/fakes/FakeStorageModule.java
bohdan-suprun: Yes, correct. It is not ideal, I think we should consider a better approach, but it is an important case to verify it.

bohdan-suprun: > Please raise a tech debt issue for this.

Here it is https://issues.folio.org/browse/MODINV-213

PR #179, opened 2020-03-17
MODINV-184 - Implement action handler for Create Item action

RuslanLavrov:
https://issues.folio.org/browse/MODINV-184

### Approach:
* implemented action handler for Create Item action.
*  published event about Item creation using data-import-processing-core

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateItemEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateItemEventHandler.java
@@ -22,14 +22,15 @@ import org.folio.inventory.support.JsonHelper;
 import org.folio.processing.events.services.handler.EventHandler;
 import org.folio.processing.exceptions.EventProcessingException;
 import org.folio.processing.mapping.MappingManager;
+import org.folio.rest.jaxrs.model.EntityType;
 
-import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.UUID;
@@ -43,6 +44,7 @@ import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ContentType.ACTI
 public class CreateItemEventHandler implements EventHandler {
 
   public static final String ITEM_CREATED_EVENT_TYPE = "DI_INVENTORY_ITEM_CREATED";
+  private static final String PAYLOAD_HAS_NO_DATA_MSG = "Failed to handle event payload, cause event payload context does not contain MARC_BIBLIOGRAPHIC data";
 
   private static final Logger LOG = LoggerFactory.getLogger(CreateItemEventHandler.class);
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateItemEventHandler.java
```diff
@@ -0,0 +1,162 @@
...
+    ArrayList<String> errorMessages = new ArrayList<>();
+    for (String fieldPath : requiredFields) {
+      String field = StringUtils.substringBefore(fieldPath, ".");
+      String nestedField = StringUtils.substringAfter(fieldPath, ".");
+      if (!isExistsRequiredProperty(itemAsJson, field, nestedField)) {
+        errorMessages.add(String.format("Field '%s' is a required field and can not be null", fieldPath));
+      }
+    }
+    String statusName = JsonHelper.getNestedProperty(itemAsJson, "status", "name");
+    if (StringUtils.isNotBlank(statusName) && !ItemStatusName.isStatusCorrect(statusName)) {
```
bohdan-suprun: Status is also required field, so `status == null` is not allowed.

src/test/java/org/folio/inventory/dataimport/handlers/actions/CreateItemEventHandlerTest.java
```diff
@@ -0,0 +1,331 @@
...
+  @Mock
+  private Storage mockedStorage;
+  @Mock
+  private ItemCollection mockedItemCollection;
+  @Mock
+  private Reader fakeReader;
+  @Spy
+  private MarcBibReaderFactory fakeReaderFactory = new MarcBibReaderFactory();
+
+  protected JobProfile jobProfile = new JobProfile()
```
bohdan-suprun:
```suggestion
  private JobProfile jobProfile = new JobProfile()
```

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateItemEventHandler.java
RuslanLavrov: if nested field ` name != null ` then ` status != null ` as well, therefore Item with status == null will not be created

src/test/java/org/folio/inventory/dataimport/handlers/actions/CreateItemEventHandlerTest.java
RuslanLavrov: thank you

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateItemEventHandler.java
bohdan-suprun: Yep, sorry, I meant that `status.name == null` is also unacceptable. See https://github.com/folio-org/mod-inventory-storage/blob/master/ramls/item.json#L245.

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateItemEventHandler.java
```diff
@@ -0,0 +1,170 @@
...
+  @Override
+  public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload dataImportEventPayload) {
+    CompletableFuture<DataImportEventPayload> future = new CompletableFuture<>();
+    try {
+      HashMap<String, String> payloadContext = dataImportEventPayload.getContext();
+      if (payloadContext == null || StringUtils.isBlank(payloadContext.get(EntityType.MARC_BIBLIOGRAPHIC.value()))) {
+        LOG.error(PAYLOAD_HAS_NO_DATA_MSG);
+        future.completeExceptionally(new EventProcessingException(PAYLOAD_HAS_NO_DATA_MSG));
+        return future;
+      }
```
VRohach: @RuslanLavrov please add some tests for this case

RuslanLavrov: already added

PR #181, opened 2020-03-19
MODINV-182: Implement action handler for Create Instance action

VRohach:
JIRA-ticket: https://issues.folio.org/browse/MODINV-182


VRohach:
Tests will be added soon.

PR #183, opened 2020-03-23
MODINV-215 Exception handling for /inventory/instances endpoints

Svitlana-Zmiivska:
https://issues.folio.org/browse/MODINV-215

**Purpose:**

Need to handle exception for /inventory/instances endpoints in order to understand what real happens.

**Approach:**

- add failure handle during manipulation with preceding/succeeding titles or instance relationships
- add constraint check for instance relationships and preceding-succeeding title APIs.

### DIFF
--- a/src/main/java/org/folio/inventory/storage/external/CollectionResourceRepository.java
+++ b/src/main/java/org/folio/inventory/storage/external/CollectionResourceRepository.java
@@ -1,29 +1,28 @@
 package org.folio.inventory.storage.external;
 
 import org.folio.inventory.storage.external.exceptions.ExternalResourceFetchException;
-import org.folio.inventory.support.http.client.OkapiHttpClient;
 import org.folio.inventory.support.http.client.Response;
 
-import java.net.URL;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
 
-public class CollectionResourceRepository extends CollectionResourceClient{
+public class CollectionResourceRepository {
 
-  public CollectionResourceRepository(OkapiHttpClient client, URL collectionRoot) {
-    super(client, collectionRoot);
+  private CollectionResourceClient resourceClient;
+
+  public CollectionResourceRepository(CollectionResourceClient resourceClient) {
+    this.resourceClient = resourceClient;
   }
 
   public CompletableFuture<Response> post(Object resourceRepresentation) {
     CompletableFuture<Response> future = new CompletableFuture<>();
-    post(resourceRepresentation, future::complete);
+    resourceClient.post(resourceRepresentation, future::complete);
 
     return future.thenCompose(response ->  handleResponse(response, 201));
   }
 
   public CompletableFuture<Response> put(String id, Object resourceRepresentation) {
     CompletableFuture<Response> future = new CompletableFuture<>();
-    put(id, resourceRepresentation, future::complete);
+    resourceClient.put(id, resourceRepresentation, future::complete);
 
     return future.thenCompose(response ->  handleResponse(response, 204));
   }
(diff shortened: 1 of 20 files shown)


src/main/java/org/folio/inventory/storage/external/CollectionResourceRepository.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.folio.inventory.storage.external.exceptions.ExternalResourceFetchException;
+import org.folio.inventory.support.http.client.OkapiHttpClient;
+import org.folio.inventory.support.http.client.Response;
+
+import java.net.URL;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+public class CollectionResourceRepository extends CollectionResourceClient{
```
bohdan-suprun: I'm not sure if we should use extension here.  As I understand the `CollectionResourceRepository ` provides an alternative way of interaction with endpoints, so I'd use delegation here.

src/main/java/org/folio/inventory/storage/external/CollectionResourceRepository.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public CompletableFuture<Response> delete(String id) {
+    CompletableFuture<Response> future = new CompletableFuture<>();
+
+    delete(id, future::complete);
+
+    return future.thenCompose(response -> handleResponse(response, 204));
+  }
+
+  private CompletionStage<Response> handleResponse(Response response,
```
bohdan-suprun:
```suggestion
  private CompletableFuture<Response> handleResponse(Response response,
```

src/test/java/api/InstanceRelationshipsTest.java
```diff
@@ -136,6 +141,299 @@ public void canFetchMultipleInstancesWithAllRelationshipsTypes() throws Exceptio
...
+  @Test
+  public void cannotCreateAnInstanceWithNonExistedPrecedingTitleId()
+    throws InterruptedException, MalformedURLException, TimeoutException,
+    ExecutionException {
+
+    String precedingTitleId = UUID.randomUUID().toString();
+    final JsonObject expectedErrorResponse = new JsonObject()
+      .put("message", String.format("Cannot set preceding_succeeding_title.precedinginstanceid = " +
+        "%s because it does not exist in instance.id.", precedingTitleId));
+    precedingSucceedingTitlesClient.emulateFailure(
```
bohdan-suprun: There is record-processor approach for FakeOkapiStorage modules (e.g. support.fakes.processors.StorageRecordPreProcessors#setEffectiveCallNumberComponents), I'd use them in this case, instead of failure emulation. 

What do you think?

src/test/java/api/InstanceRelationshipsTest.java
```diff
@@ -136,6 +141,299 @@ public void canFetchMultipleInstancesWithAllRelationshipsTypes() throws Exceptio
...
+    JsonObject smallAngryPlanetJson = smallAngryPlanet(UUID.randomUUID());
+    smallAngryPlanetJson.put("childInstances", childInstances);
+
+    Response response = instancesClient.attemptToCreate(smallAngryPlanetJson);
+
+    assertThat(response.getStatusCode(), is(422));
+    assertThat(response.getContentType(), is("application/json"));
+    assertThat(response.getJson(), is(expectedErrorResponse));
+  }
+
```
bohdan-suprun: Could you please add a test that verifies that we can not create an instance with non existed relationship type id?

src/test/java/api/InstanceRelationshipsTest.java
```diff
@@ -136,6 +141,299 @@ public void canFetchMultipleInstancesWithAllRelationshipsTypes() throws Exceptio
...
+  @Test
+  public void canForwardInstancePrecedingSucceedingTitlesUpdateFailure()
```
bohdan-suprun: Is it possible to use parametrized test for the `canForwardInstancePrecedingSucceedingTitles*` tests?

src/main/java/org/folio/inventory/storage/external/CollectionResourceRepository.java
marcjohnson-kint: Agreed, my understanding of the intention of a repository is that it provides a higher level abstraction over the client.

What benefits does inheriting from the client instead of using the client have?

PR #185, opened 2020-03-25
MODINV-216: Implement mark item withdrawn endpoint.

bohdan-suprun:
https://issues.folio.org/browse/MODINV-216 - Mark item withdrawn

### Changes:
* Introduce `/inventory/items/:id/mark-withdrawn` API endpoint:
   * Marks an item as withdrawn;
   * If there is unexpired request in fulfillment (request with `Open - *` status other then `Open - Not yet filled`), changes it's status to `Open - Not yet filled`.


marcjohnson-kint:
I am marking this do not merge because no functional changes should be merged to mod-inventory until the exceptional release has been made which includes FoliJet's data import work

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -87,6 +89,20 @@ public void register(Router router) {
...
+
+  private CompletableFuture<Void> markAsWithdrawn(
+    RoutingContext routingContext, WebContext webContext, Clients clients) {
+
+    final ItemService itemService = new ItemService(storage
+      .getItemCollection(webContext), clients);
+
+    return itemService.processMarkItemWithdrawn(webContext)
+      .thenAccept(item ->
+        respondWithItemRepresentation(item, 201, routingContext, webContext));
```
Svitlana-Zmiivska: How about using the predefined constant?  e.x. HttpStatus.HTTP_CREATED.toInt()

src/main/java/org/folio/inventory/services/ItemService.java
```diff
@@ -0,0 +1,68 @@
...
+      .thenCompose(itemCollection::update);
+  }
+
+  private CompletableFuture<Item> updateRequestStatusIfRequired(Item item) {
+    final CqlQuery query = exactMatch("itemId", item.id)
+      .and(exactMatch("status", RequestStatus.OPEN_AWAITING_PICKUP.getValue()));
+
+    return requestStorageClient.getMany(query, 1, 0)
+      .thenApply(response -> {
+        final JsonArray requests = response.getJson().getJsonArray("requests");
```
Svitlana-Zmiivska: How about moving out this code to a separate method in order to increase readability?

src/main/java/org/folio/inventory/services/ItemService.java
```diff
@@ -0,0 +1,68 @@
...
+        return requests != null && requests.size() > 0
+          ? new StoredRequestView(requests.getJsonObject(0).getMap())
+          : null;
+      }).thenCompose(request -> {
+        if (request != null && request.getHoldShelfExpirationDate() != null
```
Svitlana-Zmiivska: How about moving out this code to a separate method in order to increase readability?

descriptors/ModuleDescriptor-template.json
```diff
@@ -263,7 +289,7 @@
   "requires": [
     {
       "id": "item-storage",
-      "version": "8.1"
+      "version": "8.3"
```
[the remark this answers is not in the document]
bohdan-suprun: It is "requires" section, I need item-storage of 8.3 version where the Withdrawn status is supported.

src/main/java/org/folio/inventory/services/ItemService.java
bohdan-suprun: Refactored, it should be more clearer now.

src/main/java/org/folio/inventory/services/ItemService.java
bohdan-suprun: Refactored, it should be more clearer now.

marcjohnson-kint:
@bohdan-suprun Is there a sequence to the two pull requests raised to for MODINV-216, does one need to go before the other?

bohdan-suprun:
@marcjohnson-kint, no, they can be reviewed independently, but have to be merged together.

marcjohnson-kint:
> @marcjohnson-kint, no, they can be reviewed independently, but have to be merged together.

What do you mean by merged together? 

Given that we are using the preview functionality for the withdrawal of an item feature, which pull request is being used in that environment?

bohdan-suprun:
@marcjohnson-kint, The whole work was split into 2 PRs, the first PR `MODINV-216: Refactor inventory to support request-storage API.` is `MODINV-216-require-request-storage-api` branch forked from `master`.

The second PR (this one) is `MODINV-216-add-mark-withdrawn-api` branch for which `MODINV-216-require-request-storage-api` is base branch.

Then I need to get approval for the changes and merge `MODINV-216-add-mark-withdrawn-api` into `MODINV-216-require-request-storage-api` and then merge it into `master`.

Does it sounds fine?

marcjohnson-kint:
> Then I need to get approval for the changes and merge `MODINV-216-add-mark-withdrawn-api` into `MODINV-216-require-request-storage-api` and then merge it into `master`.
> 
> Does it sounds fine?

Why does MODINV-216-add-mark-withdrawn-api need merging into MODINV-216-require-request-storage-api?

This seems a little in reverse to me, as you described the latter as being the base for the former above.

I really like the idea of smaller batches of changes in the pull requests, and I'm really glad you did that. I hope it helped during the initial work.

I think the value of small pull requests is that they can be merged separately, reducing the latency to getting changes into master and ideally producing more cohesive merges.

However, I think that goal is is somewhat counter the preview process goals (large, long-lived branches is one of the reasons I have reservations about the process). I think this is an excellent topic for us to reflect upon after the pilot. 

Have I interpreted your intent correctly? Was there another reason for the separate pull requests?

Personally, I might close MODINV-216-require-request-storage-api as it is already included in MODINV-216-add-mark-withdrawn-api and I think it is simpler to merge that at the end of the preview, rather than trying to work back out through the other branch.

Does that make sense?

bohdan-suprun:
Yes, the goal was to have small PRs, to simplify code review. 

We can close the `MODINV-216-require-request-storage-api`.

marcjohnson-kint:
> Yes, the goal was to have small PRs, to simplify code review.

Personally, I love that goal. I appreciate you doing it. I'm sorry I couldn't better support it on this occasion.

Unfortunately, I think it came about at a challenging time for reviews (the team shrank and we were in the midst of the quarterly release) and that it may not fit well with the new preview process.

> We can close the `MODINV-216-require-request-storage-api`.

bohdan-suprun:
@marcjohnson-kint, closed, this PR contains all the changes now.

marcjohnson-kint:
> @marcjohnson-kint, closed, this PR contains all the changes now.

Thanks. Please could the outstanding code smells be resolved.

src/main/java/org/folio/inventory/support/CompletableFutures.java
```diff
@@ -0,0 +1,16 @@
+package org.folio.inventory.support;
+
+import java.util.concurrent.CompletableFuture;
+
+public final class CompletableFutures {
```
julianladisch, line 5: This should be marked as deprecated with a comment saying "When we switch to Java 11 remove this class and use CompletableFuture#failedFuture(Throwable) instead. https://stackoverflow.com/a/49432327"

descriptors/ModuleDescriptor-template.json
```diff
@@ -40,7 +40,33 @@
             "inventory-storage.instances.collection.get",
             "inventory-storage.instances.item.get"
           ]
-        }, {
+        },
+        {
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/inventory/items/{id}/mark-withdrawn",
```
marcjohnson-kint, line 48: Could this be shortened to `withdraw`, is the marking aspect significant?

```suggestion
          "pathPattern": "/inventory/items/{id}/withdraw",
```

src/main/java/org/folio/inventory/domain/view/request/StoredRequestView.java
```diff
@@ -0,0 +1,41 @@
+package org.folio.inventory.domain.view.request;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import org.joda.time.DateTime;
+
+public class StoredRequestView {
```
marcjohnson-kint, line 10: What is a stored request view?

src/main/java/org/folio/inventory/resources/AbstractInventoryResource.java
```diff
@@ -0,0 +1,61 @@
...
+import org.folio.inventory.storage.external.Clients;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.vertx.core.Handler;
+import io.vertx.core.http.HttpClient;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+
+public abstract class AbstractInventoryResource {
```
marcjohnson-kint, line 20: Why does this need to be a abstract class?

Is the intent to re-use the handle method? If so, could it instead be included in it's own class, maybe something like `DefaultHandler`?

src/main/java/org/folio/inventory/services/ItemService.java
```diff
@@ -0,0 +1,88 @@
...
+import org.folio.inventory.storage.external.Clients;
+import org.folio.inventory.storage.external.CollectionResourceRepository;
+import org.folio.inventory.storage.external.CqlQuery;
+import org.folio.inventory.validation.ItemMarkAsWithdrawnValidators;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ItemService {
```
marcjohnson-kint, line 28: Could this service be specific only to withdrawing an item? What makes it more general than that?

src/main/java/org/folio/inventory/support/CompletableFutures.java
marcjohnson-kint: @julianladisch When might FOLIO be upgrading to Java 11?

I've put deprecation warnings in other parts of the system, and they mostly generate noise if there isn't intent to replace the behaviour fairly quickly.

src/main/java/org/folio/inventory/support/http/server/ForwardResponse.java
```diff
@@ -3,28 +3,33 @@
...
+    forwardTo.setStatusCode(statusCode);
 
-    if(forwardFrom.hasBody()) {
-      Buffer buffer = Buffer.buffer(forwardFrom.getBody(), "UTF-8");
+    if (StringUtils.isNotBlank(body)) {
+      Buffer buffer = Buffer.buffer(body, "UTF-8");
 
-      forwardTo.putHeader(HttpHeaders.CONTENT_TYPE, forwardFrom.getContentType());
+      forwardTo.putHeader(HttpHeaders.CONTENT_TYPE, contentType);
       forwardTo.putHeader(HttpHeaders.CONTENT_LENGTH, Integer.toString(buffer.length()));
```
marcjohnson-kint, line 27: Whilst these lines are changing, please could the header constants be statically imported.

src/main/java/org/folio/inventory/domain/view/request/StoredRequestView.java
bohdan-suprun: It is a circulation request.

src/main/java/org/folio/inventory/resources/AbstractInventoryResource.java
bohdan-suprun: It is an attempt to generalize inventory resources. Most of them don't have a base class.

src/main/java/org/folio/inventory/services/ItemService.java
bohdan-suprun: I can rename it to `MoveItemToStatusService`, in this case it can be reused for similar updates (for example for missing items).

What do you think?

descriptors/ModuleDescriptor-template.json
bohdan-suprun: I just want to have a general name convention. In this case an endpoint to mark an item missing will be `/inventory/items/{id}/miss|lose`?

marcjohnson-kint: > I just want to have a general name convention. In this case an endpoint to mark an item missing will be `/inventory/items/{id}/miss|lose`?

Interesting. I think that might be challenging. Could you expand a little on what you are thinking of as a convention, maybe with examples of other endpoints?

I think the example of an item being lost vs. withdrawn is especially interesting.

I think that withdrawing an item, is something that a library staff member actively decides to do. Whereas I think an item being lost is an event that has already occurred, and the staff member is changing the system to catch up with reality.

I think this is a domain modelling challenge as much as it is an API design challenge. My thinking was primarily on trying to succinctly express the domain intent.

src/main/java/org/folio/inventory/domain/view/request/StoredRequestView.java
marcjohnson-kint: Apologies for asking a vague question. 

What do `stored` and `view` mean in this context? 

Does inventory store it's own representation of the request?

src/main/java/org/folio/inventory/resources/AbstractInventoryResource.java
marcjohnson-kint: Ok, let's leave this in place. I'll close this as resolved.

(I'm aware that I am actively biased against inheritance)

src/main/java/org/folio/inventory/services/ItemService.java
marcjohnson-kint: Are we intending to try to reuse this for other actions?

Personally, I prefer to defer trying to generalise something (especially to reduce duplication) till the system has multiple examples to guide the generalisation.

src/main/java/org/folio/inventory/services/ItemService.java
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  public CompletableFuture<Item> processMarkItemWithdrawn(WebContext context) {
+    final String itemId = context.getStringParameter("id", null);
+
+    return itemCollection.findById(itemId)
+      .thenCompose(ItemMarkAsWithdrawnValidators::itemIsFound)
+      .thenCompose(ItemMarkAsWithdrawnValidators::itemHasAllowedStatusToMarkAsWithdrawn)
+      .thenCompose(this::updateRequestStatusIfRequired)
+      .thenApply(item -> item.changeStatus(ItemStatusName.WITHDRAWN))
```
marcjohnson-kint, line 46: Can this be statically imported to reduce the noise on this line?

src/main/java/org/folio/inventory/services/ItemService.java
```diff
@@ -0,0 +1,88 @@
...
+      .thenCompose(itemCollection::update);
+  }
+
+  private CompletableFuture<Item> updateRequestStatusIfRequired(Item item) {
+    final CqlQuery query = exactMatch("itemId", item.id)
+      .and(match("status", "Open"))
+      .and(notEqual("status", RequestStatus.OPEN_NOT_YET_FILLED.getValue()));
```
marcjohnson-kint, line 53: Can this be statically imported to reduce the noise on this line?

src/test/java/api/items/ItemMarkWithdrawnApiTest.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  @Parameters({
+    "On order",
+    "Checked out",
+    "Withdrawn",
+    "Claimed returned"
+  })
+  @Test
+  public void cannotMakeItemWithdrawnIfStatusIsNotAllowed(String initialStatus) throws Exception {
```
marcjohnson-kint, line 70:
```suggestion
  public void cannotWithdrawIItemWhenNotInAllowedStatus(String initialStatus) throws Exception {
```

src/test/java/api/items/ItemMarkWithdrawnApiTest.java
```diff
@@ -0,0 +1,218 @@
...
+  @Parameters({
+    "Available",
+    "In transit",
+    "Awaiting pickup",
+    "Awaiting delivery",
+    "Missing",
+    "Paged"
+  })
+  @Test
+  public void canMakeItemWithdrawn(String initialStatus) throws Exception {
```
marcjohnson-kint, line 43:
```suggestion
  public void canWithdrawItemWhenInAllowedStatus(String initialStatus) throws Exception {
```

src/test/java/api/items/ItemMarkWithdrawnApiTest.java
```diff
@@ -0,0 +1,218 @@
...
+      .canCirculate());
+
+    final Response response = markItemWithdrawn(createdItem.getId());
+
+    assertThat(response, hasValidationError("Item is not allowed to be marked as Withdrawn",
+      "status.name", initialStatus));
+  }
+
+  @Test
+  public void shouldReturnNotFoundWhenSpecifiedWrongItemId() throws Exception {
```
marcjohnson-kint, line 89:
```suggestion
  public void shouldWithdrawItemThatCannotBeFound() throws Exception {
```

src/test/java/api/items/ItemMarkWithdrawnApiTest.java
```diff
@@ -0,0 +1,218 @@
...
+      is(404));
+  }
+
+  @Test
+  @Parameters({
+    "Open - Awaiting delivery",
+    "Open - Awaiting pickup",
+    "Open - In transit"
+  })
+  public void canChangeRequestInFulfillmentToOpen(String requestStatus) throws Exception {
```
marcjohnson-kint, line 100:
```suggestion
  public void shouldChangeRequestBeingFulfillmentBackToNotYetFulfilled(String requestStatus) throws Exception {
```

src/test/java/api/items/ItemMarkWithdrawnApiTest.java
```diff
@@ -0,0 +1,218 @@
...
+      .getString("status"), is("Open - Not yet filled"));
+  }
+
+  @Test
+  @Parameters({
+    "Open - Awaiting delivery",
+    "Open - Awaiting pickup",
+    "Open - In transit"
+  })
+  public void shouldNotChangeExpiredRequestInFulfillmentToOpen(String requestStatus) throws Exception {
```
marcjohnson-kint, line 132:
```suggestion
  public void shouldNotChangeExpiredRequestInFulfillmentBackToNotYetFilled(String requestStatus) throws Exception {
```

src/test/java/api/items/ItemMarkWithdrawnApiTest.java
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  @Test
+  @Parameters({
+    "Closed - Cancelled",
+    "Closed - Filled",
+    "Closed - Pickup expired",
+    "Closed - Unfilled"
+  })
+  public void shouldNotChangeClosedRequestToOpen(String requestStatus) throws Exception {
```
marcjohnson-kint, line 165:
```suggestion
  public void shouldNotReopenClosedRequest(String requestStatus) throws Exception {
```

descriptors/ModuleDescriptor-template.json
bohdan-suprun: I think endpoints can be named like `/mark-{target-status}`, e.g. `/mark-missing`, `/mark-withdrawn`, etc.

Please note, that UI changes has been already merged, it will require a follow up PR for UI.

src/main/java/org/folio/inventory/domain/view/request/StoredRequestView.java
bohdan-suprun: The stored means the request from storage but not from `mod-circulation`. And by view I mean that the class contains all the properties as raw key-values and some delegating methods to update/read these properties.

Do you think it should be renamed?

src/main/java/org/folio/inventory/services/ItemService.java
bohdan-suprun: I'm fine to make it specific for withdraw case. Will rename it to `WithdrawItemService`.

src/main/java/org/folio/inventory/support/CompletableFutures.java
bohdan-suprun: I've added a comment to the method. I think Sonar will report an error if I explicitly add `@Deprecated` annotation.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Yeah, it's too late to change this API design at this point.

I think we need to be careful not to try too hard to fit these into the same naming pattern, I think that might be a missed opportunity for modelling these processes rather than only as status changes.

Anyway, I'm resolving this because I gave this feedback too late :-(

src/main/java/org/folio/inventory/domain/view/request/StoredRequestView.java
marcjohnson-kint: > The stored means the request from storage but not from `mod-circulation`. And by view I mean that the class contains all the properties as raw key-values and some delegating methods to update/read these properties.
> 
> Do you think it should be renamed?

Thank you. Those aspects don't come across naturally for me from the name alone. Personally, I think it could be called a `Request` without losing much of the descriptive power.

I think the view aspect is (hopefully) an internal modelling, if it doesn't affect the behaviour or the client API then I think it can be kept private and not part of the name.

The storage aspect is interesting, I think that could be expressed in the class that is responsible for fetching it from storage and updating it in storage. Unless, there are specific aspects of the storage representation that this class leaks out to the rest of the module?

Does that make sense?

bohdan-suprun: Class has been renamed to `Request` and `RequestRepository` has been created which uses this class.

src/main/java/org/folio/inventory/storage/external/repository/RequestRepository.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import org.folio.inventory.domain.view.request.Request;
+import org.folio.inventory.storage.external.Clients;
+import org.folio.inventory.storage.external.CollectionResourceRepository;
+import org.folio.inventory.storage.external.CqlQuery;
+
+public class RequestRepository {
+  private static final List<String> REQUEST_IN_FULFILLMENT_STATUSES =
+    getRequestInFulfillmentStatuses();
```
marcjohnson-kint, line 25: Why is this needed as a static constant, rather than being determined when it is needed?

src/test/java/api/items/ItemMarkWithdrawnApiTest.java
```diff
@@ -0,0 +1,218 @@
...
+      is(404));
+  }
+
+  @Test
+  @Parameters({
+    "Open - Awaiting delivery",
+    "Open - Awaiting pickup",
+    "Open - In transit"
+  })
+  public void shouldChangeRequestBeingFulfillmentBackToNotYetFilled(String requestStatus) throws Exception {
```
marcjohnson-kint, line 100:
```suggestion
  public void shouldChangeRequestBeingFulfilledBackToNotYetFilled(String requestStatus) throws Exception {
```

src/main/java/org/folio/inventory/storage/external/repository/RequestRepository.java
bohdan-suprun: I've removed the constant.

PR #189, opened 2020-04-01
MODDICORE-38: DataImportEventPayload is not processing correctly

OleksiiKuzminov:


OleksiiKuzminov:
urgent bugfix. lib version will be changed to release after bugfix release fo data-import-processing-core

marcjohnson-kint:
I understand this was an urgent change that was required in a hurry. 

Please can all pull requests for a module reference JIRA issue for that project, as that is needed to do version management

PR #199, opened 2020-04-13
MODINV-218: Require title for unconnected preceding/succeeding titles

bohdan-suprun:
https://issues.folio.org/browse/MODINV-218 - Title should be required for unconnected preceding / succeeding titles.

### Changes:

* Introduce validation to prevent an unconnected title will be created/updated without title.
* Update InstanceBatch as well.
* Refactor Instance.update/create endpoints to use completable futures.

### DIFF
--- a/src/main/java/org/folio/inventory/support/EndpointFailureHandler.java
+++ b/src/main/java/org/folio/inventory/support/EndpointFailureHandler.java
@@ -2,6 +2,8 @@ package org.folio.inventory.support;
 
 import static org.folio.inventory.support.http.server.JsonResponse.unprocessableEntity;
 
+import java.util.function.Function;
+
 import org.folio.inventory.exceptions.AbstractInventoryException;
 import org.folio.inventory.storage.external.exceptions.ExternalResourceFetchException;
 import org.folio.inventory.support.http.server.ClientErrorResponse;
@@ -23,7 +25,9 @@ public final class EndpointFailureHandler {
     handleFailure(result.cause(), context);
   }
 
-  public static void handleFailure(Throwable failure, RoutingContext context) {
+  public static void handleFailure(Throwable originFailure, RoutingContext context) {
+    final Throwable failure = getKnownException(originFailure);
+
     if (failure instanceof UnprocessableEntityException) {
       UnprocessableEntityException validationFailure =
         (UnprocessableEntityException) failure;
@@ -41,7 +45,20 @@ public final class EndpointFailureHandler {
     }
   }
 
+  public static <T> Function<Throwable, T> doExceptionally(
+    RoutingContext context) {
+
+    return failure -> {
+      handleFailure(failure, context);
+      return null;
+    };
+  }
+
   public static Throwable getKnownException(Throwable throwable) {
+    if (throwable instanceof AbstractInventoryException) {
+      return throwable;
+    }
+
     return throwable.getCause() instanceof AbstractInventoryException
       ? throwable.getCause() : throwable;
   }
--- a/src/main/java/org/folio/inventory/validation/InstancesValidators.java
+++ b/src/main/java/org/folio/inventory/validation/InstancesValidators.java
@@ -1,0 +1,71 @@
+package org.folio.inventory.validation;
+
+import static java.util.concurrent.CompletableFuture.completedFuture;
+import static org.folio.inventory.support.CompletableFutures.failedFuture;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+import org.apache.commons.lang3.StringUtils;
+import org.folio.inventory.domain.instances.Instance;
+import org.folio.inventory.domain.instances.titles.PrecedingSucceedingTitle;
+import org.folio.inventory.support.http.server.ValidationError;
+import org.folio.inventory.validation.exceptions.NotFoundException;
+import org.folio.inventory.validation.exceptions.UnprocessableEntityException;
+
+public final class InstancesValidators {
+
+  private InstancesValidators() {}
+
+  public static CompletableFuture<Instance> refuseWhenNoTitleForPrecedingSucceedingUnconnectedTitle(
+    Instance instance) {
+
+    final ValidationError succeedingError = new ValidationError(
+      "Title is required for unconnected succeeding title", "succeedingTitles.title", null);
+    final ValidationError precedingError = new ValidationError(
+      "Title is required for unconnected preceding title", "precedingTitles.title", null);
+
+    return refuseWhenNoTitleForUnconnectedTitle(instance, instance.getSucceedingTitles(), succeedingError)
+      .thenCompose(prev -> refuseWhenNoTitleForUnconnectedTitle(instance, instance.getPrecedingTitles(), precedingError));
+
+  }
+
+  private static CompletableFuture<Instance> refuseWhenNoTitleForUnconnectedTitle(
+    Instance instance, List<PrecedingSucceedingTitle> titles, ValidationError error) {
+
+    return isTitleMissingForUnconnectedPrecedingSucceeding(titles)
+      ? failedFuture(new UnprocessableEntityException(error))
+      : completedFuture(instance);
+  }
+
+  public static boolean isTitleMissingForUnconnectedPrecedingSucceeding(List<PrecedingSucceedingTitle> titles) {
+    if (titles == null || titles.isEmpty()) {
+      return false;
+    }
+
+    return titles.stream()
+      .filter(InstancesValidators::isUnconnectedPrecedingSucceedingTitle)
+      .anyMatch(title -> StringUtils.isBlank(title.title));
+  }
+
+  private static boolean isUnconnectedPrecedingSucceedingTitle(PrecedingSucceedingTitle title) {
+    return title.precedingInstanceId == null && title.succeedingInstanceId == null;
+  }
+
+  public static CompletionStage<Instance> refuseWhenInstanceNotFound(Instance instance) {
+    return instance == null
+      ? failedFuture(new NotFoundException("Instance not found"))
+      : completedFuture(instance);
+  }
+
+  public static CompletionStage<Instance> refuseWhenHridChanged(
+    Instance existingInstance, Instance updatedInstance) {
+
+    return Objects.equals(existingInstance.getHrid(), updatedInstance.getHrid())
+      ? completedFuture(existingInstance)
+      : failedFuture(new UnprocessableEntityException("HRID can not be updated",
+      "hrid", updatedInstance.getHrid()));
+  }
+}
--- a/src/test/java/api/InstancesApiExamples.java
+++ b/src/test/java/api/InstancesApiExamples.java
@@ -63,10 +63,6 @@ public class InstancesApiExamples extends ApiTests {
   private final String tagNameOne = "important";
   private final String tagNameTwo = "very important";
 
-  public InstancesApiExamples() throws MalformedURLException {
-    super();
-  }
-
   @Test
   public void canCreateInstanceWithoutAnIDAndHRID()
     throws InterruptedException,
@@ -557,7 +553,7 @@ public class InstancesApiExamples extends ApiTests {
     assertNotNull(putResponse.getJson().getJsonArray("errors"));
     JsonArray errors = putResponse.getJson().getJsonArray("errors");
     assertThat(errors.size(), is(1));
-    assertThat(errors.getString(0), is(
+    assertThat(errors.getJsonObject(0).getString("message"), is(
       "Instance is controlled by MARC record, these fields are blocked and can not be updated: " +
         "physicalDescriptions,notes,languages,identifiers,instanceTypeId,modeOfIssuanceId,subjects," +
         "source,title,indexTitle,publicationFrequency,electronicAccess,publicationRange," +
--- a/src/test/java/api/InstancesApiExamples.java
+++ b/src/test/java/api/InstancesApiExamples.java
@@ -136,12 +136,12 @@ public class InstancesApiExamples extends ApiTests {
     final JsonObject tags = createdInstance.getJsonObject(TAGS_KEY);
     assertTrue(tags.containsKey(TAG_LIST_KEY));
     final JsonArray tagList = tags.getJsonArray(TAG_LIST_KEY);
-    assertThat((ArrayList<String>)tagList.getList(), hasItem(tagNameOne));
+    assertThat(tagList, hasItem(tagNameOne));
 
     JsonArray natureOfContentTermIds = createdInstance.getJsonArray("natureOfContentTermIds");
     assertThat(natureOfContentTermIds.size(), is(2));
-    assertThat((ArrayList<String>)natureOfContentTermIds.getList(), hasItem(ApiTestSuite.getAudiobookNatureOfContentTermId()));
-    assertThat((ArrayList<String>)natureOfContentTermIds.getList(), hasItem(ApiTestSuite.getBibliographyNatureOfContentTermId()));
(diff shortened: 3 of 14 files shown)


src/main/java/org/folio/inventory/support/EndpointFailureHandler.java
```diff
@@ -41,7 +45,20 @@ public static void handleFailure(Throwable failure, RoutingContext context) {
...
+      return null;
+    };
+  }
+
   public static Throwable getKnownException(Throwable throwable) {
+    if (throwable instanceof AbstractInventoryException) {
+      return throwable;
+    }
+
     return throwable.getCause() instanceof AbstractInventoryException
```
julianladisch, line 62: Can you use this code that is more readable? This avoid a merge conflict.
https://github.com/folio-org/mod-inventory/pull/185/files#diff-6e5db33f5ac7695ffb682193802ab08cR53-R57

bohdan-suprun: Sure, changed.

src/main/java/org/folio/inventory/support/EndpointFailureHandler.java
```diff
@@ -45,6 +47,15 @@ public static void handleFailure(Throwable failure, RoutingContext context) {
     }
   }
 
+  public static <T> Function<Throwable, T> doExceptionally(
+    RoutingContext context) {
```
marcjohnson-kint, line 51: Personally, I'd include the parameter on the same line, even though it is slightly over 80 characters (in practice, I typically accept between 80 and 100 characters, especially as we are likely to move towards standard that uses 100)

```suggestion
  public static <T> Function<Throwable, T> doExceptionally(RoutingContext context) {
```

src/main/java/org/folio/inventory/validation/InstancesValidators.java
```diff
@@ -0,0 +1,71 @@
...
+    return title.precedingInstanceId == null && title.succeedingInstanceId == null;
+  }
+
+  public static CompletionStage<Instance> refuseWhenInstanceNotFound(Instance instance) {
+    return instance == null
+      ? failedFuture(new NotFoundException("Instance not found"))
+      : completedFuture(instance);
+  }
+
+  public static CompletionStage<Instance> refuseWhenHridChanged(
```
marcjohnson-kint, line 63: Why does this return a `CompletionStage` rather than a `CompletableFuture`?

src/main/java/org/folio/inventory/validation/InstancesValidators.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.CompletionStage;
+
+import org.apache.commons.lang3.StringUtils;
+import org.folio.inventory.domain.instances.Instance;
+import org.folio.inventory.domain.instances.titles.PrecedingSucceedingTitle;
+import org.folio.inventory.support.http.server.ValidationError;
+import org.folio.inventory.exceptions.NotFoundException;
+import org.folio.inventory.exceptions.UnprocessableEntityException;
+
+public final class InstancesValidators {
```
marcjohnson-kint, line 18: Could this class be specific to (preceding or succeeding) unconnected titles, rather than instances in general?

That might mean some of these method names could be significantly simplified.

The methods could still take an instance as a parameter

src/test/java/api/InstancesApiExamples.java
```diff
@@ -54,17 +58,20 @@
...
   private static final InventoryConfiguration config = new InventoryConfigurationImpl();
   private final String tagNameOne = "important";
   private final String tagNameTwo = "very important";
 
-  public InstancesApiExamples() throws MalformedURLException {
-    super();
+  @After
+  public void expireFailureEmulation() throws Exception {
+    instancesStorageClient.expireFailureEmulation();
   }
```
marcjohnson-kint, line 75: What does expire failure emulation mean?

src/test/java/api/InstancesApiExamples.java
```diff
@@ -557,7 +563,7 @@ public void canNotUpdateAnExistingMARCInstanceIfBlockedFieldsAreChangedToNulls()
     assertNotNull(putResponse.getJson().getJsonArray("errors"));
     JsonArray errors = putResponse.getJson().getJsonArray("errors");
     assertThat(errors.size(), is(1));
-    assertThat(errors.getString(0), is(
+    assertThat(errors.getJsonObject(0).getString("message"), is(
```
marcjohnson-kint, line 566: This change suggests that the structure of the response has changed, from a string only to an JSON object. Is that the case?

src/test/java/api/InstancesApiExamples.java
```diff
@@ -873,22 +879,59 @@ public void cannotRemoveHRID() throws Exception {
...
+  @Test
+  public void canFrowardInstanceCreateFailureFromStorage() throws Exception {
+    final String expectedErrorMessage = "Instance-storage is temporary unavailable for create";
+
+    instancesStorageClient.emulateFailure(new EndpointFailureDescriptor()
+      .setFailureExpireDate(DateTime.now(DateTimeZone.UTC).plusSeconds(2).toDate())
+      .setBody(expectedErrorMessage)
+      .setContentType("plain/text")
+      .setStatusCode(500)
+      .setMethod(HttpMethod.POST));
```
marcjohnson-kint, line 891: Personally, I'd statically import the method, as I think it makes this line more readable

```suggestion
      .setMethod(POST));
```

src/test/java/api/InstancesApiExamples.java
```diff
@@ -873,22 +879,59 @@ public void cannotRemoveHRID() throws Exception {
     assertThat(existingInstance, is(createdInstance));
   }
 
+  @Test
+  public void canFrowardInstanceCreateFailureFromStorage() throws Exception {
+    final String expectedErrorMessage = "Instance-storage is temporary unavailable for create";
+
+    instancesStorageClient.emulateFailure(new EndpointFailureDescriptor()
+      .setFailureExpireDate(DateTime.now(DateTimeZone.UTC).plusSeconds(2).toDate())
```
marcjohnson-kint, line 887: Personally, I'd statically import the timezone, as I think it makes this line more readable

```suggestion
      .setFailureExpireDate(DateTime.now(UTC).plusSeconds(2).toDate())
```

src/test/java/api/InstancesApiExamples.java
```diff
@@ -873,22 +879,59 @@ public void cannotRemoveHRID() throws Exception {
...
+
+    final Response response = instancesClient.attemptToCreate(smallAngryPlanet(UUID.randomUUID()));
+
+    assertThat(response.getStatusCode(), is(500));
+    assertThat(response.getBody(), is(expectedErrorMessage));
+  }
+
+  @Test
+  public void canFrowardInstanceUpdateFailureFromStorage() throws Exception {
+    final String expectedErrorMessage = "Instance-storage is temporary unavailable for updates";
```
marcjohnson-kint, line 901: How does the system know that the fault is temporary?

src/test/java/api/PrecedingSucceedingTitlesApiExamples.java
```diff
@@ -132,6 +141,135 @@ public void canUpdateAnInstanceWithUnconnectedSucceedingTitles()
       null);
   }
 
+  @Test
+  public void titleIsRequiredToCreateAnInstanceWithUnconnectedSucceedingTitles()
```
marcjohnson-kint, line 145:
```suggestion
  public void titleIsRequiredToCreateAnInstanceWithUnconnectedSucceedingTitle()
```

src/test/java/api/PrecedingSucceedingTitlesApiExamples.java
```diff
@@ -132,6 +141,135 @@ public void canUpdateAnInstanceWithUnconnectedSucceedingTitles()
...
+        .add(succeedingTitle));
+
+    final Response response = instancesClient.attemptToCreate(smallAngryPlanetJson);
+
+    assertThat(response, hasValidationError("Title is required for unconnected succeeding title",
+      SUCCEEDING_TITLES_KEY + ".title", null));
+  }
+
+  @Test
+  public void titleIsRequiredToUpdateAnInstanceWithUnconnectedSucceedingTitles()
```
marcjohnson-kint, line 163:
```suggestion
  public void titleIsRequiredToUpdateAnInstanceWithUnconnectedSucceedingTitle()
```

src/test/java/api/PrecedingSucceedingTitlesApiExamples.java
```diff
@@ -132,6 +141,135 @@ public void canUpdateAnInstanceWithUnconnectedSucceedingTitles()
...
+  public void titleIsRequiredToCreateAnInstanceWithUnconnectedSucceedingTitles()
+    throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
+
+    final JsonObject succeedingTitle = createSemanticWebUnconnectedTitle(UUID.randomUUID().toString());
+    succeedingTitle.remove("title");
+
+    final JsonObject smallAngryPlanetJson = smallAngryPlanet(UUID.randomUUID())
+      .put(SUCCEEDING_TITLES_KEY, new JsonArray()
+        .add(createSemanticWebUnconnectedTitle(UUID.randomUUID().toString()))
+        .add(succeedingTitle));
```
marcjohnson-kint, line 154: Why are two unconnected titles needed to demonstrate this behaviour?

src/main/java/org/folio/inventory/validation/InstancesValidators.java
bohdan-suprun: This class contains some method that are not specific for preceding/succeeding titles: `refuseWhenInstanceNotFound`, `refuseWhenHridChanged`.

I can extract methods that are used for titles validation into a separate class, does it make sence?

src/test/java/api/InstancesApiExamples.java
bohdan-suprun: Yes, but here used a not standardized validation error response.

src/test/java/api/InstancesApiExamples.java
bohdan-suprun: I'm configuring instance-storage fake API to emulate failures:
```
instancesStorageClient.emulateFailure(new EndpointFailureDescriptor()
      .setFailureExpireDate(DateTime.now(DateTimeZone.UTC).plusSeconds(2).toDate())
      .setBody(expectedErrorMessage)
      .setContentType("plain/text")
      .setStatusCode(500)
      .setMethod(HttpMethod.PUT));
```

This means that each PUT request to /inventory-storage/instances will be rejected with status 500 and with plain/text error saying `Instance-storage is temporary unavailable for updates` till `now() + 2s`.

src/test/java/api/PrecedingSucceedingTitlesApiExamples.java
bohdan-suprun: One of them is a valid unconnected relationship but another one is not valid.

src/test/java/api/InstancesApiExamples.java
bohdan-suprun: This means to disable failure emulation for instance-storage fake endpoint by setting `failureExpireDate` (when the endpoint should stop emulating failures) to a datetime in the past.

src/main/java/org/folio/inventory/validation/InstancesValidators.java
bohdan-suprun: Changed to `CompletableFuture`.

src/main/java/org/folio/inventory/validation/InstancesValidators.java
marcjohnson-kint: I think it makes sense for those to be in separate classes

src/test/java/api/InstancesApiExamples.java
marcjohnson-kint: Could it be renamed to `disableFailureEmulation` instead, given that is the intent?

```suggestion
  public void disableFailureEmulation() throws Exception {
    instancesStorageClient.expireFailureEmulation();
  }
```

src/test/java/api/InstancesApiExamples.java
marcjohnson-kint: Is this defined in an interface that is version 1.0 or greater?

src/test/java/api/PrecedingSucceedingTitlesApiExamples.java
marcjohnson-kint: Why does this test need a valid relationship as well as an invalid one to demonstrate the behaviour?

src/test/java/api/InstancesApiExamples.java
bohdan-suprun: The error message is returned by `/inventory/instances` endpoint.

src/test/java/api/PrecedingSucceedingTitlesApiExamples.java
bohdan-suprun: To verify that validation checks each item in array, not only the first one.

src/main/java/org/folio/inventory/validation/InstancesValidators.java
bohdan-suprun: Extracted to `InstancePrecedingSucceedingTitleValidators` class.

src/test/java/api/InstancesApiExamples.java
marcjohnson-kint: Does the RAML include a definition for the new structure?

src/test/java/api/InstancesApiExamples.java
marcjohnson-kint: Ok, thanks. This test is checking that the error from storage comes all the way through to the client, is that the case?

src/test/java/api/PrecedingSucceedingTitlesApiExamples.java
marcjohnson-kint: Ok, please could a comment be added to this test to explain that reasoning

src/main/java/org/folio/inventory/validation/InstancePrecedingSucceedingTitleValidators.java
```diff
@@ -0,0 +1,53 @@
...
+
+    final ValidationError succeedingError = new ValidationError(
+      "Title is required for unconnected succeeding title", "succeedingTitles.title", null);
+    final ValidationError precedingError = new ValidationError(
+      "Title is required for unconnected preceding title", "precedingTitles.title", null);
+
+    return refuseWhenUnconnectedHasNoTitle(instance, instance.getSucceedingTitles(), succeedingError)
+      .thenCompose(prev -> refuseWhenUnconnectedHasNoTitle(instance, instance.getPrecedingTitles(), precedingError));
+
+  }
```
marcjohnson-kint, line 30: Please could the line lengths be consistent in this method. If the method parameter is being split onto a separate line due to length, that would suggest that at least the second line of the return statement is too long.

src/test/java/api/InstancesApiExamples.java
bohdan-suprun: No, unfortunately. The RAML does not even have the 422 response code definition.

src/test/java/api/InstancesApiExamples.java
bohdan-suprun: Yes, correct.

src/test/java/api/InstancesApiExamples.java
marcjohnson-kint: Then I think we should either introduce the 422 response code and switch to using that, or leave the existing structure as is.

Which would you rather do?

(AFAIK FOLIO does not have clear guidance on whether introducing a new kind of response is compatibility breaking)

bohdan-suprun: I've introduced the 422 response and incremented inventory API version.

PR #205, opened 2020-04-30
MODINV-283 - Inventory search hangs

RuslanLavrov:
JIRA-ticket: https://issues.folio.org/browse/MODINV-283

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
@@ -1,6 +1,6 @@
 package org.folio.inventory.dataimport.handlers.matching.loaders;
 
-import io.vertx.core.WorkerExecutor;
+import io.vertx.core.Vertx;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
@@ -24,10 +24,10 @@ public abstract class AbstractLoader<T> implements MatchValueLoader {
 
   private static final Logger LOG = LoggerFactory.getLogger(AbstractLoader.class);
 
-  protected WorkerExecutor workerExecutor;
+  private Vertx vertx;
 
-  public AbstractLoader(WorkerExecutor workerExecutor) {
-    this.workerExecutor = workerExecutor;
+  public AbstractLoader(Vertx vertx) {
+    this.vertx = vertx;
   }
 
   @Override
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
```diff
@@ -40,7 +40,7 @@ public LoadResult loadEntity(LoadQuery loadQuery, DataImportEventPayload eventPa
     loadResult.setEntityType(getEntityType().value());
     Context context = constructContext(eventPayload.getTenant(), eventPayload.getToken(), eventPayload.getOkapiUrl());
 
-    workerExecutor.executeBlocking(blockingFuture -> {
+    vertx.runOnContext(blockingFuture -> {
```
bohdan-suprun, line 43: Hi @RuslanLavrov, what is the difference between previous approach and the current one?

As far as I understand from the logs provided, the issue was caused because vert.x event bus thread was blocked by `future.join()` call.

RuslanLavrov: Hi @bohdan-suprun. The difference is to make the blocking method call `future.join()` from a thread of additional worker pool instead of threads, which serve inventory verticle.

marcjohnson-kint: @RuslanLavrov 

> The difference is to make the blocking method call `future.join()` from a thread of additional worker pool instead of threads, which serve inventory verticle.

Does that mean that all of the threads in the dedicated pool for event processing could block instead of the shared pool?

Why does this operation need to be blocked at all? What is the join needed for?

RuslanLavrov: yes, threads from the additional pool will be used for running blocking method, it is needed
to return loadResult

marcjohnson-kint: @RuslanLavrov 

> it is needed to return loadResult

Apologies, I really don't follow. Why is a blocking operation required to return a result? 

Why could the process not be invoked asynchronously? It is still possible to return a result at the end.

marcjohnson-kint: Could this method return a CompletableFuture<LoadResult> instead of performing the join within it, then the client could decide how to continue?

This would avoid the need for a blocking call.

PR #207, opened 2020-05-04
MODINV-288 Added 'bareHoldingsItems' field to holdings-record schema

KaterynaSenchenko:


ramls/holdings-record.json
```diff
@@ -259,6 +259,20 @@
...
+        "type": "object",
+        "$ref": "item.json"
+      },
+      "readonly": true,
+      "folio:isVirtual": true,
+      "folio:linkBase": "item-storage/items",
+      "folio:linkFromField": "id",
+      "folio:linkToField": "holdingsRecordId",
+      "folio:includedElement": "items"
+    },
```
marcjohnson-kint, line 275: What is this property meant to be used for? At the moment, virtual (fake) properties are only used by mod-graphql, is this being used in production?

KaterynaSenchenko: HoldingsRecord entity created in mod-inventory-storage contains this field, and it cannot be mapped to the entity in mod-inventory unless we define this field in the schema. I don't know where it is being used, this is a quick fix to resolve the mapping error that occurs due to schema inconsistencies between mod-inventory and mod-inventory-storage

marcjohnson-kint: > HoldingsRecord entity created in mod-inventory-storage contains this field

No holdings records should contain this property, this is a fake property. It is unfortunate that these fake properties become real when RAML Module Builder interprets them, which causes confusion and bugs.

> and it cannot be mapped to the entity in mod-inventory unless we define this field in the schema. 

This property should not be mapped in inventory.

> I don't know where it is being used, this is a quick fix to resolve the mapping error that occurs due to schema inconsistencies between mod-inventory and mod-inventory-storage

Inconsistencies in what respect? Is this because mod-inventory now creates holdings records during the new data import process?

KaterynaSenchenko: > Inconsistencies in what respect? Is this because mod-inventory now creates holdings records during the new data import process?
Yes, new Holdings are created in mod-inventory and the json response from mod-inventory-storage contains this field. In order to map this json to HoldingsRecord we need to define this field

KaterynaSenchenko: @marc-johnson, should I create a bug in mod-inventory-storage instead?

marcjohnson-kint: >  should I create a bug in mod-inventory-storage instead?

What would the bug in mod-inventory-storage be?

If it is that these virtual properties can be assigned values, then you are more than welcome to raise it, it is likely to be complicated to resolve.

KaterynaSenchenko:
Merging this quick fix as agreed in the comments of https://issues.folio.org/browse/MODINV-288

PR #208, opened 2020-05-05
MODINV-284: Archive event payload (#193)

VRohach:
1. Unzipping added.
2. 'org.apache.commons' changed on 'commons-io' with the actual version.
3. Tests changed.


VRohach:
This PR can be merged only after merging https://github.com/folio-org/data-import-processing-core/pull/45.

PR #209, opened 2020-05-15
MODINV-277:Disallow editing of Preceding/Succeeding fields when Instance has an underlying MARC record

VRohach:
1. Preceding/Succeding titles added to the INSTANCE_BLOCKED_FIELDS.
2. Preceding/Succeding titles added to the InstancesApiExamples.

### DIFF
--- a/src/main/java/org/folio/inventory/config/InventoryConfigurationImpl.java
+++ b/src/main/java/org/folio/inventory/config/InventoryConfigurationImpl.java
@@ -30,7 +30,9 @@ public class InventoryConfigurationImpl implements InventoryConfiguration {
     Instance.TITLE_KEY,
     Instance.INDEX_TITLE_KEY,
     Instance.INSTANCE_TYPE_ID_KEY,
-    Instance.MODE_OF_ISSUANCE_ID_KEY
+    Instance.MODE_OF_ISSUANCE_ID_KEY,
+    Instance.PRECEDING_TITLES_KEY,
+    Instance.SUCCEEDING_TITLES_KEY
     );
 
   public InventoryConfigurationImpl() {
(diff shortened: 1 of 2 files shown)


marcjohnson-kint:
Does this stop preceding and succeeding titles being changed in the API, or does it only inform the reference UI of the preference? 

If it does, please could a test be added to demonstrate this behaviour.

VRohach:
> Does this stop preceding and succeeding titles being changed in the API, or does it only inform the reference UI of the preference?
> 
> If it does, please could a test be added to demonstrate this behaviour.

Hello @marcjohnson-kint ! I think that this behaviour is already demonstrated int this test-case: "canNotUpdateAnExistingMARCInstanceIfBlockedFieldsAreChanged"

marcjohnson-kint:
> > Does this stop preceding and succeeding titles being changed in the API, or does it only inform the reference UI of the preference?
> > If it does, please could a test be added to demonstrate this behaviour.
> 
> Hello @marcjohnson-kint ! I think that this behaviour is already demonstrated in this test-case: "canNotUpdateAnExistingMARCInstanceIfBlockedFieldsAreChanged"

Does that work as well for special properties which are treated differently, like preceding and succeeding titles?

VRohach:
> > > Does this stop preceding and succeeding titles being changed in the API, or does it only inform the reference UI of the preference?
> > > If it does, please could a test be added to demonstrate this behaviour.
> > 
> > 
> > Hello @marcjohnson-kint ! I think that this behaviour is already demonstrated in this test-case: "canNotUpdateAnExistingMARCInstanceIfBlockedFieldsAreChanged"
> 
> Does that work as well for special properties which are treated differently, like preceding and succeeding titles?

I think you are right. @OleksiiKuzminov could you please correct me, if I`m wrong?

OleksiiKuzminov:
> Does this stop preceding and succeeding titles being changed in the API, or does it only inform the reference UI of the preference?
> 
> If it does, please could a test be added to demonstrate this behaviour.

https://github.com/folio-org/mod-inventory/blob/2401aaf6fc908d2dafa5641038029c0b884d8bf5/src/main/java/org/folio/inventory/resources/Instances.java#L206

All properties from InventoryConfigurationImpl are protected in the update method

PR #211, opened 2020-05-18
MODINV-207 Implemented Update action handler for Inventory Instances

KaterynaSenchenko:


src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateInstanceEventHandler.java
```diff
@@ -0,0 +1,106 @@
...
+        LOGGER.error(message);
+        future.completeExceptionally(new EventProcessingException(message));
+        return future;
+      }
+      JsonObject mappingRules = new JsonObject(eventPayload.get(MAPPING_RULES_KEY));
+      MappingParameters mappingParameters = new JsonObject(eventPayload.get(MAPPING_PARAMS_KEY)).mapTo(MappingParameters.class);
+      Record marcRecord = new JsonObject(eventPayload.get(MARC_KEY)).mapTo(Record.class);
+
+      JsonObject parsedRecord = JsonObject.mapFrom(marcRecord.getParsedRecord().getContent());
+      String instanceId = marcRecord.getExternalIdsHolder().getInstanceId();
```
VRohach, line 51: @KaterynaSenchenko  I think that there will be exception if ExternalIdsHolder will be empty. I think it will be better to handle it.

KaterynaSenchenko: This Instance update is triggered by the update of the underlying SRS MARC Record - it must contain the id of the Instance it is linked to. In case there is no Instance id (which we do not expect), the exception will be caught below, logged and the future will be completed exceptionally

VRohach: @KaterynaSenchenko Thank you for the description!

PR #212, opened 2020-05-19
MODINV-291: Add 'Lost and paid' status to allowed item statuses list

bohdan-suprun:
https://issues.folio.org/browse/MODINV-291 - Add 'Lost and paid' status to allowed item statuses list.

### Changes:
* Add `Lost and paid` status to allowed item's statuses list;
* Update inventory API version;
* Fix some code analysis issues;

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
-      "version": "10.4",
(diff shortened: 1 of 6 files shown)


bohdan-suprun:
> Does the `inventory-batch` interface version need to change?
> 
> Does the `instance-storage-batch` dependency version need to change?

No, neither of these APIs need version update. Instance is not affected by the item update.

PR #214, opened 2020-05-21
MODINV-286: Behavior of Withdrawn items on Inventory item record 

jeremythuff:
Resolves [MODINV-286](https://issues.folio.org/browse/MODINV-286)

### DIFF
--- a/src/test/java/org/folio/inventory/domain/items/ItemTest.java
+++ b/src/test/java/org/folio/inventory/domain/items/ItemTest.java
@@ -1,15 +1,12 @@
 package org.folio.inventory.domain.items;
 
 import static org.hamcrest.Matchers.instanceOf;
-import static org.junit.Assert.assertEquals;
 
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 import io.vertx.core.json.JsonObject;
-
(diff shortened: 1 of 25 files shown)


src/test/java/org/folio/inventory/domain/items/ItemTest.java
```diff
@@ -21,4 +24,14 @@ public void cannotCreateItemIfStatusIsNull() {
     new Item("id", "holding-id", null, "material-type-id",
       "permanent-loan-type-id", new JsonObject());
   }
+
+  @Test
+  public void canMarkWithdrawnItemAsMissing() {
```
bohdan-suprun, line 29: Could it be API test instead? The scenario should be following:

`api.items.ItemApiExamples`:
1. Create an available item;
1. Mark it as withdrawn via calling API endpoint `/inventory/items/:itemId//mark-withdrawn` (see api.items.ItemMarkWithdrawnApiTest#markItemWithdrawn);
1. Mark item as missing by replacing the item (use `api.support.http.ResourceClient#replace(UUID, JsonObject)`);
1. Verify item status changed.

jeremythuff: Ok this should be done

PR #219, opened 2020-05-28
MODINV-243 Enable Action profile action for incoming MARC Bibs: Replace Inventory Instance

OleksiiKuzminov:


marcjohnson-kint:
This pull request appears to have been merged despite having failed the quality gate for code duplication. Could someone help me understand why that was done?

cc: @craigmcnally

OleksiiKuzminov:
This PR was merged cause no comments from reviewers and we need to have these changes on env to test before demo
Code duplications scan works not very good and doesn’t check strings in log methods

marcjohnson-kint:
> This PR was merged cause no comments from reviewers and we need to have these changes on env to test before demo

Ok, this pull request was expedited because of a 

> Code duplications scan works not very good and doesn’t check strings in log methods

I don't think we want to encourage folks from ignoring the automated checks that are in place. If the checks or limits are inappropriate, please could this topic be raised with the technical leads or another group.

cc: @skoczko @craigmcnally

PR #228, opened 2020-06-12
MODINV-308 - Create Move schema

siarhei-charniak:
https://issues.folio.org/browse/MODINV-308 - Create Move schema.

created schema and sample

BE design: https://wiki.folio.org/pages/viewpage.action?pageId=36576153


ramls/move.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Ids holder for moving items between holdings",
+  "type": "object",
+  "properties": {
+    "to": {
+      "type": "string",
```
khandramai, line 7: Does uuid.json already contain variable type? If yes we actually don't need it here.

ramls/move.json
```diff
@@ -0,0 +1,24 @@
...
+      "description": "Unique identifier of instance/holding to which holdings/items to be moved",
+      "$ref": "uuid.json"
+    },
+    "ids": {
+      "description": "Array of holding/item unique identifiers which will be moved to the new instance/holding",
+      "type": "array",
+      "items": {
+        "type": "string",
```
khandramai, line 15: The same https://github.com/folio-org/mod-inventory/pull/228/files#r439264154

ramls/move.json
siarhei-charniak: removed

khandramai:
@folio-org/inventory-backend-code-owners 

Hi guys, in order to implement [Move Holdings/Items](https://wiki.folio.org/pages/viewpage.action?pageId=36576153) feature I would like to ask you about code review of our PRs to inventory.

Thanks,
Slava

ramls/move.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Ids holder for moving items between holdings",
+  "type": "object",
+  "properties": {
+    "to": {
+      "description": "Unique identifier of instance/holding to which holdings/items to be moved",
```
julianladisch, line 7:
```suggestion
      "description": "Id of the instance to which the holdings are moved, or id of the holding to which the items are moved.,
```

ramls/move.json
```diff
@@ -0,0 +1,22 @@
...
+      "$ref": "uuid.json"
+    },
+    "ids": {
+      "description": "Array of holding/item unique identifiers which will be moved to the new instance/holding",
```
julianladisch, line 11:
```suggestion
      "description": "Ids of the holdings to move, or ids of the items to move.",
```

ramls/move.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Ids holder for moving items between holdings",
+  "type": "object",
+  "properties": {
+    "to": {
+      "description": "Id of the instance to which the holdings are moved, or id of the holding to which the items are moved.",
```
bohdan-suprun, line 7: How it is going to distinguish whether `to` is an instance ID or a holding ID? Will there be two separate APIs which will have the same schema?

If the above, I'd prefer to have two separate schemes as well:

* Move holdings 
```java
{
  "toInstanceId": required, uuid,
  "holdingsRecordIds": required, array
}
```

* move items:
```java
  "toHoldingsRecordId": required, uuid,
  "itemIds": required, array
}
```

What do you think?

khandramai: Hi @bohdan-suprun ,
we planned to use the same schema for both separate APIs for items and holdings moving. Actually, we don't need to introduce separate schemas for holdings and items because we are going to delegate all processing logic to corresponding API. Even if we send non-existent IDs for some items/holdings we just return the response with list of IDs for wich we could not update items/holdings.

khandramai: It is simple to introduce separate schemas for items/holdings moving. So let me please know if you see reason to make this. Thanks!

bohdan-suprun: Thanks, @khandramai. Leaving it with you to decide.

bohdan-suprun: > It is simple to introduce separate schemas for items/holdings moving. So let me please know if you see reason to make this.

Yes, I think it will make API clearer for consumers. And less error prone, in terms of specifying wrong UUID for a property.

bohdan-suprun:
@siarhei-charniak, @khandramai there is a schema error reported:

```
ERROR: lint-raml-schema: Trouble loading move.json: Invalid control character at: line 7 column 126 (char 298)
``` 

please take a look.

ramls/move.json
khandramai: @bohdan-suprun, thanks. It makes sense for me.
@siarhei-charniak could you please implement this proposal?

siarhei-charniak:
> @siarhei-charniak, @khandramai there is a schema error reported:
> 
> ```
> ERROR: lint-raml-schema: Trouble loading move.json: Invalid control character at: line 7 column 126 (char 298)
> ```
> 
> please take a look.

Already fixed

ramls/holdings_move.json
```diff
@@ -0,0 +1,22 @@
...
+      "description": "Ids of the holdings to move.",
+      "type": "array",
+      "items": {
+        "$ref": "uuid.json"
+      }
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "toInstanceId", "holdingsRecordIds"
```
marcjohnson-kint, line 20: Typically I think each property is presented on a separate line

siarhei-charniak: fixed

PR #230, opened 2020-06-18
MODINV-311: Adding or changing notes in quickMARC creates error in Instance record notes [BUGFIX].

VRohach:
1. Notes for holdings and items moved to the separate files.

### DIFF
--- a/ramls/holdings-notes.json
+++ b/ramls/holdings-notes.json
@@ -1,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Note for holding record",
+  "javaType": "org.folio.rest.jaxrs.model.HoldingsNotes",
+  "type": "object",
+  "additionalProperties": false,
+    "properties": {
+      "holdingsNoteTypeId": {
+        "type": "string",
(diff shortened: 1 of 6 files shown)


VRohach:
> Please can a test be added to demonstrate the bug and that it is fixed.
> 
> Why have the properties for the notes been extracted into separate schema? How does that fix the bug?

Yes, sure. I added them.

The main problem was regarding conflicts of generated "Note"-classes. The had the same names, but with different fields in it(for Holdings, Item, and Instance).

marcjohnson-kint:
> > Please can a test be added to demonstrate the bug and that it is fixed.
> > Why have the properties for the notes been extracted into separate schema? How does that fix the bug?
> 
> Yes, sure. I added some checks for this logic.
> 
> The main problem was regarding conflicts of generated "Note"-classes. They had the same names, but with different fields in it.
> 
> So, I moved Note from Holdings and Item entities in the separate file with specifying unique "javaType".

@VRohach Thank you for explaining the issue and this workaround.

It's a shame that the Java code generation has to leak into the interface definitions. That challenge is way beyond the scope of this issue.

PR #232, opened 2020-06-19
MODINV-309 - API inventory/items/move for items Move

khandramai:
[MODINV-309](https://issues.folio.org/browse/MODINV-309) - API inventory/items/move for items Move


ramls/inventory.raml
```diff
@@ -76,7 +77,30 @@ resourceTypes:
               body:
                 text/plain:
                   example: "Internal server error"
-
+    /move:
```
bohdan-suprun, line 80: does it make sense to have a separate move API? (i.e. inventory-move.raml)?

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public static final String TO_HOLDINGS_RECORD_ID = "toHoldingsRecordId";
+  public static final String ITEM_IDS = "itemIds";
+
+  private String toHoldingsRecordId;
+  private List<String> itemIds = new ArrayList<>();
+
+  public ItemsMove() {
+    super();
+  }
```
bohdan-suprun, line 16: Can it be removed?

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -107,6 +115,57 @@ public void register(Router router) {
...
+    JsonObject itemsMoveJsonRequest = routingContext.getBodyAsJson();
+
+    Optional<ValidationError> validationError = itemsMoveHasRequiredFields(itemsMoveJsonRequest);
+    if (validationError.isPresent()) {
+      unprocessableEntity(routingContext.response(), validationError.get());
+    } else {
+      ItemsMove itemsMove = MoveUtil.jsonToItemsMove(itemsMoveJsonRequest);
+      List<String> itemIdsToUpdate = itemsMove.getItemIds();
+      try {
+        storage.getItemCollection(context).findByCql(
```
bohdan-suprun, line 130: This client is not safe for fetching records using a long CQL query. It causes `414 URI Too Long` error when with > ~40-50 UUIDs in query. Please use `org.folio.inventory.storage.external.MultipleRecordsFetchClient` instead.

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -107,6 +115,57 @@ public void register(Router router) {
...
+  private void updateManyItemsWithNewHoldingsRecordId(
+    RoutingContext routingContext,
+    WebContext context,
+    MultipleRecords<Item> itemsToUpdate,
+    String toHoldingsRecordId,
+    List<String> itemIdsToUpdate) {
+
+    ItemCollection storageItemCollection = storage.getItemCollection(context);
+
+    List<CompletableFuture<Item>> updates = itemsToUpdate.records.stream()
```
bohdan-suprun, line 150: There is new batch API for item update `/item-storage/batch/synchronous` (use `?upsert=true`)

bohdan-suprun, line 150: Do we need some additional validation, e.g.?:
* Is holdings record exists;
* Is it allowed to move to a holdings from an another instance;
* I am not sure if it is ok to move a MARC controlled item, maybe @OleksiiKuzminov can clarify.

bohdan-suprun, line 150: For cross instance updates we need to notify mod-circulation about it, because there are some places where item data is duplicated and stored in mod-circulation-storage (e.g. requests store identifiers).

src/main/java/org/folio/inventory/support/MoveUtil.java
```diff
@@ -0,0 +1,28 @@
...
+    return new ItemsMove().withToHoldingsRecordId(itemRequest.getString(TO_HOLDINGS_RECORD_ID))
+      .withItemIds(toListOfStrings(itemRequest.getJsonArray(ItemsMove.ITEM_IDS)));
+  }
+
+  public static JsonObject itemsMoveMapToJson(ItemsMove itemsMove) {
+    JsonObject itemsMoveJson = new JsonObject();
+    itemsMoveJson.put(TO_HOLDINGS_RECORD_ID, itemsMove.getToHoldingsRecordId());
+    itemsMoveJson.put(ITEM_IDS, new JsonArray(itemsMove.getItemIds()));
+    return itemsMoveJson;
+  }
```
marcjohnson-kint, line 27: Why does the production code need to create a JSON representation a request to move items, is this not received from a client?

src/main/java/org/folio/inventory/support/MoveUtil.java
```diff
@@ -0,0 +1,28 @@
...
+
+public final class MoveUtil {
+
+  private MoveUtil() {
+  }
+
+  public static ItemsMove jsonToItemsMove(JsonObject itemRequest) {
+    return new ItemsMove().withToHoldingsRecordId(itemRequest.getString(TO_HOLDINGS_RECORD_ID))
+      .withItemIds(toListOfStrings(itemRequest.getJsonArray(ItemsMove.ITEM_IDS)));
+  }
```
marcjohnson-kint, line 20: Could this either be a static method on the `ItemsMove` class or a separate class named specifically to do this mapping?

src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
```diff
@@ -23,6 +23,10 @@ public static void created(HttpServerResponse response,
     response(response, body, 201);
   }
 
+  public static void createdWithEmptyResponse(HttpServerResponse response) {
+    response(response, new JsonObject(), 201);
+  }
```
marcjohnson-kint, line 28: By empty response, is the intention for a response with a JSON object with no properties or a response with no body at all?

src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
```diff
@@ -67,6 +71,12 @@ public static void unprocessableEntity(HttpServerResponse response, String error
     response(response, new JsonObject().put("errors", error), 422);
   }
 
+  public static void someOfEntitiesNotFound(HttpServerResponse response, List<String> ids) {
+    JsonObject errors = new JsonObject();
+    errors.put("ids", ids);
+    response(response, new JsonObject().put("errors", errors), 404);
+  }
```
marcjohnson-kint, line 78: If this is using the standard JSON error representation, this might be better communicated to the client by using the 422 status code instead.

src/main/java/org/folio/inventory/validation/ItemsMoveValidator.java
```diff
@@ -0,0 +1,37 @@
...
+import org.folio.inventory.support.JsonArrayHelper;
+import org.folio.inventory.support.JsonHelper;
+import org.folio.inventory.support.http.server.ValidationError;
+
+import io.vertx.core.json.JsonObject;
+
+public class ItemsMoveValidator {
+
+  private ItemsMoveValidator() {
+  }
```
marcjohnson-kint, line 19:
```suggestion
  private ItemsMoveValidator() { }
```

src/test/java/api/support/ApiRoot.java
```diff
@@ -37,6 +37,12 @@ public static URL items()
     return new URL(String.format("%s/items", inventory()));
   }
 
+  public static URL move()
```
marcjohnson-kint, line 40: As I understand it, there will later be different kinds of move requests. Could it be worthwhile to refer to this one as specific to items?

```suggestion
  public static URL moveItems()
```

src/test/java/api/support/builders/ItemsMoveRequestBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+    if (value != null) {
+      itemRequest.put(property, value.toString());
+    }
+  }
+
+  private void includeWhenPresent(JsonObject itemRequest, String property, JsonArray value) {
+    if (value != null) {
+      itemRequest.put(property, value);
+    }
+  }
```
marcjohnson-kint, line 40: Are these duplicated from another JSON builder?

src/test/java/api/support/http/BusinessLogicInterfaceUrls.java
```diff
@@ -26,6 +26,10 @@ public static URL isbns(String subPath) {
     return getUrl("/isbn", subPath);
   }
 
+  public static URL move(String subPath) {
+    return getUrl("/inventory/items/move", subPath);
+  }
```
marcjohnson-kint, line 31: Why is a sub path needed?

src/test/java/api/items/ItemApiMoveExamples.java
```diff
@@ -0,0 +1,148 @@
...
+import io.vertx.core.json.JsonObject;
+import junitparams.JUnitParamsRunner;
+
+@RunWith(JUnitParamsRunner.class)
+public class ItemApiMoveExamples extends ApiTests {
+
+  private static final String HOLDINGS_RECORD_ID = "holdingsRecordId";
+
+  @Test
+  public void canCreateItemsMove() throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
```
marcjohnson-kint, line 44: As I understand it, this isn't creating any new resources or records, it is moving existing items to a different holdings record?

```suggestion
  public void canMoveIItemsToDifferentHoldingsRecord() throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
```

src/test/java/api/items/ItemApiMoveExamples.java
```diff
@@ -0,0 +1,148 @@
...
+      .withStatus(ItemStatusName.AVAILABLE.value()));
+
+    JsonObject itemsMoveRequestBody = new ItemsMoveRequestBuilder(newHoldingId,
+        new JsonArray(Arrays.asList(createItem1.getId(), createItem2.getId()))).create();
+
+    CompletableFuture<Response> postItemsMoveCompleted = new CompletableFuture<>();
+    okapiClient.post(ApiRoot.move(), itemsMoveRequestBody, ResponseHandler.any(postItemsMoveCompleted));
+    Response postItemsMoveResponse = postItemsMoveCompleted.get(5, TimeUnit.SECONDS);
+
+    assertThat(postItemsMoveResponse.getStatusCode(), is(201));
```
marcjohnson-kint, line 69: What new records or resource are being created?

If there aren't any, why does this respond with a CREATED status rather than OK?

src/test/java/api/items/ItemApiMoveExamples.java
```diff
@@ -0,0 +1,148 @@
...
+      .getJson();
+    JsonObject updatedItem2 = itemsClient.getById(createItem2.getId())
+      .getJson();
+
+    Assert.assertEquals(newHoldingId.toString(), updatedItem1.getString(HOLDINGS_RECORD_ID));
+    Assert.assertEquals(newHoldingId.toString(), updatedItem2.getString(HOLDINGS_RECORD_ID));
+  }
+
+  @Test
+  public void canCreatePartialItemsMove() throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
```
marcjohnson-kint, line 82:
```suggestion
  public void shouldReportErrorsWhenOnlySomeRequestedItemsCouldNotBeMoved() throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
```

src/test/java/api/items/ItemApiMoveExamples.java
```diff
@@ -0,0 +1,148 @@
...
+    assertThat(notFoundIds.size(), is(1));
+    assertThat(notFoundIds.get(0), equalTo(nonExistedItemId.toString()));
+
+    JsonObject updatedItem1 = itemsClient.getById(createItem1.getId())
+      .getJson();
+    assertThat(newHoldingId.toString(), equalTo(updatedItem1.getString(HOLDINGS_RECORD_ID)));
+  }
+
+  @Test
+  public void cannotCreateMoveWithoutToHoldingsRecordId()
```
marcjohnson-kint, line 121:
```suggestion
  public void cannotMoveItemsToUnspecifiedHoldingsRecord()
```

src/test/java/api/items/ItemApiMoveExamples.java
```diff
@@ -0,0 +1,148 @@
...
+      throws InterruptedException, MalformedURLException, TimeoutException, ExecutionException {
+    CompletableFuture<Response> postCompleted = new CompletableFuture<>();
+    JsonObject itemMoveWithoutToHoldingsRecordId = new ItemsMoveRequestBuilder(null,
+        new JsonArray(Collections.singletonList(UUID.randomUUID()))).create();
+    okapiClient.post(ApiRoot.move(), itemMoveWithoutToHoldingsRecordId, ResponseHandler.any(postCompleted));
+    Response postResponse1 = postCompleted.get(5, TimeUnit.SECONDS);
+    assertThat(postResponse1.getStatusCode(), is(422));
+    assertThat(postResponse1.getContentType(), containsString(APPLICATION_JSON));
```
marcjohnson-kint, line 129: I think this test could be easier to follow if the various steps were separate by more whitespace

marcjohnson-kint, line 129: Please also include checks that the correct error was provided in the response

src/test/java/api/items/ItemApiMoveExamples.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @Test
+  public void cannotCreateMoveWithoutItemIds()
```
marcjohnson-kint, line 133:
```suggestion
  public void cannotMoveUnspecifiedItems()
```

src/test/java/api/items/ItemApiMoveExamples.java
```diff
@@ -0,0 +1,148 @@
...
+      throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException {
+    CompletableFuture<Response> postCompleted = new CompletableFuture<>();
+    JsonObject itemMoveWithoutItemIds = new ItemsMoveRequestBuilder(UUID.randomUUID(), new JsonArray()).create();
+    okapiClient.post(ApiRoot.move(), itemMoveWithoutItemIds, ResponseHandler.any(postCompleted));
+    Response postResponse2 = postCompleted.get(5, TimeUnit.SECONDS);
+    assertThat(postResponse2.getStatusCode(), is(422));
+    assertThat(postResponse2.getContentType(), containsString(APPLICATION_JSON));
```
marcjohnson-kint, line 140: Similarly to the above, please also check the specific error is included in the response and I think it would be easier to follow this test if the steps were separated by some whitespace

ramls/inventory.raml
```diff
@@ -76,7 +77,30 @@ resourceTypes:
...
+      type:
+        collection-item:
+          exampleItem: !include examples/items_move.json
+          schema: move
+      post:
+        responses:
+          201:
+            description: "Items moved to another holdings record"
+          404:
+            description: "Some of items not moved"
```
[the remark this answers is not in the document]
khandramai: Yes, you are correct. Do you agree with this?

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
```diff
@@ -0,0 +1,35 @@
+package org.folio.inventory.domain.items;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ItemsMove {
```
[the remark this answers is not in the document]
khandramai: Yes, OK. From my perspective MoveItemsCommand looks better.

src/main/java/org/folio/inventory/resources/Items.java
khandramai: @bohdan-suprun 
Does using of PUT inventory/items/ instead of items-storage/items/ cover mod-circulation notifications? I mean if I use items updating on inventory (not items-storage directly) then we don't need to implement additional logic for circulation notifications?

ramls/inventory.raml
khandramai: It's OK for me. Let's move this endpoint to separate raml API. +  we will add the same for holdings in the nearest future.

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
khandramai: This was created to prevent code smell for empty public constructor since I use Builder. OK I can use parameterized constructor here instead of empty one with builders.

descriptors/ModuleDescriptor-template.json
```diff
@@ -66,6 +66,15 @@
...
         },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/inventory/items/move",
+          "permissionsRequired": ["inventory.items.move.item.post"],
+          "modulePermissions": [
+            "inventory-storage.items.item.put",
+            "inventory-storage.items.collection.get"
+          ]
+        },
```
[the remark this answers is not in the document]
khandramai: Yes, OK. And I'll move this to separate rail as we discussed here https://github.com/folio-org/mod-inventory/pull/232#discussion_r442786805

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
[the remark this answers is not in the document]
khandramai: Sure, I'll use parameterized constructor here (see https://github.com/folio-org/mod-inventory/pull/232#discussion_r442828291)

src/main/java/org/folio/inventory/resources/Items.java
marcjohnson-kint: We almost certainly should stop an attempt to move to a holdings record that does not exist

marcjohnson-kint: >  am not sure if it is ok to move a MARC controlled item

That is an excellent requirements question

marcjohnson-kint: At the moment, that information is not kept in sync until the request itself is updated. This is a known significant limitation.

I don't think we should try to extend that as part of this work. 

@bohdan-suprun does that make sense?

ramls/inventory.raml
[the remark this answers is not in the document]
marcjohnson-kint: I don't think that a 404 Not Found should be used to report that some items were not moved.

I don't understand how those fit together.

I think it depends why an item cannot be moved. Personally, I think if this occurs after the initial validation and some items can / have been moved, it would be clearer to return a 200 which describes which items have been moved and which ones have not (and why).

ramls/inventory.raml
marcjohnson-kint: Please make the version of this new interface 0.1 to mark it as experimental

ramls/inventory.raml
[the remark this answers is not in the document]
khandramai: As an option we can return 202 Accepted instead of 201 Created just to indicate that request was accepted but part of items couldn't updated and completely remove 404 from contract. But from my perspective we need to show that smth goes wrong by http status. 

@NikitaSedyx from the UI perspective what option is more appropriate to show error: 1) return 404 with list of non-updated item ids or 2) return all the time successful http code (even though some of items weren't updated) with list of non-updated item ids if some of them weren't updated?

src/main/java/org/folio/inventory/resources/Items.java
bohdan-suprun: @khandramai, @marcjohnson-kint sure, I think it is out of scope to sync circulation. It worth letting POs know that we have this challenge.

src/test/java/api/support/http/BusinessLogicInterfaceUrls.java
bohdan-suprun: It seems this is redundant method, there is already one defined in [src/test/java/api/support/ApiRoot.java](https://github.com/folio-org/mod-inventory/pull/232/files#diff-91a850ea70ae373f4acd55dd58aba487R40)

ramls/inventory.raml
marcjohnson-kint: > does it make sense to have a separate move API? (i.e. inventory-move.raml)?

Maybe give it the name `inventory-move-items` for moving items between holdings, to distinguish it from moving holdings between instances

ramls/inventory.raml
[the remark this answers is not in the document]
marcjohnson-kint: > As an option we can return 202 Accepted instead of 201 Created just to indicate that request was accepted but part of items couldn't updated and completely remove 404 from contract. But from my perspective we need to show that smth goes wrong by http status.

I agree that 404 is not appropriate and should be removed. I also agree that the response needs to convey that something went wrong. 

I don't think 202 is an appropriate status code to convey partial success or failure, that is both synchronous and known at the point of initial response.

As I understand it, 202 is intended for asynchronous processing of a request (and is meant to also convey a way of determining the status of that processing).

`The 202 (Accepted) status code indicates that the request has been accepted for processing, but the processing has not been completed.` (https://tools.ietf.org/html/rfc7231#section-6.3.3)

I'm not sure there is a status code that can convey that partial status appropriately on it's own. 

As I understand, HTTP is intended to be resource-centric. This can present challenges when a resource represents a process, especially when that process involves manipulating more than one resources. 

I think the scenario being described, from the perspective of the resource describing the process, has been handled ok. That within that process, some parts have succeeded and others have failed is below that level and cannot be conveyed in a single status.

Alternatively, any single failure could be considered a failure of the process overall, if that works better for folks. Either way, the client is going to need to know that this status really means that it needs to `investigate further by inspecting the body (or other state of the system)`. 

Personally I think that 200 is more suited to this investigation than 4XX or 5XX.

I'm sorry that I didn't manage to convey this challenge better in the discussions on the design document when I suggested that it might be better for a client to submit multiple requests rather than a group / batch.

src/main/java/org/folio/inventory/resources/Items.java
khandramai: @bohdan-suprun @marcjohnson-kint I have concern on it. Looks like this API works like transaction: it updates all or nothing from items list [item-sync.html](https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/p/item-sync.html). But we need to know what items aren't updated. What do you think on this?

marcjohnson-kint: I believe that is true, it was decided that those batch APIs would be all or nothing. That may well make them inappropriate for use in this case, and mean the records have to be updated individually.

ramls/inventory.raml
khandramai: @marcjohnson-kint let it be more general inventory-move.raml. I think it will be good to have items and holdings move in the one file.

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
khandramai: ItemsMove is removed at all.

src/main/java/org/folio/inventory/resources/Items.java
khandramai: Done

ramls/inventory.raml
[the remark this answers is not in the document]
khandramai: API will return 201 all the time. In case of some failures it will return list of nonUpdatedIds but the same http status 201.

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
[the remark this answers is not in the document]
khandramai: ItemsMove is removed at all.

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public ItemsMove withToHoldingsRecordId(String toHoldingsRecordId) {
+    this.toHoldingsRecordId = toHoldingsRecordId;
+    return this;
+  }
+
+  public ItemsMove withItemIds(List<String> itemIds) {
+    this.itemIds = itemIds;
+    return this;
+  }
```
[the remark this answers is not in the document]
khandramai: ItemsMove is removed at all.

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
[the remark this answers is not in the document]
khandramai: ItemsMove is removed at all.

src/main/java/org/folio/inventory/domain/items/ItemsMove.java
```diff
@@ -0,0 +1,35 @@
...
+    super();
+  }
+
+  public String getToHoldingsRecordId() {
+    return toHoldingsRecordId;
+  }
+
+  public List<String> getItemIds() {
+    return itemIds;
+  }
```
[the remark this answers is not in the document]
khandramai: ItemsMove is removed at all. Let's operate with JSON from request.

src/main/java/org/folio/inventory/support/MoveUtil.java
khandramai: Removed. Not needed.

src/main/java/org/folio/inventory/support/MoveUtil.java
khandramai: Removed. Not needed.

src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
khandramai: Empty response (without empty JSON). Fixed.

src/main/java/org/folio/inventory/validation/ItemsMoveValidator.java
khandramai: done

src/test/java/api/support/builders/ItemsMoveRequestBuilder.java
khandramai: refactored to prevent duplication

src/test/java/api/items/ItemApiMoveExamples.java
khandramai: @marcjohnson-kint so, this is not clear. From the one side this is POST with creation of some kind "virtual move" of items. So 201 is more appropriate from my perspective. 200 looks useful for GET some resource. But here we don't need to retrieve any resource in the happy path. So it's up to you. Let me know if I need to change this code.

khandramai:
@bohdan-suprun @marcjohnson-kint please, review last changes. Thank you!

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
-      "version": "10.5",
+      "version": "10.6",
```
bohdan-suprun, line 7: This seems should be reverted

src/main/java/org/folio/inventory/resources/MoveApi.java
```diff
@@ -0,0 +1,134 @@
...
+    Optional<ValidationError> validationError = itemsMoveHasRequiredFields(itemsMoveJsonRequest);
+    if (validationError.isPresent()) {
+      unprocessableEntity(routingContext.response(), validationError.get());
+    } else {
+      String toHoldingsRecordId = itemsMoveJsonRequest.getString(TO_HOLDINGS_RECORD_ID);
+      List<String> itemIdsToUpdate = toListOfStrings(itemsMoveJsonRequest.getJsonArray(ITEM_IDS));
+      try {
+
+        OkapiHttpClient okapiClient = createHttpClient(routingContext, context);
+        CollectionResourceClient itemsStorageClient = createItemsStorageClient(okapiClient, context);
```
bohdan-suprun, line 64: Take a look at `org.folio.inventory.storage.external.Clients`, you can add a `getItemsStorageClient` method in that class. And then extend this `moveItems` method to `moveItems(RoutingContex, WebContext, Clients)`.

You also need to change the `register` method to:

```
@Override
public void register(Router router) {
router.post("/inventory/items/move")
    .handler(handle(this::moveItems));
}
```

src/main/java/org/folio/inventory/resources/MoveApi.java
```diff
@@ -0,0 +1,134 @@
...
+        MultipleRecordsFetchClient multipleRecordsFetchClient = createItemsFetchClient(itemsStorageClient);
+
+        multipleRecordsFetchClient.find(itemIdsToUpdate, this::fetchItemsCql)
```
bohdan-suprun, line 67: Could you please verify that the holdings with the id exists before starting the fetch and update

src/main/java/org/folio/inventory/resources/MoveApi.java
```diff
@@ -0,0 +1,134 @@
...
+      }
+    }
+  }
+
+  private void updateItems(RoutingContext routingContext, WebContext context, String toHoldingsRecordId,
+      List<String> itemIdsToUpdate, List<Item> itemsToUpdate) {
+    ItemCollection storageItemCollection = storage.getItemCollection(context);
+
+    List<CompletableFuture<Item>> updates = itemsToUpdate.stream()
+      .map(item -> storageItemCollection.update(item.withHoldingId(toHoldingsRecordId)))
```
bohdan-suprun, line 86: It seems the items should already have the new holdings id, it is populated in the calling method.

src/main/java/org/folio/inventory/support/ItemUtil.java
```diff
@@ -26,10 +29,155 @@
...
+  public static final String MATERIAL_TYPE = "materialType";
+  public static final String PERMANENT_LOCATION = "permanentLocation";
+  public static final String TEMPORARY_LOCATION = "temporaryLocation";
+  public static final String PERMANENT_LOAN_TYPE = "permanentLoanType";
+  public static final String TEMPORARY_LOAN_TYPE = "temporaryLoanType";
 
   private ItemUtil() {
   }
 
+  public static Item jsonToServerItem(JsonObject itemFromServer) {
```
bohdan-suprun, line 50: I'd use `storedItem` here, what do you think? 

```suggestion
  public static Item storedItemJsonToItem(JsonObject itemFromServer) {
```

src/main/java/org/folio/inventory/support/ItemUtil.java
```diff
@@ -26,10 +29,155 @@
...
+      .withStatisticalCodeIds(statisticalCodeIds)
+      .withPurchaseOrderLineidentifier(itemFromServer.getString(Item.PURCHASE_ORDER_LINE_IDENTIFIER))
+      .withTags(tags)
+      .withLastCheckIn(LastCheckIn.from(itemFromServer.getJsonObject("lastCheckIn")))
+      .withEffectiveCallNumberComponents(
+        EffectiveCallNumberComponents.from(itemFromServer.getJsonObject("effectiveCallNumberComponents")));
+
+  }
+
+  public static final JsonObject serverItemToJson(Item item) {
```
bohdan-suprun, line 129:
```suggestion
  public static final JsonObject toStoredItemRepresentation(Item item) {
```

src/main/java/org/folio/inventory/support/ItemUtil.java
bohdan-suprun: Or
```suggestion
  public static Item fromStoredItemRepresentation(JsonObject itemFromServer) {
```

src/main/java/org/folio/inventory/resources/MoveApi.java
khandramai: Done. Unit test to cover this case was was added.

src/main/java/org/folio/inventory/resources/MoveApi.java
khandramai: @bohdan-suprun are you fine to save current approach to maintain uniformity with other API's implementation? I would like to skip this improvement because MoveApi was implemented based on ItemApi and other resources.

khandramai:
@bohdan-suprun @marc-johnson-kint I added validation that the holdings record exist. Thanks!

src/main/java/org/folio/inventory/resources/MoveApi.java
bohdan-suprun: sure, no prob. I'm resolving the conversation.

descriptors/ModuleDescriptor-template.json
```diff
@@ -235,6 +235,21 @@
...
+    {
+      "id": "inventory-move",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["POST"],
+          "pathPattern": "/inventory/items/move",
+          "permissionsRequired": ["inventory.items.move.item.post"],
+          "modulePermissions": [
+            "inventory-storage.items.item.put",
```
bohdan-suprun, line 247: Please add permission to fetch a holdings record by id.

ramls/inventory.raml
marcjohnson-kint: Ok, that means that any potential implementor would need to implement both.

src/main/java/org/folio/inventory/resources/Items.java
marcjohnson-kint: > I am not sure if it is ok to move a MARC controlled item

What was the answer to this question?

ramls/inventory.raml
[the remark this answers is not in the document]
marcjohnson-kint: @khandramai I personally think that responding with an overall process status is a good idea.

Please can you help me understand why 201 CREATED was chosen over 200 OK? What resource is being created during this process?

src/test/java/api/items/ItemApiMoveExamples.java
marcjohnson-kint: > From the one side this is POST with creation of some kind "virtual move" of items. So 201 is more appropriate from my perspective. 200 looks useful for GET some resource.

I afraid I don't really follow how any resources are created. If a resource was created to represent the move itself, I could understand using 201 CREATED rather than 200 OK. Maybe that is what you mean by `virtual move`?

ramls/inventory.raml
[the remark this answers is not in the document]
khandramai: I mean some kind we "virtual move" created. It looks not like pure REST but some action was performed and we have result of items moving. But I'm fine with 200 if you think that it is more appropriate.

src/main/java/org/folio/inventory/resources/MoveApi.java
```diff
@@ -0,0 +1,143 @@
...
+              MultipleRecordsFetchClient multipleRecordsFetchClient = createItemsFetchClient(itemsStorageClient);
+
+              multipleRecordsFetchClient.find(itemIdsToUpdate, this::fetchItemsCql)
+                .thenAccept(jsons -> {
+                  List<Item> itemsToUpdate = jsons.stream()
+                    .map(ItemUtil::fromStoredItemRepresentation)
+                    .map(item -> item.withHoldingId(toHoldingsRecordId))
+                    .collect(toList());
+                  updateItems(routingContext, context, itemIdsToUpdate, itemsToUpdate);
+                });
```
marcjohnson-kint, line 78: What happens when this process fails? Is the intention that the `catch` statement below it will be executed?

ramls/inventory.raml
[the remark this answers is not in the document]
marcjohnson-kint: Personally, I think 200 is more appropriate, as the process completed and no new resources were created.

This might be because I'm missing what this `virtual move` is. Does it change any other state of the system other than the items that have been moved?

src/main/java/org/folio/inventory/resources/MoveApi.java
khandramai: this catch for checked  exception: java.net.MalformedURLException from createHttpClient and createItemsStorageClient methods. I think server error will be returned without items updating.

ramls/inventory.raml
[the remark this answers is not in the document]
khandramai: No, just item's state will be updated.

src/main/java/org/folio/inventory/resources/MoveApi.java
marcjohnson-kint: What happens if an exception is thrown within this block of code?

khandramai: ServerErrorResponse.internalError(routingContext.response(), e) will be executed and process will be stopped.

ramls/inventory.raml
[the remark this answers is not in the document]
khandramai: OK, I created 200 instead of 201.

src/main/java/org/folio/inventory/resources/MoveApi.java
marcjohnson-kint: How will that code be executed? By the `catch` statement below?

I'm asking because I don't think that exceptions thrown within `thenAccept` methods will throw the exception in the invoking context. Instead, they will provide a new `CompletionStage` that has failed exceptionally.

khandramai: Ah! OK. I have added exceptionally block for this possible exception. Thanks!

ramls/inventory.raml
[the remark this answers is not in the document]
marcjohnson-kint: Thank you

src/main/java/org/folio/inventory/resources/Items.java
bohdan-suprun: Asked Ann-Marie in jira.

bohdan-suprun: Response:
> Hi Bohdan Suprun SRS does not track which holdings or instance an item is linked to, so there should not be any issues with moving an item from one holdings to another, or from one instance to another.
>
> It will be slightly different when you get to the story of moving a holdings from one instance to another, but that's not in this story.

khandramai: So, we can merge it as is. Am I correct?

bohdan-suprun: Yes. We should keep in mind this restriction for holdings move feature.

khandramai:
So @marcjohnson-kint looks like all the comments were resolved. Are you OK to merge this?

khandramai:
Hi @marcjohnson-kint assuming all discussions have been resolved and fixed, I'm proceeding with merge of this PR. Could you please confirm?

marcjohnson-kint:
> Hi @marcjohnson-kint assuming all discussions have been resolved and fixed, I'm proceeding with merge of this PR. Could you please confirm?

I have some follow up questions, they might be mostly about requirements. I think those can wait.

I don't know when I'll have time to review this pull request again. I suggest, given it's already approved, that you merge at your convenience (I'll comment on it later if I feel the need to, but that is unlikely).

PR #238, opened 2020-06-29
Upgrade to log4j 2 MODINV-197

marcjohnson-kint:


pom.xml
```diff
@@ -30,14 +38,19 @@
...
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+      <version>${log4j2.version}</version>
     </dependency>
     <dependency>
-      <groupId>log4j</groupId>
-      <artifactId>log4j</artifactId>
-      <version>1.2.17</version>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-slf4j-impl</artifactId>
```
julianladisch, line 52: Can you add a comment why SLF 1.7 API is used (log4j-slf4j-impl) and not SLF 1.8 API (log4j-slf4j18-impl)?
https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/index.html

marcjohnson-kint: @julianladisch 

I based this change on work I did months ago. I wasn't aware that the other library existed. Are you saying we should use `log4j-slf4j18-impl` instead?

julianladisch: Java 11 uses the Java Module System and therefore requires SLF4J 1.8.x and log4j-slf4j18-impl. Using SLF 1.7 creates technical debt.

PR #243, opened 2020-07-23
MODINV-324 Add source-record states storing mechanism for QM edit workflow

OleksiiKuzminov:
fixed payloads

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateInstanceEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateInstanceEventHandler.java
@@ -3,6 +3,7 @@ package org.folio.inventory.dataimport.handlers.actions;
 import io.vertx.core.Future;
 import io.vertx.core.MultiMap;
 import io.vertx.core.Vertx;
+import io.vertx.core.json.Json;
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateInstanceEventHandler.java
```diff
@@ -64,23 +62,18 @@ public UpdateInstanceEventHandler(Storage storage, Context context) {
...
+          requestHeaders
+            .entries()
+            .forEach(entry -> headers.put(entry.getKey(), entry.getValue()));
+          OkapiConnectionParams params = new OkapiConnectionParams(headers, vertx);
+          if (ar.succeeded()) {
+
+            sendEventWithPayload(Json.encode(eventPayload), "QM_INVENTORY_INSTANCE_UPDATED", params)
+              .setHandler(v -> future.complete(ar.result()));
+          } else {
+            sendEventWithPayload(Json.encode(eventPayload), "QM_ERROR", params)
```
VRohach, line 75: @OleksiiKuzminov  I think that zipping should be added.

VRohach: Done.

PR #246, opened 2020-07-27
MODINV-299: Add 'Aged to lost' item status.

bohdan-suprun:
https://issues.folio.org/browse/MODINV-299 - Add 'Aged to lost' status to allowed item statuses list

### Changes: 

* Add new `Aged to lost` item status;
* Update `inventory` API version.
* Require `item-storage:8.5`


bohdan-suprun:
Depends on https://github.com/folio-org/mod-inventory-storage/pull/478

PR #248, opened 2020-08-03
MODINV-329 - Move: endpoints invalid status code

khandramai:
Fix for http status code mismatching [MODINV-329](https://issues.folio.org/browse/MODINV-329).

### DIFF
--- a/src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
+++ b/src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
@@ -31,13 +31,13 @@ public class JsonResponse {
   }
 
   public static void successWithEmptyBody(HttpServerResponse response) {
-    emptyResponse(response, 200);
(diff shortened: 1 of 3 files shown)


marcjohnson-kint:
It's unfortunately too late now to provide this feedback. 

I don't understand why the implementation was changed from 200 to 201, rather than changing the interface definition from 201 to 200 (assuming this was considered a mistake). 

From what I understand of this process, no new HTTP resources (or records) are created, and so 201 isn't really an appropriate status code.

khandramai:
It's OK. I can introduce needed changes in the [PR#249](https://github.com/folio-org/mod-inventory/pull/249). So, let's decide about HTTP code, content-type and body content. It's rather simple to do this yet. @marcjohnson-kint  let's continue discussion in the [PR#249](https://github.com/folio-org/mod-inventory/pull/249).

PR #249, opened 2020-08-04
MODINV-329 - Move: endpoints invalid status code

khandramai:
Follow-up to [PR#248](https://github.com/folio-org/mod-inventory/pull/248). UI-client requires empty json in body instead of empty body.

### DIFF
--- a/ramls/inventory-move.raml
+++ b/ramls/inventory-move.raml
@@ -26,7 +26,7 @@ traits:
       application/json:
         type: items_move
     responses:
-      201:
+      200:
         description: "Items moved to another holdings record"
         body:
           application/json:
@@ -49,7 +49,7 @@ traits:
       application/json:
         type: holdings_move
     responses:
-      201:
+      200:
         description: "Holdings record moved to another instance"
         body:
           application/json:
--- a/ramls/inventory-move.raml
+++ b/ramls/inventory-move.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Inventory Move API
-version: v0.1
+version: v0.2
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost
 
--- a/src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
+++ b/src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
@@ -1,6 +1,5 @@
 package org.folio.inventory.support.http.server;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
@@ -29,16 +28,6 @@ public class JsonResponse {
                              JsonObject body) {
 
     response(response, body, 200);
-  }
-
-  public static void successWithEmptyIds(HttpServerResponse response) {
-    successWithIds(response, new ArrayList<>());
-  }
-
-  public static void successWithIds(HttpServerResponse response, List<String> ids) {
-    JsonObject nonUpdatedIds = new JsonObject();
-    nonUpdatedIds.put("nonUpdatedIds", ids);
-    response(response, nonUpdatedIds, 200);
   }
 
   public static void unprocessableEntity(
@@ -95,5 +84,4 @@ public class JsonResponse {
     response.write(buffer);
     response.end();
   }
-
 }
--- a/src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
+++ b/src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
@@ -31,13 +31,13 @@ public class JsonResponse {
   }
 
   public static void successWithEmptyBody(HttpServerResponse response) {
-    emptyResponse(response, 200);
+    emptyResponse(response, 201);
   }
 
   public static void successWithIds(HttpServerResponse response, List<String> ids) {
     JsonObject nonUpdatedIds = new JsonObject();
     nonUpdatedIds.put("nonUpdatedIds", ids);
-    response(response, nonUpdatedIds, 200);
+    response(response, nonUpdatedIds, 201);
   }
 
   public static void unprocessableEntity(
--- a/src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
+++ b/src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
@@ -1,5 +1,6 @@
 package org.folio.inventory.support.http.server;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
@@ -30,14 +31,14 @@ public class JsonResponse {
     response(response, body, 200);
   }
 
-  public static void successWithEmptyBody(HttpServerResponse response) {
-    emptyResponse(response, 201);
+  public static void successWithEmptyIds(HttpServerResponse response) {
+    successWithIds(response, new ArrayList<>());
   }
 
   public static void successWithIds(HttpServerResponse response, List<String> ids) {
     JsonObject nonUpdatedIds = new JsonObject();
     nonUpdatedIds.put("nonUpdatedIds", ids);
-    response(response, nonUpdatedIds, 201);
+    response(response, nonUpdatedIds, 200);
   }
 
   public static void unprocessableEntity(
@@ -95,9 +96,4 @@ public class JsonResponse {
     response.end();
   }
 
-  private static void emptyResponse(HttpServerResponse response, int statusCode) {
-    response.setStatusCode(statusCode);
-    response.putHeader(HttpHeaders.CONTENT_TYPE, String.format("%s; charset=utf-8", ContentType.APPLICATION_JSON));
-    response.end(new JsonObject().toBuffer());
-  }
 }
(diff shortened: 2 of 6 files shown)


marcjohnson-kint:
Why does the UI require sending empty JSON?

NikitaSedyx:
@marc-johnson  https://github.com/folio-org/stripes-connect/blob/master/RESTResource/RESTResource.js#L530

uladislausamets:
> @marcjohnson-kint @NikitaSedyx @uladislausamets ,
> let's decide about final version - (1) HTTP code 200, Content-Type: application/json, Empty json in the body OR HTTP code 200, no content type, empty body?

@khandramai 1 version looks good for me

marcjohnson-kint:
> @marc-johnson https://github.com/folio-org/stripes-connect/blob/master/RESTResource/RESTResource.js#L530
> it throws an error that json can't be parsed

That code appears to be for response interpretation, how does that affect the body of the request, or have I misunderstood?

NikitaSedyx:
this endpoint can return data as well, so empty object works here

marcjohnson-kint:
> this endpoint can return data as well, so empty object works here

What data does it usually respond with? Under what circumstances does it respond with no data?

khandramai:
If all items/holdings were moved - empty json, if smth was failed - body with failed ids.

marcjohnson-kint:
> If all items/holdings were moved - empty json, if smth was failed - body with failed ids. But http status for both cases is 200.

Can you provide an example of the JSON for when some records failed to be moved?

khandramai:
> > If all items/holdings were moved - empty json, if smth was failed - body with failed ids. But http status for both cases is 200.
> 
> Can you provide an example of the JSON for when some records failed to be moved?

https://github.com/folio-org/mod-inventory/blob/master/ramls/examples/items_move.json

marcjohnson-kint:
> > > If all items/holdings were moved - empty json, if smth was failed - body with failed ids. But http status for both cases is 200.
> > 
> > 
> > Can you provide an example of the JSON for when some records failed to be moved?
> 
> https://github.com/folio-org/mod-inventory/blob/master/ramls/examples/items_move.json

That appears to be an example of the JSON for a request to move items, not a response when some moves have failed

NikitaSedyx:
@marc-johnson 
https://github.com/folio-org/mod-inventory/blob/master/ramls/move_response.json here you can see that endpoint can returns data, we just need an empty object to make UI work
it was discussed and reviewed by you and Slava in the main PR (where endpoint was introduced)

current discussion (when everything was discussed before) blocks merge of this fix and UI implementation with PO review

marcjohnson-kint:
> @marc-johnson
> https://github.com/folio-org/mod-inventory/blob/master/ramls/move_response.json here you can see that endpoint can return data, we just need an empty object to make UI work
> it was discussed and reviewed by you and Slava in the main PR (where endpoint was introduced)
> 
> current discussion (when everything was discussed before) blocks merge of this fix and UI implementation with PO review

@NikitaSedyx 

I provided feedback on #248 after it was merged and closed (an inherent challenge in FOLIO's development process).

@khandramai expressed [interest](https://github.com/folio-org/mod-inventory/pull/248#issuecomment-668608113) in having a conversation about it in this pull request:
> It's OK. I can introduce needed changes in the PR#249. So, let's decide about HTTP code, content-type and body content. It's rather simple to do this yet. @marcjohnson-kint let's continue discussion in the PR#249.

I was trying to honour that expression of interest.

Folks are under no obligation to pay attention to, engage with or undertake my feedback. If you have a pressing need to merge this pull request, then do so and either defer or disregard my thoughts.

I understand that sometimes folks are under pressure to get things done. Alas, that can sometimes create tension during design conversations :-/ 

I apologise if my feedback was inappropriate, ill timed or not useful. Instead I should have responded during the initial process (which is challenging given the agreed development process and the breadth of work being undertaken).

FWIW, I think that an empty JSON object, whilst compliant (because our response schema are often vague), is not in the spirit of the schema you shared with me. Given that existing design, I think it a more appropriate that all 200 (or 201 if folks decide not to change it) responses should express the `application/json` content type and provide a body with a JSON object that always has the `nonUpdatedIds` property.

cc: @craigmcnally @zburke

NikitaSedyx:
I agree it can always contain `nonUpdatedIds`, but it doesn't introduce big difference for me - empty object or object with empty `nonUpdatedIds` in case everything is fine

the main point JSON should be sent in response

marcjohnson-kint:
> I agree it can always contain `nonUpdatedIds`, but it doesn't introduce big difference for me - empty object or object with empty `nonUpdatedIds` in case everything is fine
> 
> the main point JSON should be sent in response

Then we can agree to disagree and you should disregard my feedback. To me, including an empty JSON object is making more work for the client rather than appeasing it.

I won't comment further on this pull request, as it is clear that my feedback is a distraction or unwanted.

khandramai:
@marcjohnson-kint thank you for your interest to this PR. Your idea

> Given that existing design, I think it a more appropriate that all 200 (or 201 if folks decide not to change it) responses should express the application/json content type and provide a body with a JSON object that always has the nonUpdatedIds property.

LGTM. Taking into account that UI client can accept response with empty `nonUpdatedIds` and this can't affect current logic I think to improve API with these changes.

@marcjohnson-kint @NikitaSedyx thank you for valuable discussion!

NikitaSedyx:
> To me, including an empty JSON object is making more work for the client rather than appeasing it
actually it doesn't, that's why I commented before that there is not big difference

so if @khandramai is ok with it, so I'm fine too

@marc-johnson thank you for your input! empty `nonUpdatedIds` is valuable note

khandramai:
@folio-org/inventory-backend-code-owners ,
could you please assist in review? Thank you!

src/main/java/org/folio/inventory/support/http/server/JsonResponse.java
```diff
@@ -30,14 +31,14 @@ public static void success(HttpServerResponse response,
...
+  public static void successWithEmptyIds(HttpServerResponse response) {
+    successWithIds(response, new ArrayList<>());
   }
 
   public static void successWithIds(HttpServerResponse response, List<String> ids) {
     JsonObject nonUpdatedIds = new JsonObject();
     nonUpdatedIds.put("nonUpdatedIds", ids);
-    response(response, nonUpdatedIds, 201);
+    response(response, nonUpdatedIds, 200);
   }
```
bohdan-suprun, line 42: Should we move these methods to some separate class or to MoveResource? As far as I can see, this `JsonResponse` contains common responses, that can be reused, but `successWithIds` is a specific API response.

ramls/inventory-move.raml
```diff
@@ -26,7 +26,7 @@ traits:
       application/json:
         type: items_move
     responses:
-      201:
+      200:
```
bohdan-suprun, line 29: Please update API versions.

khandramai: Current version is v0.1. What is expecting version - 1.0 or 0.2?

bohdan-suprun: I think `0.2`. It is a breaking change I assume, but API is still experimental (major version is `0`) so we need to update only minor version.

bohdan-suprun:
Hi @marcjohnson-kint, @khandramai,  I see there was some  conversation, is it something that you've agreed? 

Use 200 code and always return a json response, even for success?

marcjohnson-kint:
> Hi @marcjohnson-kint, @khandramai, I see there was some conversation, is it something that you've agreed?
> 
> Use 200 code and always return a json response, even for success?

@bohdan-suprun I'm not sure I understand your question. Are you asking if 200 / 201 are appropriate status codes even when some items are not moved?

The extent of the conversation yesterday was whether 200 or 201 was a more appropriate status code. I suggested that 201 was not appropriate as no new resources (or records) were being created as part of this process.

The API design was not discussed here in a more general sense.

khandramai:
@folio-org/inventory-backend-code-owners , last remarks were applied. Could you please review?

bohdan-suprun:
> > Hi @marcjohnson-kint, @khandramai, I see there was some conversation, is it something that you've agreed?
> > Use 200 code and always return a json response, even for success?
> 
> @bohdan-suprun I'm not sure I understand your question. Are you asking if 200 / 201 are appropriate status codes even when some items are not moved?
> 
> The extent of the conversation yesterday was whether 200 or 201 was a more appropriate status code. I suggested that 201 was not appropriate as no new resources (or records) were being created as part of this process.
> 
> The API design was not discussed here in a more general sense.

I was asking if anything else was discussed except the status and returning json

zburke:
> FWIW, I think that an empty JSON object, whilst compliant (because our response schema are often vague), is not in the spirit of the schema you shared with me. Given that existing design, I think it a more appropriate that all 200 (or 201 if folks decide not to change it) responses should express the `application/json` content type and provide a body with a JSON object that always has the `nonUpdatedIds` property.

If a response says it's JSON, the UI expects it to contain JSON. Hopefully this part is not controversial. Thus removing 
```
JsonResponse.successWithEmptyBody(response);
```
is a straight up bugfix 💯 . Good! 

I don't have strong feelings about the shape of the response as long as it matches the documentation. If I understand correctly, this PR now sends a 200 with the shape `{ "nonUpdatedIds": [] }`, so the shape matches but now the response-code doesn't. I agree with @marcjohnson-kint that since no resources being created, 201 is incorrect. So, this PR looks good to me, but now we need a new ticket to fix the documentation. Is that correct?

khandramai:
@zburke if you meant API documentation then I have already fixed raml-file. Documentation already contained response schema6 for example, https://s3.amazonaws.com/foliodocs/api/mod-inventory/inventory-move.html#inventory_items_move_post. Please, don't pay attention on http status. This will be fixed after current PR merging.

PR #262, opened 2020-09-05
MODSOURMAN-281 - Added post-processing handler for record modified event

RuslanLavrov:
should be merged after folio-org/data-import-processing-core#91

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
@@ -6,6 +6,7 @@ import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
 import org.folio.ActionProfile;
 import org.folio.DataImportEventPayload;
+import org.folio.DataImportEventTypes;
 import org.folio.inventory.common.Context;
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
```diff
@@ -105,6 +105,16 @@ public boolean isEligible(DataImportEventPayload dataImportEventPayload) {
...
   }
 
+  @Override
+  public boolean isPostProcessingNeeded() {
+    return true;
+  }
+
+  @Override
+  public String getPostProcessingInitializationEventType() {
+    return "DI_INVENTORY_INSTANCE_CREATED_READY_FOR_POST_PROCESSING";
```
VRohach, line 115: I think it will be better to use enum here, not simple String.

VRohach (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
RuslanLavrov: sure, I need to merge processing-core PR to be able to use event types enum.

KaterynaSenchenko (approved):

OleksiiKuzminov (approved):

PR #264, opened 2020-09-18
MODDICORE-58: MARC-MARC matching for 001 and 999 ff fields

VRohach:
1. mod-source-record-storage-client version bumped up.


RuslanLavrov (approved):

KaterynaSenchenko (approved):

adamdickmeiss (changes requested):
Doesn't compile. Probably due to using SNAPSHOTs which is a bad idea.

PR #266, opened 2020-09-28
Propagate storage data elements to support Inventory as shared index

nielserik:
- propagate Instance.matchKey
  - propagate Item.status enumeration option "Unknown"
  - requires new minor versions of instance-storage, item-storage

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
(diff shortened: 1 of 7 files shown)


nielserik:
This purpose of this PR is only to propagate new data elements from Inventory Storage as per pull request https://github.com/folio-org/mod-inventory-storage/pull/512  and approval thus only makes sense if the PR in mod-inventory-storage is approved. 

Please see the comments on PR 512 in mod-inventory-storage regarding the motivation for this PR.

PR #267, opened 2020-09-29
MODINV-296 - Field mappings: Item - Item status handling logic

RuslanLavrov:
ticket: https://issues.folio.org/browse/MODINV-296

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
@@ -37,6 +37,7 @@ import static java.util.Objects.isNull;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.folio.ActionProfile.Action.UPDATE;
 import static org.folio.DataImportEventTypes.DI_INVENTORY_ITEM_UPDATED;
+import static org.folio.inventory.domain.items.Item.STATUS_KEY;
 import static org.folio.rest.jaxrs.model.EntityType.ITEM;
 import static org.folio.rest.jaxrs.model.EntityType.MARC_BIBLIOGRAPHIC;
 import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ContentType.ACTION_PROFILE;
@@ -73,7 +74,7 @@ public class UpdateItemEventHandler implements EventHandler {
         return future;
       }
 
-      String oldItemStatus = new JsonObject(payloadContext.get(ITEM.value())).getJsonObject("status").getString("name");
+      String oldItemStatus = new JsonObject(payloadContext.get(ITEM.value())).getJsonObject(STATUS_KEY).getString("name");
       preparePayloadForMappingManager(dataImportEventPayload);
       MappingManager.map(dataImportEventPayload);
       JsonObject itemAsJson = new JsonObject(payloadContext.get(ITEM.value()));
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
```diff
@@ -44,8 +45,12 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(UpdateItemEventHandler.class);
 
+  public static final String ERROR_MSG_KEY = "ERROR_MSG";
+  public static final String FAILED_EVENT_KEY = "FAILED_EVENT";
   private static final String PAYLOAD_HAS_NO_DATA_MSG = "Failed to handle event payload, cause event payload context does not contain MARC_BIBLIOGRAPHIC data or ITEM to update";
+  private static final String STATUS_UPDATE_ERROR_MSG = "Could not change item status '%s' to '%s'";
   private static final String ITEM_PATH_FIELD = "item";
+  private static final String[] protectedStatusesFromUpdate = new String[]{"Aged to lost", "Awaiting delivery", "Awaiting pickup", "Checked out", "Claimed returned", "Declared lost", "Paged", "Recently returned"};
```
bohdan-suprun, line 53: Is it possible to use `ItemStatus` enum values?

src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
```diff
@@ -68,6 +73,7 @@ public UpdateItemEventHandler(Storage storage) {
         return future;
       }
 
+      String oldItemStatus = new JsonObject(payloadContext.get(ITEM.value())).getJsonObject("status").getString("name");
       preparePayloadForMappingManager(dataImportEventPayload);
       MappingManager.map(dataImportEventPayload);
       JsonObject itemAsJson = new JsonObject(payloadContext.get(ITEM.value()));
```
bohdan-suprun, line 79:
```suggestion
      JsonObject itemAsJson = new JsonObject(payloadContext.get(ITEM.value()));
      
      String oldItemStatus = itemAsJson.getJsonObject("status").getString("name");
      
      preparePayloadForMappingManager(dataImportEventPayload);
      MappingManager.map(dataImportEventPayload);
```

src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
bohdan-suprun, line 53: Can you please capitalize the name of the constant and use `Set` instead of array, as it is more efficient for `contains` checks.

RuslanLavrov: yes agree, thank you

RuslanLavrov: do you mean one of the existing enums?

bohdan-suprun: Yes, there is an enum with all the possible statuses, just think if we can use enum constants instead of raw string.

RuslanLavrov: got it, but here we only need some statuses which are disallowed being updated. The existing enum is used for subsequent status validation after mapping.

KaterynaSenchenko (approved):

OleksiiKuzminov (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
```diff
@@ -68,29 +75,44 @@ public UpdateItemEventHandler(Storage storage) {
...
         future.completeExceptionally(new EventProcessingException(msg));
         return future;
       }
 
+      String newItemStatus = mappedItemAsJson.getJsonObject(STATUS_KEY).getString("name");
+      boolean statusWasUpdated = !oldItemStatus.equals(newItemStatus);
+      boolean isOldStatusProtected = PROTECTED_STATUSES_FROM_UPDATE.contains(oldItemStatus);
+      if(statusWasUpdated && isOldStatusProtected) {
+        mappedItemAsJson.getJsonObject(STATUS_KEY).put("name", oldItemStatus);
+      }
```
VRohach, line 98: I think it will be better to extract this logic(regarding Item`s statuses) to the separate method.

VRohach (approved):

PR #269, opened 2020-10-02
MODINV-335: Migrate to java 11.

bohdan-suprun:
https://issues.folio.org/browse/MODINV-335 - Migrate to java 11.


bohdan-suprun:
Merging without approval because CF team has limited capacity this week and this change blocks folijet team.

PR #273, opened 2020-10-09
Update SRS-client dependency to v4.1.0 and di-core to v2.2.0

KaterynaSenchenko:


KaterynaSenchenko:
data-import-processing-core v2.2.0 and mod-source-record-storage-client v4.1.0 are released - merging this PR without approves in order to prevent build failures as current SNAPSHOT versions are no longer available

PR #279, opened 2020-11-05
MODINV-355: Marking item missing clears check in note and check out note fields

mrdpopov:
## Purpose

This fixes the issue: when item is marked missing, its circulation notes were not populated.

Jira: https://issues.folio.org/browse/MODINV-355


bohdan-suprun (approved):
Good catch!

PR #280, opened 2020-11-08
MODINV-357 - Can't move item between holdings on a single instance (some items)

khandramai:
The new field sourceId introduced in the https://github.com/folio-org/mod-inventory-storage/pull/502 for HoldingsRecord (mod-inventory-storage) schema isn't reflected in the corresponding schema in the mod-inventory. As a result exception happens in the Move API.

Approach -> schema for HoldingsRecord was updated.


src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleCollection.java
```diff
@@ -94,10 +94,14 @@ public void findById(String id,
           case 200:
             JsonObject instanceFromServer = new JsonObject(responseBody);
 
-            T found = mapFromJson(instanceFromServer);
-
-            resultCallback.accept(new Success<>(found));
-            break;
+            try {
```
khandramai, line 97: @marc-johnson @bohdan-suprun I have introduced this code block to protect flow from uncompleted future when mapping json -> object failed due to schemas mismatching. Is it make sense to do this for another operations like add, findAll? This is not relevant to the current bug but in theory could be useful in other cases.

khandramai: Let me please know if you think this is overkill. Then I'll just leave the schema change and that will solve current issue.

khandramai: Let me please know if you think this is overkill. Then I'll just leave the schema change and that will fix current issue.

marcjohnson-kint: I don't think this is overkill. It seems like a good idea to trap these kinds of errors and report them accordingly.

By all means do it for other operations. I don't know how likely this will be triggered for different record types due to how the mapping occurs.

siarhei-charniak (approved):

bohdan-suprun (approved):

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleCollection.java
khandramai: As discussed with @bohdan-suprun I have protected only getById operation for now. The other cases are the matter of tech debt and aren't connected with this particular case. So I'm going to merge this PR as is.

bohdan-suprun: Tech debt is created https://issues.folio.org/browse/MODINV-359

PR #281, opened 2020-11-10
Support both "http.port" and "port" property variables

julianladisch:
"http.port" is the command-line option name that all RMB based
modules support:
https://github.com/folio-org/raml-module-builder#command-line-options


adamdickmeiss (commented):
Why support both??

marcjohnson-kint:
Please could documentation be added to explain that both properties are supported (and state the precedence)

hjiebsco (approved):

PR #291, opened 2020-12-14
MODINV-356 : Implement API to move items into the new statuses.

AnShapovalov:
MODINV-356 : Implement API to move items into the new statuses.

PR #292, opened 2020-12-14
MODINV-369: Item with Lost and paid status throws an error when withdrawn

K-Felk:
The module has a "whitelist" of statuses from which an item can be moved
to withdrawn.  "Lost and Paid" was not on that list, hence the error.  I
just added the status to the list.

refs: https://issues.folio.org/browse/MODINV-369


marcjohnson-kint (approved):
I would ask for a test to be written for this, however there are discussions being had on #291 about how testing these endpoints could be changed.

K-Felk:
Ah, thank you for mentioning this-I figured there was a test already written for this, I had meant to amend it to include the new status.  I have now done so.

PR #294, opened 2021-01-12
MODINV-356 Extract validators

AnShapovalov:


src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
```diff
@@ -0,0 +1,81 @@
...
+    ItemStatusName targetItemStatusName = ItemStatusName.IN_PROCESS;
+    final AbstractTargetValidator statusValidator = validator.getValidator(targetItemStatusName);
+
+    Set<ItemStatusName> allowedStatuses = statusValidator.getAllStatusesAllowedToMark();
+    System.out.println("Allowed statuses for:"+statusValidator.getItemStatusName());
+    allowedStatuses.stream().forEach(x -> {
+      System.out.println("\t"+x);
+      Item item = new Item(null, null, new Status(x), null, null, null);
+      assertTrue(statusValidator.isItemAllowedToMark(item));
+    });
```
marcjohnson-kint, line 37: This code seems to check that the statuses that production code thinks should be allowed are allowed by the validator.

Have I understood this correctly?

marcjohnson-kint (commented):
Thank you for removing the duplication, I think the definition of the validators is clearer and easier to understand now.

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
AnShapovalov: Sure

src/main/java/org/folio/inventory/resources/ItemStatusURL.java
```diff
@@ -0,0 +1,37 @@
...
+
+    itemStatusNameUrlMap.put(ItemStatusName.IN_PROCESS, "/mark-in-process");
+    itemStatusNameUrlMap.put(ItemStatusName.IN_PROCESS_NON_REQUESTABLE, "/mark-in-process-non-requestable");
+//    itemStatusNameUrlMap.put(ItemStatusName.INTELLECTUAL_ITEM, "/mark-intellectual-item");
+//    itemStatusNameUrlMap.put(ItemStatusName.LONG_MISSING, "/mark-long-missing");
+//    itemStatusNameUrlMap.put(ItemStatusName.MISSING, "/mark-missing");
+//    itemStatusNameUrlMap.put(ItemStatusName.RESTRICTED, "/mark-restricted");
+//    itemStatusNameUrlMap.put(ItemStatusName.UNAVAILABLE, "/mark-unavailable");
+//    itemStatusNameUrlMap.put(ItemStatusName.UNKNOWN, "/mark-unknown");
+//    itemStatusNameUrlMap.put(ItemStatusName.WITHDRAWN, "/mark-withdrawn");
```
marcjohnson-kint, line 25: As this pull request is intended to only cover the in process statuses, please remove this commented out code until those other statuses are implemented.

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -93,10 +95,15 @@ public void register(Router router) {
     router.put(RELATIVE_ITEMS_PATH_ID).handler(this::update);
     router.delete(RELATIVE_ITEMS_PATH_ID).handler(this::deleteById);
 
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process")
-      .handler(handle(this::markAsInProcess));
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process-non-requestable")
-      .handler(handle(this::markAsInProcessNonRequestable));
+    Arrays.stream(ItemStatusName.values())
+      .map(x->ItemStatusURL.getUrlForItemStatusName(x))
```
marcjohnson-kint, line 99: I find it easier to understand expressions like this when the parameters have meaningful names. Alternatively, use a method reference instead.

Similarly for the optional value presence filter below this.

```suggestion
      .map(statusName -> ItemStatusURL.getUrlForItemStatusName(statusName))
```

```suggestion
      .map(ItemStatusURL::getUrlForItemStatusName)
```

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -93,10 +95,15 @@ public void register(Router router) {
...
+      .filter(x->x.isPresent())
+      .forEach(x->router.post(RELATIVE_ITEMS_PATH_ID+x.get()).handler(handle(this::markItemAsTargetStatus)));
+
+//    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process")
+//      .handler(handle(this::markAs));
+//    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process-non-requestable")
+//      .handler(handle(this::markAs));
```
marcjohnson-kint, line 106: As this code has been replaced by the above, please remove this commented out code. If it is needed in the future, it can be recovered from source control.

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -93,10 +95,15 @@ public void register(Router router) {
...
     router.delete(RELATIVE_ITEMS_PATH_ID).handler(this::deleteById);
 
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process")
-      .handler(handle(this::markAsInProcess));
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process-non-requestable")
-      .handler(handle(this::markAsInProcessNonRequestable));
+    Arrays.stream(ItemStatusName.values())
+      .map(x->ItemStatusURL.getUrlForItemStatusName(x))
+      .filter(x->x.isPresent())
+      .forEach(x->router.post(RELATIVE_ITEMS_PATH_ID+x.get()).handler(handle(this::markItemAsTargetStatus)));
```
marcjohnson-kint, line 101: I think it would be easier to understand this code if it was extracted to a method with an intent revealing name. Maybe something like:

```suggestion
      .forEach(itemStatusName -> registerHandlerForChangingItemStatus(itemStatusName, router);
```

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -114,6 +121,22 @@ public void register(Router router) {
 
   }
 
+  private CompletableFuture<Void> markItemAsTargetStatus(
+    RoutingContext routingContext, WebContext webContext, Clients clients) {
+    Optional<ItemStatusName> itemStatusName = ItemStatusURL.getItemStatusNameForUrl(routingContext.request().uri());
```
marcjohnson-kint, line 126: I find it easier to distinguish the first line of a method after a multi-line signature if there is a whitespace gap between them

```suggestion
  private CompletableFuture<Void> markItemAsTargetStatus(RoutingContext routingContext,
    WebContext webContext, Clients clients) {
    
    final var itemStatusName = ItemStatusURL.getItemStatusNameForUrl(routingContext.request().uri());
```

src/main/java/org/folio/inventory/services/MoveItemIntoStatusService.java
```diff
@@ -57,14 +61,38 @@ public MoveItemIntoStatusService(ItemCollection itemCollection, Clients clients)
       .thenCompose(itemCollection::update);
   }
 
+  public CompletableFuture<Item> processMarkItem(WebContext context,ItemStatusName statusName) {
```
marcjohnson-kint, line 64: Given that we have settled on the language of `mark` to refer to a status transition rather than `move`, I think it might make sense to rename this method (and class).

What do you think to `markItemStatusAs`?

```suggestion
  public CompletableFuture<Item> markItemStatusAs(WebContext context, ItemStatusName statusName) {
```

src/main/java/org/folio/inventory/services/MoveItemIntoStatusService.java
```diff
@@ -57,14 +61,38 @@ public MoveItemIntoStatusService(ItemCollection itemCollection, Clients clients)
...
+//  public CompletableFuture<Item> processMarkItemInProcess(WebContext context) {
+//    final String itemId = context.getStringParameter("id", null);
+//
+//    return itemCollection.findById(itemId)
+//      .thenCompose(ItemsValidator::refuseWhenItemNotFound)
+//      .thenCompose(MarkAsInProcessValidators::itemHasAllowedStatusToMarkAsInProcess)
+//      .thenCompose(this::updateRequestStatusIfRequired)
+//      .thenApply(item -> item.changeStatus(IN_PROCESS))
+//      .thenCompose(itemCollection::update);
+//  }
```
marcjohnson-kint, line 85: Please remove this commented out code. If it is needed in future it can be retrieved from version control.

src/main/java/org/folio/inventory/validation/experimental/AbstractTargetValidator.java
```diff
@@ -0,0 +1,44 @@
...
+import org.folio.inventory.support.http.server.ValidationError;
+
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+
+import static org.folio.inventory.support.CompletableFutures.failedFuture;
+
+public abstract class AbstractTargetValidator {
+  private ItemStatusName itemStatusName;
+  private Set<ItemStatusName> allowedStatusToMark;
```
marcjohnson-kint, line 15: I believe both of these can be made final, as they are initialised in the constructor and not changed.

```suggestion
  private final ItemStatusName itemStatusName;
  private final Set<ItemStatusName> allowedStatusToMark;
```

src/main/java/org/folio/inventory/validation/experimental/AbstractTargetValidator.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  public Set<ItemStatusName> getAllStatusesAllowedToMark() {
+    return allowedStatusToMark;
+  }
+
+  public ItemStatusName getItemStatusName() {
+    return itemStatusName;
+  }
+
```
marcjohnson-kint, line 43: Please remove this redundant blank line, I don't think it improves the readability of the code

src/main/java/org/folio/inventory/validation/experimental/InProcessNonRequestableTargetValidator.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.folio.inventory.domain.items.Item;
+import org.folio.inventory.domain.items.ItemStatusName;
+
+import java.util.Set;
+
+import static java.util.Set.of;
+
+public class InProcessNonRequestableTargetValidator extends AbstractTargetValidator {
+
```
marcjohnson-kint, line 11: Please remove this redundant blank line, I don't think it improves the readability of the code

src/main/java/org/folio/inventory/validation/experimental/InProcessNonRequestableTargetValidator.java
```diff
@@ -0,0 +1,26 @@
...
+      ItemStatusName.AWAITING_PICKUP,
+      ItemStatusName.IN_TRANSIT,
+      ItemStatusName.LOST_AND_PAID,
+      ItemStatusName.MISSING,
+      ItemStatusName.ORDER_CLOSED,
+      ItemStatusName.PAGED,
+      ItemStatusName.WITHDRAWN
+    ));
+  }
+
```
marcjohnson-kint, line 25: Please remove this redundant blank line, I don't think it improves the readability of the code

src/main/java/org/folio/inventory/validation/experimental/InProcessTargetValidator.java
```diff
@@ -0,0 +1,23 @@
+package org.folio.inventory.validation.experimental;
+
+import org.folio.inventory.domain.items.ItemStatusName;
+
+import static java.util.Set.of;
+
+public class InProcessTargetValidator extends AbstractTargetValidator {
+
```
marcjohnson-kint, line 8: Please remove this redundant blank line, I don't think it improves the readability of the code

src/main/java/org/folio/inventory/validation/experimental/InProcessTargetValidator.java
```diff
@@ -0,0 +1,23 @@
...
+      ItemStatusName.AWAITING_PICKUP,
+      ItemStatusName.IN_TRANSIT,
+      ItemStatusName.LOST_AND_PAID,
+      ItemStatusName.MISSING,
+      ItemStatusName.ORDER_CLOSED,
+      ItemStatusName.PAGED,
+      ItemStatusName.WITHDRAWN
+    ));
+  }
+
```
marcjohnson-kint, line 22: Please remove this redundant blank line, I don't think it improves the readability of the code

src/test/java/org/folio/inventory/resources/ItemStatusURLTest.java
```diff
@@ -0,0 +1,19 @@
...
+  @Test
+  public void canGetItemStatusByFullUrl() {
+    assertEquals(ItemStatusName.IN_PROCESS, ItemStatusURL.getItemStatusNameForUrl("http://dummy.net/239853/mark-in-process").get());
+  }
+
+  @Test
+  public void canGetUrlByItemStatus() {
+    assertEquals("/mark-in-process", ItemStatusURL.getUrlForItemStatusName(ItemStatusName.IN_PROCESS).get());
+  }
+
```
marcjohnson-kint, line 18: Please remove this redundant blank line, I don't think it improves the readability of the code

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
marcjohnson-kint: Apologies for the delayed response, I didn't realise you had replied to this comment.

How does it test the workflow?

From my perspective, this test only checks that the production code is consistent with itself. 

I think tests should demonstrate the intended behaviour rather than what the code says it does. As an example, if a status was incorrectly removed from the production code, this test would still pass. It does not raise awareness of the impact of a change or that something might have been incorrectly changed. 

Does that make sense?

I also think that tests should act as documentation for the behaviour of the system. When I read this test, it tells me that the validator will allow a status change when the source status is in the set defined by the system. I would find it more useful for understanding the intended behaviour if it described which of those statuses were allowed.

What do you think?

Below is an example of a test that I think better describes the intended behaviour of the system:
```
  @SneakyThrows
  @Parameters({
    "Available",
    "In transit",
    "Awaiting pickup",
    "Missing",
    "Withdrawn",
    "Lost and paid",
    "Paged",
    "Awaiting delivery",
    "Order closed"
  })
  @Test
  public void itemCanBeMarkedAsInProcessWhenInAcceptableSourceStatus(String sourceStatus) {
    final var validator = getValidator(IN_PROCESS);

    final var item = itemInStatus(forName(sourceStatus));

    final var validationFuture = validator.itemHasAllowedStatusToMark(item);

    validationFuture.get(1, TimeUnit.SECONDS);

    // Validator responds with a successful future when valid
    assertThat(validationFuture.isDone(), is(true));
  }
```

src/main/java/org/folio/inventory/resources/ItemStatusURL.java
AnShapovalov: I have been planned to implement other statuses after aproving of "comon idea" within this PR. This commented code is my "roadmap". 

2

marcjohnson-kint: Ok, if that is going to be done in this pull request, then there is no need to remove it

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
AnShapovalov: Marc, you are right. I thoght more about main code structure than about tests. I think it would be great way if we could to take Item statuses and Transitions some where else where current approved reqiremnts would stored. But right now Statuses and allowed Transitions statically saved in test is a single way despite it time consuming and weak informativity (i mean you should review all tests and validators files to have a full picture of workflow). Also. I think,  I can mock MoveItemToStatusService class and make tests on it. Does it make sence?

marcjohnson-kint: >  I think it would be great way if we could to take Item statuses and Transitions some where else where current approved requirements would stored.

I'm not sure I understand what you mean. Please could you explain this more.

> right now Statuses and allowed Transitions statically saved in test is a single way despite it time consuming and weak informativity

What part of this do you consider time consuming?

What do you mean by `weak informativity`?

> I think, I can mock MoveItemToStatusService class and make tests on it. Does it make sense?

I'm ok with that if you want to. If there is behaviour in that class that the API tests do not cover then please do.

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -93,10 +96,10 @@ public void register(Router router) {
     router.put(RELATIVE_ITEMS_PATH_ID).handler(this::update);
     router.delete(RELATIVE_ITEMS_PATH_ID).handler(this::deleteById);
 
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process")
-      .handler(handle(this::markAsInProcess));
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process-non-requestable")
-      .handler(handle(this::markAsInProcessNonRequestable));
+    Arrays.stream(ItemStatusName.values()).
+      filter(itemStatusName -> ItemStatusURL.getUrlForItemStatusName(itemStatusName).isPresent())
+      .forEach(itemStatusName -> registerHandlerForChangingItemStatus(itemStatusName,router));
```
marcjohnson-kint, line 101: Why has the structure of this changed so that it is necessary to call `ItemStatusURL.getUrlFor` twice?

Personally, I think this is less clear than the transformation, filter then register. Did my suggestions cause this confusion?

AnShapovalov: Is it will be OK?
    Arrays.stream(ItemStatusName.values()).map(itemStatusName -> ItemStatusURL.getUrlForItemStatusName(itemStatusName))
      .filter(itemStatusUrl -> itemStatusUrl.isPresent())
      .forEach(itemStatusUrl -> registerHandlerForChangingItemStatusUrl(itemStatusUrl.get(),router));

  private void registerHandlerForChangingItemStatusUrl(String itemStatusUrl, Router router) {
    router.post(RELATIVE_ITEMS_PATH_ID + itemStatusUrl)
      .handler(handle(this::markItemAsTargetStatus));
  }

or may be so:
registerHandlerForItemStatuses(router);

private void registerHandlerForItemStatuses(Router router) {
    Arrays.stream(ItemStatusName.values())
      .map(itemStatusName -> ItemStatusURL.getUrlForItemStatusName(itemStatusName))
      .filter(itemStatusUrl -> itemStatusUrl.isPresent())
      .forEach(itemStatusUrl -> router.post(RELATIVE_ITEMS_PATH_ID + itemStatusUrl)
        .handler(handle(this::markItemAsTargetStatus)));
  }

marcjohnson-kint: Yes, I think both are ok. 

I prefer the first option. Having thought more about the language used in this area, I wonder if a better name for the last method might be `registerMarkAsHandler`. What do you think.

If you choose to take the second option, then the `.handler` needs to be indented because it is a continuation of the `router.post` line rather than part of the stream processing chain.

AnShapovalov: RegisterItemStatusTransitionUrlAndHandler - is it precision description of action. May be you, as native speaker, could  interpret it somehow better.

marcjohnson-kint: I think the name should be consistent with the domain language we have chosen, which is to precede each status with `mark as`. 

I think that means using a name like `registerMarkItemAsHandler`. This gives it a degree of symmetry with the handler method name `markItemAsTargetStatus`. If `registerMarkItemAsTargetStatus` works better for you, I'm ok with that too.

AnShapovalov: let's finish on "registerMarkItemAsHandler".

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -93,12 +95,12 @@ public void register(Router router) {
...
 
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process")
-      .handler(handle(this::markAsInProcess));
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-in-process-non-requestable")
-      .handler(handle(this::markAsInProcessNonRequestable));
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-intellectual-item")
-      .handler(handle(this::markAsIntellectualItem));
+    Arrays.stream(ItemStatusName.values()).map(itemStatusName -> ItemStatusURL.getUrlForItemStatusName(itemStatusName))
+      .filter(itemStatusUrl -> itemStatusUrl.isPresent())
+      .forEach(itemStatusUrl -> registerMarkItemAsHandler(itemStatusUrl.get(),router));
```
marcjohnson-kint, line 100: FOLIO's code standards state that their should be a space after the comma separating parameters 

```suggestion
      .forEach(itemStatusUrl -> registerMarkItemAsHandler(itemStatusUrl.get(), router));
```

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -131,7 +152,7 @@ public void register(Router router) {
     final MoveItemIntoStatusService moveItemIntoStatusService = new MoveItemIntoStatusService(storage
       .getItemCollection(webContext), clients);
 
-    return moveItemIntoStatusService.processMarkItemInProcess(webContext)
+    return moveItemIntoStatusService.processMarkItem(webContext, ItemStatusName.IN_PROCESS)
```
marcjohnson-kint, line 155: I think we should name this service consistently with the chosen domain language, `mark as` rather than `move`.

```suggestion
    return markItemStatusService.markItemAs(IN_PROCESS, webContext)
```

src/main/java/org/folio/inventory/validation/experimental/AbstractTargetValidator.java
```diff
@@ -0,0 +1,44 @@
...
+import org.folio.inventory.domain.items.ItemStatusName;
+import org.folio.inventory.exceptions.UnprocessableEntityException;
+import org.folio.inventory.support.http.server.ValidationError;
+
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+
+import static org.folio.inventory.support.CompletableFutures.failedFuture;
+
+public abstract class AbstractTargetValidator {
```
marcjohnson-kint, line 13: As this validator is focused on the item status I think it would help comprehension for this to be included in the name

```suggestion
public abstract class AbstractTargetItemStatusValidator {
```

src/main/java/org/folio/inventory/validation/experimental/IntellectualItemTargetValidator.java
```diff
@@ -0,0 +1,21 @@
+package org.folio.inventory.validation.experimental;
+
+import org.folio.inventory.domain.items.ItemStatusName;
+
+import static java.util.Set.of;
+
+public class IntellectualItemTargetValidator extends AbstractTargetValidator {
```
marcjohnson-kint, line 7:
```suggestion
public class IntellectualItemTargetItemStatusValidator extends AbstractTargetItemStatusValidator {
```

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
```diff
@@ -0,0 +1,93 @@
...
+    validationFuture.get(1, TimeUnit.SECONDS);
+
+    // Validator responds with a successful future when valid
+    assertThat(validationFuture.isDone()).isTrue();
+  }
+
+  @AfterClass
+  public static void tearDown() {
+//    assertFalse(true);
+  }
```
marcjohnson-kint, line 92: Please remove this unused code as it is unnecessary

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
```diff
@@ -0,0 +1,93 @@
...
+    Set<ItemStatusName> disallowedStatuses = new HashSet<>();
+    Collections.addAll(disallowedStatuses,ItemStatusName.values());
+    disallowedStatuses.removeAll(allowedStatuses);
+    System.out.println("Disallowed statuses for:"+statusValidator.getItemStatusName());
+    disallowedStatuses.stream().forEach(x -> {
+      System.out.println("\t"+x);
+      Item item = new Item(null, null, new Status(x), null, null, null);
+      assertThat(statusValidator.isItemAllowedToMark(item)).isFalse();
+    });
+  }
```
marcjohnson-kint, line 61: Given what we talked about with how this test only demonstrates consistency between the production code and itself, why is it still present?

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
```diff
@@ -0,0 +1,93 @@
...
+    "Awaiting pickup",
+    "Missing",
+    "Withdrawn",
+    "Lost and paid",
+    "Paged",
+    "Awaiting delivery",
+    "Order closed"
+  })
+  @Test
+  public void itemCanBeMarkedAsInProcessWhenInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 76: I think there needs to be a corresponding negative test for the source statuses which aren't allowed to be marked as in process.

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.folio.inventory.domain.items.ItemStatusName.IN_PROCESS;
+import static org.folio.inventory.domain.items.ItemStatusName.forName;
+
+@RunWith(JUnitParamsRunner.class)
+public class TargetItemStatusValidatorTest {
```
marcjohnson-kint, line 25: I think it might make sense to have the tests split by target status in order to keep the size of the test classes small (and keep them related to the classes actually doing the work)

```suggestion
public class InProcessTargetItemStatusValidatorTest {
```

src/test/java/org/folio/inventory/validation/experimental/IntellectualItemStatusValidatorTest.java
```diff
@@ -0,0 +1,60 @@
...
+    validationFuture.get(1, TimeUnit.SECONDS);
+
+    // Validator responds with a successful future when valid
+    assertThat(validationFuture.isDone()).isTrue();
+  }
+
+  @AfterClass
+  public static void tearDown() {
+//    assertFalse(true);
+  }
```
marcjohnson-kint, line 59: Please remove this unused code

src/test/java/org/folio/inventory/validation/experimental/IntellectualItemStatusValidatorTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.folio.inventory.domain.items.ItemStatusName.IN_PROCESS;
+
+@RunWith(JUnitParamsRunner.class)
+public class IntellectualItemStatusValidatorTest {
+  private static TargetItemStatusValidator validator;
+
+  @BeforeClass
+  public static void setUp() throws Exception {
+    validator = new TargetItemStatusValidator();
+  }
```
marcjohnson-kint, line 27: Is this field used anymore?

src/test/java/org/folio/inventory/resources/ItemStatusURLTest.java
```diff
@@ -0,0 +1,30 @@
...
+  public void canGetItemStatusByFullUrlAndCanGetUrlByStatusName(String url, String statusName) {
+    String dummyUrl = "http://dummy.net/239853$URL$";
+    final var itemStatus= ItemStatusURL.getItemStatusNameForUrl(dummyUrl.replace("$URL$",url));
+    assertThat(itemStatus.isPresent()).isTrue();
+    ItemStatusName itemStatusName = ItemStatusName.forName(statusName);
+    assertThat(itemStatus.get()).isEqualTo(itemStatusName);
+    var itemStatusUrl = ItemStatusURL.getUrlForItemStatusName(itemStatusName);
+    assertThat(itemStatusUrl.isPresent()).isTrue();
+    assertThat(itemStatusUrl.get()).isEqualTo(url);
+  }
```
marcjohnson-kint, line 29: This appears to be two different scenarios:
* Determining the URL for a given target status
* Determining the intended target status for a given URL

I think splitting them up would help the reader to understand them better and provide better diagnostics when the tests fail.

src/test/java/org/folio/inventory/validation/experimental/IntellectualItemStatusValidatorTest.java
AnShapovalov: Of course. It gives validatators for each test.

src/test/java/org/folio/inventory/resources/ItemStatusURLTest.java
AnShapovalov: May be it have a sense but looking on the other side does it required additional set of parameters which will need additionally manage when workflow will changed. As for me each manually managed block - is a source for mistakes. May be adding the comments would enough?

src/main/java/org/folio/inventory/resources/Items.java
AnShapovalov: That's legacy code it will be deleted at all.

src/test/java/org/folio/inventory/resources/ItemStatusURLTest.java
marcjohnson-kint: A good start could be to separate these things into sections in the code with comments.

I agree that ideally we wouldn't need to maintain another map of statuses to URLs. I think that could be avoided by changing the production code to use explicit configuration.

src/test/java/org/folio/inventory/validation/experimental/IntellectualItemStatusValidatorTest.java
marcjohnson-kint: As far as I could tell, the only test in this class does not use that field.

src/main/java/org/folio/inventory/resources/Items.java
marcjohnson-kint: Then why has it been changed in this pull request if it will be deleted?

Does that mean the `MoveItemIntoStatusService` class is going to be deleted?

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -111,52 +109,39 @@ public void register(Router router) {
...
-          routingContext, webContext));
+  private void registerMarkItemAsHandler(String itemStatusUrl, Router router) {
+    router.post(RELATIVE_ITEMS_PATH_ID + itemStatusUrl)
+      .handler(handle(this::markItemAsTargetStatus));
   }
 
-  private CompletableFuture<Void> markAsInProcess(
+  private CompletableFuture<Void> markItemAsTargetStatus(
     RoutingContext routingContext, WebContext webContext, Clients clients) {
+    final var itemStatusName = ItemStatusURL.getItemStatusNameForUrl(routingContext.request().uri());
```
marcjohnson-kint, line 121: When a method signature spans multiple lines, I find it easier to differentiate between the signature and the body of the method if there is whitespace between them. Please introduce a blank line between the signature and the first line.

src/main/java/org/folio/inventory/validation/experimental/AbstractTargetItemStatusValidator.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+
+import static org.folio.inventory.support.CompletableFutures.failedFuture;
+
+public abstract class AbstractTargetItemStatusValidator {
+  private ItemStatusName itemStatusName;
+  private Set<ItemStatusName> allowedStatusToMark;
+
+  protected AbstractTargetItemStatusValidator(ItemStatusName itemStatusName, Set<ItemStatusName> allowedStatusToMark) {
```
marcjohnson-kint, line 17: As this parameter represents a set of statuses I think it should have a plural name. Given that I'm suggesting it be changed, I suggest we use the term `source` to act as a complimentary term to `target`.

```suggestion
  protected AbstractTargetItemStatusValidator(ItemStatusName itemStatusName, Set<ItemStatusName> allowedSourceStatuses) {
```

src/main/java/org/folio/inventory/validation/experimental/AbstractTargetItemStatusValidator.java
```diff
@@ -0,0 +1,44 @@
...
+    this.itemStatusName = itemStatusName;
+    this.allowedStatusToMark = allowedStatusToMark;
+  }
+
+  public CompletableFuture<Item> itemHasAllowedStatusToMark(Item item) {
```
marcjohnson-kint, line 22: To me, this method refuses items from being marked when they are in an invalid source status. I suggest we change the name to reflect that intent.

```suggestion
  public CompletableFuture<Item> refuseItemWhenNotInAcceptableSourceStatus(Item item) {
```

src/main/java/org/folio/inventory/validation/experimental/TargetItemStatusValidators.java
```diff
@@ -0,0 +1,21 @@
...
+    validators.put(ItemStatusName.IN_PROCESS, new InProcessTargetItemStatusValidator());
+    validators.put(ItemStatusName.IN_PROCESS_NON_REQUESTABLE, new InProcessNonRequestableTargetItemStatusValidator());
+    validators.put(ItemStatusName.INTELLECTUAL_ITEM, new IntellectualItemTargetItemStatusValidator());
+    validators.put(ItemStatusName.LONG_MISSING, new LongMissingTargetItemStatusValidator());
+  }
+
+  public AbstractTargetItemStatusValidator getValidator(ItemStatusName itemStatusName) {
+    return validators.get(itemStatusName);
+  }
+
```
marcjohnson-kint, line 20: Please remove this unnecessary whitespace as it does not enhance the readability of the code

src/test/java/api/items/MarkItemLongMissingApiTests.java
```diff
@@ -105,7 +106,7 @@ public void cannotMarkItemLongMissingWhenNotInAllowedStatus(String initialStatus
       .canCirculate());
 
     assertThat(markItemLongMissing(createdItem), hasValidationError(
-      "Item is not allowed to be marked as Long missing", "status.name", initialStatus));
+      "Item is not allowed to be marked as:\"Long missing\"", "status.name", initialStatus));
```
marcjohnson-kint, line 109: Why has this message changed? Did the new message come from the requirements?

src/test/java/org/folio/inventory/validation/experimental/InProcessItemStatusValidatorTest.java
```diff
@@ -0,0 +1,85 @@
...
+    final var validationFuture = targetValidator.itemHasAllowedStatusToMark(item);
+
+    try {
+      validationFuture.get(1, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
+      e.printStackTrace();
+    } catch (ExecutionException e) {
+      e.printStackTrace();
+    } catch (TimeoutException e) {
+      e.printStackTrace();
```
marcjohnson-kint, line 80: I believe these catch statements without assertions would cause the test to erroneously pass.

Please consider using an [expected exception rule](https://www.baeldung.com/junit-assert-exception) to check the exception.

src/test/java/org/folio/inventory/validation/experimental/InProcessItemStatusValidatorTest.java
```diff
@@ -0,0 +1,85 @@
...
+    "In process",
+    "In process (non-requestable)",
+    "Intellectual item",
+    "Long missing",
+    "On order",
+    "Restricted",
+    "Unavailable",
+    "Unknown"})
+  @Test
+  public void itemCanNotBeMarkedAsInProcessWhenInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 68: This test name suggests that the item cannot be marked as the target status when in an acceptable status. I imagine this is a typo from copying and pasting the other test.

```suggestion
  public void itemCannotBeMarkedAsInProcessWhenInUnacceptableSourceStatus(String sourceStatus) {
```

src/test/java/org/folio/inventory/validation/experimental/InProcessItemStatusValidatorTest.java
```diff
@@ -0,0 +1,85 @@
...
+    "Claimed returned",
+    "Declared lost",
+    "In process",
+    "In process (non-requestable)",
+    "Intellectual item",
+    "Long missing",
+    "On order",
+    "Restricted",
+    "Unavailable",
+    "Unknown"})
```
marcjohnson-kint, line 66: The formatting of the end of these parameters is different to the one above. Please make them consistent. I don't mind which formatting is used.

src/test/java/org/folio/inventory/validation/experimental/InProcessNonRequestableItemStatusValidatorTest.java
```diff
@@ -0,0 +1,89 @@
...
+    "Awaiting pickup",
+    "In transit",
+    "Lost and paid",
+    "Missing",
+    "Order closed",
+    "Paged",
+    "Withdrawn"
+  })
+  @Test
+  public void itemCanBeMarkedAsInProcessNonRequestableWhenInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 44: These tests are about the `non requestable` item status, please change the names accordingly. Please check the other test classes as well.

```suggestion
  public void itemCanBeMarkedAsNonRequestableWhenInAcceptableSourceStatus(String sourceStatus) {
```

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorsTest.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.folio.inventory.domain.items.ItemStatusName.IN_PROCESS;
+import static org.folio.inventory.domain.items.ItemStatusName.forName;
+
+@RunWith(JUnitParamsRunner.class)
+public class TargetItemStatusValidatorsTest {
```
marcjohnson-kint, line 25: There is a new class specifically for the `In process` target status. Is this class still needed?

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -93,16 +95,12 @@ public void register(Router router) {
...
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-long-missing")
-      .handler(handle(this::markAsLongMissing));
-    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-missing")
-      .handler(handle(this::markAsMissing));
+    Arrays.stream(ItemStatusName.values()).map(itemStatusName -> ItemStatusURL.getUrlForItemStatusName(itemStatusName))
+      .filter(itemStatusUrl -> itemStatusUrl.isPresent())
+      .forEach(itemStatusUrl -> registerMarkItemAsHandler(itemStatusUrl.get(), router));
+
+//    router.post(RELATIVE_ITEMS_PATH_ID + "/mark-missing")
+//      .handler(handle(this::markAsMissing));
```
marcjohnson-kint, line 103: If this is now implemented in the new way, please remove the code rather than commenting it out

src/test/java/org/folio/inventory/validation/experimental/InProcessNonRequestableItemStatusValidatorTest.java
```diff
@@ -0,0 +1,81 @@
...
+    "Awaiting pickup",
+    "In transit",
+    "Lost and paid",
+    "Missing",
+    "Order closed",
+    "Paged",
+    "Withdrawn"
+  })
+  @Test
+  public void itemCanBeMarkedAsInProcessNonRequestableWhenNotInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 42:
```suggestion
  public void itemCanBeMarkedAsInProcessNonRequestableWhenInAcceptableSourceStatus(String sourceStatus) {
```

src/test/java/org/folio/inventory/validation/experimental/InProcessNonRequestableItemStatusValidatorTest.java
```diff
@@ -0,0 +1,81 @@
...
+    "In process (non-requestable)",
+    "Intellectual item",
+    "Long missing",
+    "On order",
+    "Restricted",
+    "Unavailable",
+    "Unknown"
+  })
+  @Test
+  public void itemCanNotBeMarkedAsInProcessNonRequestableWhenInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 70:
```suggestion
  public void itemCannotBeMarkedAsInProcessNonRequestableWhenNotInAcceptableSourceStatus(String sourceStatus) {
```

src/test/java/org/folio/inventory/validation/experimental/IntellectualItemStatusValidatorTest.java
```diff
@@ -0,0 +1,81 @@
...
+    "In process (non-requestable)",
+    "Intellectual item",
+    "Long missing",
+    "On order",
+    "Restricted",
+    "Unavailable",
+    "Unknown"
+  })
+  @Test
+  public void itemCanNotBeMarkedAsAsIntellectualItemWhenNotInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 70:
```suggestion
  public void itemCannotBeMarkedAsAsIntellectualItemWhenNotInAcceptableSourceStatus(String sourceStatus) {
```

src/test/java/org/folio/inventory/validation/experimental/LongMissingItemStatusValidatorTest.java
```diff
@@ -0,0 +1,80 @@
...
+    "In process (non-requestable)",
+    "Intellectual item",
+    "Long missing",
+    "On order",
+    "Restricted",
+    "Unavailable",
+    "Unknown"
+  })
+  @Test
+  public void itemCanNotBeMarkedAsLongMissingWhenNotInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 69:
```suggestion
  public void itemCannotBeMarkedAsLongMissingWhenNotInAcceptableSourceStatus(String sourceStatus) {
```

src/test/java/org/folio/inventory/validation/experimental/MissingItemStatusValidatorTest.java
```diff
@@ -0,0 +1,82 @@
...
+    "Lost and paid",
+    "Missing",
+    "On order",
+    "Order closed",
+    "Restricted",
+    "Unavailable",
+    "Unknown"
+  })
+  @Test
+  public void itemCanNotBeMarkedAsMissingWhenNotInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 71:
```suggestion
  public void itemCannotBeMarkedAsMissingWhenNotInAcceptableSourceStatus(String sourceStatus) {
```

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorsTest.java
marcjohnson-kint: This comment was marked as resolved, how has it been resolved?

src/test/java/org/folio/inventory/validation/experimental/InProcessNonRequestableItemStatusValidatorTest.java
```diff
@@ -0,0 +1,81 @@
...
+    "In process (non-requestable)",
+    "Intellectual item",
+    "Long missing",
+    "On order",
+    "Restricted",
+    "Unavailable",
+    "Unknown"
+  })
+  @Test
+  public void itemCannotBeMarkedAsInProcessNonRequestableWhenInAcceptableSourceStatus(String sourceStatus) {
```
marcjohnson-kint, line 70:
```suggestion
  public void itemCannotBeMarkedAsInProcessNonRequestableWhenNotInAcceptableSourceStatus(String sourceStatus) {
```

src/test/java/org/folio/inventory/validation/experimental/TargetItemStatusValidatorTest.java
marcjohnson-kint: Resolving this as @AnShapovalov finds the test useful for checking consistency in the implementation

marcjohnson-kint (approved):

src/main/java/org/folio/inventory/validation/experimental/UnavaliableTargetItemStatusValidator.java
```diff
@@ -0,0 +1,20 @@
+package org.folio.inventory.validation.experimental;
+
+import org.folio.inventory.domain.items.ItemStatusName;
+
+import static java.util.Set.of;
+
+public class UnavaliableTargetItemStatusValidator extends AbstractTargetItemStatusValidator {
```
marcjohnson-kint, line 7: I think there is a typo in `unavailable`

```suggestion
public class UnavailableTargetItemStatusValidator extends AbstractTargetItemStatusValidator {
```

src/test/java/api/items/MarkItemInProcessApiTests.java
```diff
@@ -61,14 +60,7 @@ public void testMarkItemInProcessCirculationNotes() throws Exception {
...
   @Parameters({
-    "Available",
-    "Awaiting delivery",
-    "In transit",
-    "Lost and paid",
-    "Missing",
-    "Order closed",
-    "Paged",
-    "Withdrawn"
+    "Available"
```
marcjohnson-kint, line 63: Why are these tests still parameterised when there is only a single parameter?

PR #296, opened 2021-01-21
MODINV-349 mod-inventory fails with OOO when a large MARC file is being imported

OleksiiKuzminov:
## Purpose
To fix an issue with OOM. Request from mod-inventory to the mod-inventory-storage takes 3-7 seconds and slow down data-import.

## Approach
Make requests to the storage for data-import handlers in an async way without waiting for a result to do not stop processing instance. The multiple waits for requests created a lot of objects in memory and perform OOM.
Added a retry mechanism in case of timeout.


src/main/java/org/folio/inventory/resources/EventHandlers.java
```diff
@@ -46,12 +47,16 @@
...
+  // value in ms
+  private static final int DEFAULT_HTTP_TIMEOUT = 3000;
 
   private final Storage storage;
 
-  public EventHandlers(final Storage storage, final HttpClient client) {
+  public EventHandlers(final Storage storage) {
     Vertx vertx = Vertx.vertx();
     this.storage = storage;
+    HttpClientOptions params = new HttpClientOptions().setConnectTimeout(DEFAULT_HTTP_TIMEOUT);
```
marcjohnson-kint, line 58: What happens when the connection duration exceeds the connect timeout?

src/main/java/org/folio/inventory/resources/EventHandlers.java
```diff
@@ -46,12 +47,16 @@
 
   private static final String DATA_IMPORT_EVENT_HANDLER_PATH = "/inventory/handlers/data-import";
   private static final String INSTANCES_EVENT_HANDLER_PATH = "/inventory/handlers/instances";
+  // value in ms
+  private static final int DEFAULT_HTTP_TIMEOUT = 3000;
```
marcjohnson-kint, line 51: Rather than use a comment, the unit could be explicit in the name

```suggestion
  private static final int DEFAULT_HTTP_TIMEOUT_IN_MILLISECONDS = 3000;
```

src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractInstanceEventHandler.java
```diff
@@ -49,38 +47,35 @@
...
+    ids.forEach(id -> precedingSucceedingTitlesRepository
+      .delete(id)
       .whenComplete((v, e) -> {
-        if (e != null) {
-          promise.fail(e);
-          return;
+          if (e != null) {
+            LOGGER.error("Error during deleting PrecedingSucceedingTitles with ids {}", e, id);
+            LOGGER.info("Error during deleting PrecedingSucceedingTitles retry delete PrecedingSucceedingTitles");
+            precedingSucceedingTitlesRepository.delete(id);
```
marcjohnson-kint, line 74: This code appears to retry the deletion if it failed. Have I interpreted that correctly?

If so, what happens if the second delete also fails?

src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractInstanceEventHandler.java
```diff
@@ -49,38 +47,35 @@
...
+          }
         }
-        promise.complete();
-      });
-    return promise.future();
+      ));
+    return Future.succeededFuture();
```
marcjohnson-kint, line 78: This code appears to finish immediately without waiting for any / all of the deletions to happen. Have I interpreted this correctly?

src/main/java/org/folio/inventory/resources/EventHandlers.java
OleksiiKuzminov: Default TimeoutExeption. In mod-inventory, it is 5 seconds, but no need to wait for more than 3 for this case

src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractInstanceEventHandler.java
OleksiiKuzminov: Yes. For the current pub-sub implementation it is one way to avoid stacking future chains and oom. In the next implementation with kafka it can be changed

KaterynaSenchenko (approved):

VRohach (approved):

PR #302, opened 2021-02-02
Set default values for kafka configs

KaterynaSenchenko:


OleksiiKuzminov (approved):

src/main/java/org/folio/inventory/Launcher.java
```diff
@@ -111,11 +111,11 @@ private static void putNonNullConfig(
...
-    configMap.put(KAFKA_HOST, System.getenv(KAFKA_HOST));
-    configMap.put(KAFKA_PORT, System.getenv(KAFKA_PORT));
-    configMap.put(OKAPI_URL, System.getenv(OKAPI_URL));
-    configMap.put(KAFKA_REPLICATION_FACTOR, System.getenv(KAFKA_REPLICATION_FACTOR));
-    configMap.put(KAFKA_ENV, System.getenv(KAFKA_ENV));
+    configMap.put(KAFKA_HOST, System.getenv().getOrDefault(KAFKA_HOST, "kafka"));
+    configMap.put(KAFKA_PORT, System.getenv().getOrDefault(KAFKA_PORT, "9092"));
+    configMap.put(OKAPI_URL, System.getenv().getOrDefault(OKAPI_URL, "http://okapi:9130"));
+    configMap.put(KAFKA_REPLICATION_FACTOR, System.getenv().getOrDefault(KAFKA_REPLICATION_FACTOR, "1"));
+    configMap.put(KAFKA_ENV, System.getenv().getOrDefault(KAFKA_ENV, "folio"));
```
marcjohnson-kint, line 118: Given that the defaults caused the challenges with the previous approach to go unnoticed because the environment they were testing in happened to be aligned with the defaults, might we consider not introducing default values for these and instead actively failing if they aren't present?

KaterynaSenchenko: I'm not sure the module should fail if replication factor is not set, for example. As for kafka host and port it makes sense to make it consistent for all modules - mod-data-import, mod-source-record-manager and mod-source-record-storage are currently using the defaults, so I provided those values here as well. It is a good point though

marcjohnson-kint: bq. I'm not sure the module should fail if replication factor is not set, for example

Why is that externally configurable, are folks expecting that to change depending upon the environment?

bq. As for kafka host and port it makes sense to make it consistent for all modules

I can understand the need for consistency.

In general, I tend to think that environmental configuration should not have defaults (I may well have done this myself in the past and regretted it).

marcjohnson-kint: @KaterynaSenchenko 

> Can we merge this PR to resolve the issues at hand?

I was in no-way intending to stop this work from being merged. Only curious as to why we were reintroducing defaults, given the challenges those presented during the recent outage. Apologies for the confusion.

PR #303, opened 2021-02-04
MODINV-378 - Item moving error

khandramai:
[MODINV-378](https://issues.folio.org/browse/MODINV-378) - Item moving error

**Purpose**
The new field _version (https://github.com/folio-org/mod-inventory-storage/blob/bc8814fd8859508e5da061961cd36d5ee8f89e20/ramls/holdingsrecord.json#L11) was introduced in scope of story [MODINVSTOR-656](https://issues.folio.org/browse/MODINVSTOR-656). This affect items moving flow with error 500 - Can`t map json to 'Holdingsrecord' entity.

**Approach**
Change from getHoldingsRecordCollection to getHoldingCollection.

### DIFF
--- a/src/test/java/api/items/ItemApiMoveExamples.java
+++ b/src/test/java/api/items/ItemApiMoveExamples.java
@@ -208,10 +208,7 @@ public class ItemApiMoveExamples extends ApiTests {
 
     Response postMoveItemsResponse = moveItems(itemsMoveRequestBody);
 
-    assertThat(postMoveItemsResponse.getStatusCode(), is(500));
-    assertThat(postMoveItemsResponse.getContentType(), containsString(TEXT_PLAIN));
-
-    assertThat(postMoveItemsResponse.getBody(), containsString("Can`t map json to 'Holdingsrecord' entity"));
+    assertThat(postMoveItemsResponse.getStatusCode(), is(200));
   }
 
   @Test
--- a/src/test/java/api/items/ItemApiMoveExamples.java
+++ b/src/test/java/api/items/ItemApiMoveExamples.java
@@ -3,7 +3,6 @@ package api.items;
 import static api.ApiTestSuite.ID_FOR_FAILURE;
 import static api.support.InstanceSamples.smallAngryPlanet;
 import static org.folio.inventory.support.http.ContentType.APPLICATION_JSON;
-import static org.folio.inventory.support.http.ContentType.TEXT_PLAIN;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
@@ -188,7 +187,7 @@ public class ItemApiMoveExamples extends ApiTests {
   }
 
   @Test
-  public void cannotMoveDueToHoldingsRecordSchemasMismatching()
+  public void canMoveToHoldingsRecordWithHoldingSchemasMismatching()
     throws MalformedURLException, InterruptedException, ExecutionException, TimeoutException, IllegalAccessException {
 
     UUID instanceId = UUID.randomUUID();
--- a/ramls/holdings-record.json
+++ b/ramls/holdings-record.json
@@ -8,10 +8,6 @@
       "description": "the system assigned unique ID of the holdings record; UUID",
       "$ref": "./uuid.json"
     },
-    "_version": {
-      "type": "integer",
-      "description": "Record version for optimistic locking"
-    },
     "hrid": {
       "type": "string",
       "description": "the human readable ID, also called eye readable ID. A system-assigned sequential ID which maps to the Instance ID"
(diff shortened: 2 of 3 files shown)


khandramai:
FYI - @axelhunn

marcjohnson-kint:
> Can`t map json to 'Holdingsrecord' entity

Why does the additional of a new property cause this mapping to fail? Why doesn't the mapping ignore unrecognised properties?

khandramai:
> > Can`t map json to 'Holdingsrecord' entity
> 
> Why does the additional of a new property cause this mapping to fail? Why doesn't the mapping ignore unrecognised properties?

Hi @marc-johnson, this because **ExternalStorageModuleHoldingsRecordCollection** is used instead of **ExternalStorageModuleHoldingCollection** here https://github.com/folio-org/mod-inventory/blob/c4ce17c29fc834736f4796dac0615e6c17a3e6ec/src/main/java/org/folio/inventory/resources/MoveApi.java#L79? Looks like both of these collections make the same functionality but with difference in implementation. Is it make sense to use ExternalStorageModuleHoldingCollection instead of ExternalStorageModuleHoldingsRecordCollection?

khandramai:
@marc-johnson , JIC I have changed this PR to use **ExternalStorageModuleHoldingCollection** instead of **ExternalStorageModuleHoldingsRecordCollection** and roll back holdings-record schema. Could you please confirm?

marcjohnson-kint:
> @marc-johnson , JIC I have changed this PR to use **ExternalStorageModuleHoldingCollection** instead of **ExternalStorageModuleHoldingsRecordCollection** and roll back holdings-record schema. Could you please confirm?

What are you asking me to confirm?

I don't know why there are two implementations and why one was chosen over the other.

What I would like is a test that reproduces the failure and demonstrates it is resolved, whichever way it is resolved.

khandramai:
OK, let's wait someone from @folio-org/inventory-backend-code-owners who could clarify what is a right way: (1) continuing using of ExternalStorageModuleHoldingsRecordCollection with holding-record schema changing or (2) switch to ExternalStorageModuleHoldingCollection without schema changing. Any of them fixing my problem with MoveApi. Unfortunately, I also don't know why there are two implementations.

I'm sorry for disturbing you. I asked you because only you demonstrated interest for this PR.

marcjohnson-kint:
> I'm sorry for disturbing you. I asked you because only you demonstrated interest for this PR.

You aren't disturbing me. I apologise, that response was rushed and cranky :-/

I was (and am) trying to understand how you would like me to help.

If you are asking for my advice, I would suggest making the smallest change that resolves the issue, that is likely to change the schema (especially as I imagine, the other processes that use that class are likely also broken due to this).

> OK, let's wait someone from @folio-org/inventory-backend-code-owners who could clarify what is a right way

I'm not sure any of the code owners will be able to answer that as this module is worked on by multiple teams, and the code owners likely don't represent all of those teams.

> Unfortunately, I also don't know why there are two implementations.

I think that is probably a question for the Folijet team, as I believe they introduced the second way.

> Any of them fixing my problem with MoveApi.

How do you know that they fix the issue?

src/test/java/api/items/ItemApiMoveExamples.java
```diff
@@ -208,10 +208,7 @@ public void cannotMoveDueToHoldingsRecordSchemasMismatching()
 
     Response postMoveItemsResponse = moveItems(itemsMoveRequestBody);
 
-    assertThat(postMoveItemsResponse.getStatusCode(), is(500));
-    assertThat(postMoveItemsResponse.getContentType(), containsString(TEXT_PLAIN));
-
-    assertThat(postMoveItemsResponse.getBody(), containsString("Can`t map json to 'Holdingsrecord' entity"));
+    assertThat(postMoveItemsResponse.getStatusCode(), is(200));
```
marcjohnson-kint, line 211: This change states that this example is now successful. The name suggests it is intended to demonstrate a failure, please either rename the test or remove it if it is no longer applicable.

(Assuming that you decide to stick with changing the move API to use a different storage collection class)

hjiebsco (commented):
Looks like this branch has not been updated to use the newer mod-inventory-storage. I am curious why would the test fail?

marcjohnson-kint:
> Looks like this branch has not been updated to use the newer mod-inventory-storage. 

I don't think I follow. What would you expect to be updated in `mod-inventory` when `mod-inventory-storage`

> I am curious why would the test fail?

That test simulates a new property being added to the `holdings-storage holdings-record` record representation without changes to the schema here as well. 

The failure is caused by that code using a copy of the schema to drive JSON mapping which disallows additional properties, so any new property will cause it to fail.

hjiebsco (commented):
The line 79 change from `getHoldingsRecordCollection` to `getHoldingCollection` seems to be fine since only `id` is used in that scope. But I am not familiar with mod-inventory code, I will let others to chime in and approve this PR.

hjiebsco (commented):
BTW you may consider updating the PR description to reflect the new approach so it won't be confusing for new reviewers.

axelhunn (approved):

siarhei-charniak (approved):

PR #304, opened 2021-02-04
MODINV-377

AnShapovalov:


src/test/java/org/folio/inventory/validation/status/GenerateItemStatusTransitionsMatrix.java
```diff
@@ -0,0 +1,60 @@
...
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@RunWith(JUnitParamsRunner.class)
+public class GenerateItemStatusTransitionsMatrix {
```
marcjohnson-kint, line 18: Is this similar to the intent of `TargetItemStatusValidatorsTest`?

I think it is pretty similar in purpose, could they be combined?

src/test/java/org/folio/inventory/validation/status/GenerateItemStatusTransitionsMatrix.java
```diff
@@ -0,0 +1,60 @@
...
+  @BeforeClass
+  public static void setUp() throws Exception {
+    targetItemStatusValidators = new TargetItemStatusValidators();
+  }
+
+  @Test
+  public void generate() {
+    final String fieldDelim = "\t";
+    final String stringDelim = "\"";
+    final List<String> sortedStatusNames = Arrays.stream(ItemStatusName.values()).map(itemStatusName -> itemStatusName.value()).sorted().collect(Collectors.toList());
```
marcjohnson-kint, line 30: I suggest splitting this line to make it easier to identify the individual steps.

```suggestion
    final List<String> sortedStatusNames = Arrays.stream(ItemStatusName.values())
      .map(itemStatusName -> itemStatusName.value())
      .sorted()
      .collect(Collectors.toList());
```

src/test/java/org/folio/inventory/validation/status/GenerateItemStatusTransitionsMatrix.java
```diff
@@ -0,0 +1,60 @@
...
+    final Date dateNow = new Date();
+    StringBuilder sb = new StringBuilder();
+    sb.append("Generated at:"+dateNow);
+    sb.append(System.lineSeparator());
+    sb.append(stringDelim+"Initial\\Target"+stringDelim+fieldDelim);
+    sortedStatusNames.stream().forEach(statusName -> sb.append(stringDelim+statusName+stringDelim+fieldDelim));
+    sb.append(System.lineSeparator());
+    sortedStatusNames.stream().forEach(initStatusName -> {
+      sb.append(stringDelim+initStatusName+stringDelim+fieldDelim);
```
marcjohnson-kint, line 39: FOLIO code standards state that there should be spaces between operators, please change the formatting of this line (and others in this file accordingly)

```suggestion
      sb.append(stringDelim + initStatusName + stringDelim + fieldDelim);
```

src/test/java/org/folio/inventory/validation/status/GenerateItemStatusTransitionsMatrix.java
AnShapovalov: TargetItemStatusValidatorsTest has annotation @Ignore by default. It contains some on demand routines used in development process and shouldn't be run on each build. GenerateItemStatusTransitionMatrix as for me should running on each build to have possibility for produced result to be automatically published at dedicated place or simply have access to it for each build without any special actions. It doesn't consume significant computation resources.

marcjohnson-kint: > GenerateItemStatusTransitionMatrix as for me should running on each build to have possibility for produced result to be automatically published at dedicated place or simply have access to it for each build without any special actions.

What is the value of executing this process on every build when it is a part of a manual process for using the output for comparison or publication?

AnShapovalov: Practically this information could be automatically shared throgh GoogleSheet, Wiki, Build host or any other way. Final decision should be discussed with person(s) who is responsible for build environment. This is a half way to this goal. Now anyone who has access can open generated file and simple copy/paste it in googlesheet and possible in Excell (not tested).

marcjohnson-kint: @AnShapovalov 

> Practically this information could be automatically shared throgh GoogleSheet, Wiki, Build host or any other way.

It could, however it currently isn't. At the moment, any output from this process has to be taken by a developer and external documentation updated manually.

Given that I suggest this be combined with the other test (neither are really tests), given that both serve the purpose of checking changes to these transitions and be executed manually.

> Final decision should be discussed with person(s) who is responsible for build environment.

I don't understand how the folks responsible for the build environments are involved in this.

> This is a half way to this goal. Now anyone who has access can open generated file and simple copy/paste it content in googlesheet and possible in Excell (not tested).

What is half way to what goal? At the moment, the only folks who have access to the generated file are developers.

Have you agreed with the relevant PO when / how the spreadsheet should be updated with the output of this test? As far as I can tell, it is currently not being used.

AnShapovalov: > I don't understand how the folks responsible for the build environments are involved in this.
It linked with a permissions to update remote storage through API such as GoogleSheet, WIKI, etc. For example Google sheets API has enough complex set of requirements for secure access implementation. Google security artefacts should be stored on host(or integrated in application) for allow client application communicate with Google API. So separating tools responsible for generating data and upload it would be right way for save build process reliable. From this viewpoint simple copy and overwrite generated file on reliable shared resource would be best way for avoid any problem related build reliability. 

>What is half way to what goal? At the moment, the only folks who have access to the generated file are developers.
Also access to it has service account which make builds for testing purposes so if it will overwrite content of generated file on specified shared resource it would be enough to moving forward.

> Have you agreed with the relevant PO when / how the spreadsheet should be updated with the output of this test? As far as I can tell, it is currently not being used.
Not yet. Initially I thought about GoogleSheet, but it contains enough borders linked with security of accounts whichusing API. If you is not an account administrator then need a lot of agreements or impossible at all by security policy of organization owner(in case when we talking about Google Docs as collaboration service for organization as I saw our PO's using it in such way). Or we could create free google account and use it for this goal. The other way is using folio Confluence or Jira where build account should have permissions to write. It should be discussed.

marcjohnson-kint: I agree that this should be discussed. I suggest that you outline the scope and intent of this work on the JIRA issue.

I don't think what you are describing aligns with Charlotte's expectations around this work. And I doubt she is expecting it to involve time from other folks outside of the team.

marcjohnson-kint:
@AnShapovalov I'm closing this pull request as the issue was closed without it being merged or an agreement on the process for keeping this information up to date

PR #305, opened 2021-02-04
MODINV-356-2Scenario-fix

AnShapovalov:


src/test/java/org/folio/inventory/validation/status/InProcessItemStatusValidatorTest.java
```diff
@@ -57,12 +62,7 @@ public void itemCanBeMarkedAsInProcessWhenInAcceptableSourceStatus(String source
   }
 
   @Parameters({
-    "Aged to lost",
-    "Checked out",
-    "Claimed returned",
-    "Declared lost",
-    "In process",
-    "On order"
+    "In process"
```
marcjohnson-kint, line 65: Why is it unacceptable to transition an `In process` item to `In process`, is that not effectively no operation at all?

AnShapovalov: That's only one parameter for negation scenario for In process target status validator. The original of source code copy you could find below.
@Parameters({
    "In process"
  })
  @Test
  public void itemCannotBeMarkedAsInProcessWhenNotInAcceptableSourceStatus(String sourceStatus) {

AnShapovalov: I didn't remove parameters for save same style for all validators tests.

marcjohnson-kint: I wasn't asking about the original source or the style of the test.

I'm asking why is it unacceptable to change an item that is `In process` to `In process`, as I believe that would effectively not do anything.

AnShapovalov: According to requirements Item Status transition to itself is not allowed. Why? I could make prediction only based on common workflow logic. For example we are definitely lost information about initial date when item was really moved in this status if status change history implementation is absent or(and) especial handler for this case wasn't defined. The same could have relation to information about person who marked item in this status initially. I'm sure that date of making of transition is defenitely significant for certain states. I believe POs could give more wide explanation about librarian workflow.

marcjohnson-kint: I thought, given that you've expressed interest in the workflow, you might have been curious about this exception. If this was allowed, however was a non-operation, meaning no changes / side effects occurred, the code could be significantly simplified (as for these target statuses, the transition would always be allowed)

marcjohnson-kint (approved):

src/test/java/org/folio/inventory/validation/status/InProcessItemStatusValidatorTest.java
AnShapovalov: If I understand you idea in right way you suggest to have an especial handler for mark Item to itself status or leave it unhandled. I met case when marking item to itself was a helpful action which confirmed it status for certain moment (for example when one responsible for item person should be changed to other by any reason and confirm item status at same time). I have no any basement to think that case like mentioned above won't occure in librarian workflow. As for me mainline is to have unified code which allow manage statuses and transiton in unified way with avoiding exceptions as possible.

marcjohnson-kint: > If I understand you idea in right way you suggest to have an especial handler for mark Item to itself status or leave it unhandled.

I'm suggesting that requests made to change an item to the status it is already in could be considered successful without any action being taken.

> I met case when marking item to itself was a helpful action which confirmed it status for certain moment (for example when one responsible for item person should be changed to other by any reason and confirm item status at same time). I have no any basement to think that case like mentioned above won't occur in librarian workflow.

As I understand it, that is the opposite of what you said above about corrupting metadata by performing a false change of status. That would be a question for the relevant Product Owner.

> As for me mainline is to have unified code which allow manage statuses and transiton in unified way with avoiding exceptions as possible. 

Some of that code could be entirely avoided by avoiding the special case of avoiding transitions to the same status.

> Would be great if "manage of item statuses and transition" role could be assigned to person who closer to business process then to programming.

I'm not sure I understand what you mean. The person responsible for managing item statuses and transitions is the relevant Product Owner not a developer.

PR #308, opened 2021-02-17
Fix package naming after di-core update

KaterynaSenchenko:


KaterynaSenchenko:
Merging this PR to fix the compilation errors after the di-core update

PR #313, opened 2021-02-24
MODINV-155 Map shelving order property for items

K-Felk:
I added code to the item representation to hold the specified data, code
in the mapping utility, and with help, code in the class that constructs 
the response to item data requests.

As this is a read-only value, I've limited the mappings for it to the 
code responsible for displaying item data.  I've checked where the other 
mapping utilities in itemUtil are called from, and the rest of them only 
seem to be used when updating or creating item records. 

I was also having problems with intermittent test failure of one of the tests:
shouldSendEventWithProcessedEventPayloadWhenProcessingCoreHandlerSuccessful
It looks like what was happening was the test wasn't waiting long enough for 
a response to a back-end request-when I increased the timeout value, the test 
started passing consistently.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ## 16.2.0 IN-PROGRESS
 
 * Map shelving order property for items (MODINV-155)
+* Provides `inventory` interface 10.9 (MODINV-155)
 * Requires `item-storage` interface 8.9 (MODINV-155)
 
 ## 16.1.0 2020-10-12
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 16.2.0 IN-PROGRESS
+
+* Map shelving order property for items (MODINV-155)
+
 ## 16.1.0 2020-10-12
 
 * Provides the ability to move items between holdings (MODINV-309)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ## 16.2.0 IN-PROGRESS
 
 * Map shelving order property for items (MODINV-155)
+* Requires `item-storage` interface 8.9 (MODINV-155)
 
 ## 16.1.0 2020-10-12
 
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -54,6 +54,11 @@
       "description": "Text, with input likely validated by the barcode scanner",
       "type": "string"
     },
+    "effectiveShelvingOrder": {
+      "type": "string",
+      "description": "A system generated normalization of the call number that allows for call number sorting in reports and search results",
+      "readonly": true
+    },
     "accessionNumber": {
       "type": "string",
       "description": "Accession number is also called inventar number. Accession number is a unique number assigned to an item in the order in which it is added to a library collection. Most libraries assign accession numbers in continous numerical sequece, but some use a code system to indicate type of material and/or year of accession in addition to order of accession"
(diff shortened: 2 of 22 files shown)


ramls/item.json
```diff
@@ -54,6 +54,11 @@
       "description": "Text, with input likely validated by the barcode scanner",
       "type": "string"
     },
+    "effectiveShelvingOrder": {
+      "type": "string",
+      "description": "A system generated normalization of the call number that allows for call number sorting in reports and search results",
+      "readonly": true
+    },
```
marcjohnson-kint, line 61: Please change the minor version of all interfaces that use this schema to reflect this new property

marcjohnson-kint: Please also change the dependency on `item-storage` to use the interface version that includes the new property in storage.

src/test/java/org/folio/inventory/dataimport/consumers/DataImportConsumerVerticleTest.java
```diff
@@ -163,7 +162,7 @@ public void shouldSendEventWithProcessedEventPayloadWhenProcessingCoreHandlerSuc
     // then
     String observeTopic = KafkaTopicNameHelper.formatTopicName(KAFKA_ENV_NAME, getDefaultNameSpace(), TENANT_ID, DI_COMPLETED.value());
     cluster.observeValues(ObserveKeyValues.on(observeTopic, 1)
-      .observeFor(20, TimeUnit.SECONDS)
+      .observeFor(30, TimeUnit.SECONDS)
```
marcjohnson-kint, line 165: Thanks for trying to resolve this, I haven't been able to reliably run this test yet

ramls/item.json
K-Felk: The only raml file I can find that directly references item.json and has versioning information is inventory.raml-I incremented that version to 10.9.  I also changed the versioning of that module in moduledescriptor-template to match.  I updated the reference to item-storage in moduledescriptor-template to match the one in mod-inventory-storage (8.9). I hope that's right?

src/test/java/org/folio/inventory/dataimport/consumers/DataImportConsumerVerticleTest.java
K-Felk: My only other guess here is that it's a tenant initialization problem, similar to what i encountered when updating mod-users to raml-module-builder 32.  That could account for the intermittent nature of the failure, and it could also explain why waiting longer makes the test pass more often.   I could try putting in wait code to check on the tenant initialization-except that it's not clear to me in this code where that's happening.

NEWS.md
```diff
@@ -1,6 +1,7 @@
 ## 16.2.0 IN-PROGRESS
 
 * Map shelving order property for items (MODINV-155)
+* Requires `item-storage` interface 8.9 (MODINV-155)
```
marcjohnson-kint, line 4:
```suggestion
* Provides `inventory` interface 10.9 (MODINV-155)
* Requires `item-storage` interface 8.9 (MODINV-155)
```

ramls/item.json
marcjohnson-kint: I think that's right

src/test/java/org/folio/inventory/dataimport/consumers/DataImportConsumerVerticleTest.java
marcjohnson-kint: I don't know either. Thank you for trying to make it more stable.

marcjohnson-kint (approved):

PR #332, opened 2021-05-10
MODINV-410 Store MARC Authority record

oleksandr-dekin:
## Purpose
Change record type MARC for more specific type, like MARC_BIB, MARC_AUTHORITY, MARC_HOLDING

## Approach
- updated data type from marc to marc_bib

## Learning
[MODINV-410](https://issues.folio.org/browse/MODINV-410)


oleksandr-dekin:
Tests will be passed when https://github.com/folio-org/mod-source-record-storage/pull/331 and https://github.com/folio-org/data-import-processing-core/pull/144  will be merged into master, because mod-inventory uses in pom.xml the next dependencies:  `data-import-processing-core` and `mod-source-record-storage-client`

psmagin (approved):

.rancher-pipeline.yml
```diff
@@ -0,0 +1,27 @@
...
+          catalogTemplate: p-htqfq:spitfire-helmcharts-mod-inventory
+          version: 0.1.32
+          answers:
+            image.repository: docker.dev.folio.org/mod-inventory
+            image.tag: spitfire-${CICD_EXECUTION_SEQUENCE}
+            postJob.enabled: true
+          targetNamespace: spitfire
+          name: mod-inventory
+timeout: 60
+notification: {}
```
marcjohnson-kint, line 27: Please remove this before the branch is merged into the mainline. Other teams might be using scratch / rancher environments and have their own configuration

ramls/examples/instancesBatch_post.json
```diff
@@ -24,7 +24,7 @@
     },
     {
       "title": "Open Bibliography for Science, Technology and Medicine",
-      "source": "MARC",
+      "source": "MARC_BIB",
```
marcjohnson-kint, line 27: Is there an issue for changing all of the existing instance records that have a source of `MARC` to `MARC_BIB` during an upgrade process?

marcjohnson-kint:
When raising pull requests, please [reference](https://wiki.folio.org/display/DD/Pull+Request+Guidelines) a JIRA issue for the project associated with the module (in this case MODINV). This is important for managing what changes were included in a particular release.

ramls/examples/instancesBatch_post.json
oleksandr-dekin: I discussed with PO and as result - we don't need to separate source by record type. I reverted changes related to source. Only dataType was changed from MARC to MARC_BIB.

Thanks for comment!

src/test/java/org/folio/inventory/dataimport/consumers/DataImportKafkaHandlerTest.java
```diff
@@ -73,7 +73,7 @@
   private JobProfile jobProfile = new JobProfile()
     .withId(UUID.randomUUID().toString())
     .withName("Create instance")
-    .withDataType(org.folio.JobProfile.DataType.MARC);
+    .withDataType(org.folio.JobProfile.DataType.MARC_BIB);
```
marcjohnson-kint, line 76: Does this mean that any existing job profiles with a data type of `MARC` need to be changed to `MARC_BIB`?

oleksandr-dekin: Yes. we shouldn't have record type `MARC`, for record type we have - `MARC_BIB`, `MARC_AUTHORITY`, `MARC_HOLDING`.
For source we have `MARC`, but in this case `MARC` is a format

marcjohnson-kint: Does that mean there is going to be a upgrade process to move the existing job profiles to the new data type?

oleksandr-dekin: Existing job profiles will be migrated in the another module - mod-data-import-converter-storage. You can find script for migration here: https://github.com/folio-org/mod-data-import-converter-storage/pull/164/files#diff-dcf9376ead7f99847995503e9dd1b771daf05bc357b1c117c2655e06b1320edb

Or maybe you talking about `mod_inventory_storage`.  But I double-check and didn't find dataType in the `instance`, `item` or holding. Only field `source`.

marcjohnson-kint: > Existing job profiles will be migrated in the another module - mod-data-import-converter-storage

Thanks for clarifying that

> Or maybe you talking about mod_inventory_storage

I wasn't, only the owner of the job profiles

Dockerfile
```diff
@@ -8,5 +8,10 @@ ENV VERTICLE_HOME /usr/verticles
 # Copy your fat jar to the container
 COPY target/${VERTICLE_FILE} ${VERTICLE_HOME}/${VERTICLE_FILE}
 
+ENV LIB_DIR ${VERTICLE_HOME}/lib
+
+RUN mkdir -p ${LIB_DIR}
+COPY lib/* ${LIB_DIR}/
```
marcjohnson-kint, line 14: Why isn't the `data-impor-processing-core-fat` library included in the jar for mod-inventory rather than externally?

oleksandr-dekin: It's only for testing on the rancher, we must include these libraries because mod-inventory use several DataTypes. There will be a rollback before the merge

PR #338, opened 2021-05-24
MODINV-388, set read-only bound-with flag on Instance and Item responses

nielserik:
Purpose:  

To set a boolean, read-only flag on Instances, to indicate whether a given title is included in a bound-with, and
likewise set a boolean, read-only flag on Items in responses, to indicate whether the given Item is a bound-with (has multiple holdings/titles included)

The flags are set in Instance and Item responses behind the scenes based on a new Inventory Storage API,  `bound-with-parts-storage'.  

The flags are set for the UI to display icons on Instances and Items to alert the user (ie the librarian) of the special characteristics of those particular entities. The logic for setting the flags is considered too complex for the UI client itself to feasibly resolve the flag directly from the bound-with API in Inventory Storage. 

A client is allowed to push the bound-with flags back to the API in storage request but the flag's setting will be ignored.  

The change is non-breaking but requires the new storage API to be available. In the branch I've upgraded `inventory`s  version from 10.10 to 10.11. I have not updated NEWS.


ramls/inventory.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Inventory API
-version: v10.10
+version: v10.11
```
marcjohnson-kint, line 3: A corresponding change should be made to the provided interfaces in the module descriptor

ramls/instance.json
```diff
@@ -403,6 +403,12 @@
         "$ref": "uuid.json"
       }
     },
+    "isBoundWith": {
+      "description": "Indicates if this Instance is included in a bound-with",
+      "type": "boolean",
+      "default": false,
+      "readonly": true
+    },
```
marcjohnson-kint, line 411: Given that an instance is an abstract description of a bibliographic entity, how can an instance be bound with anything?

marcjohnson-kint:
> A client is allowed to push the bound-with flags back to the API in storage request but the flag's setting will be ignored.

I don't think I understand this. Aren't these properties derived in the business logic APIs and as such, they don't exist in the storage APIs. Have I misunderstood?

ramls/inventory.raml
nielserik: Ha, thank you Marc. Trips this developer up every other time. Done.

nielserik:
>> A client is allowed to push the bound-with flags back to the API in storage request but the flag's setting will be ignored.

>I don't think I understand this. Aren't these properties derived in the business logic APIs and as such, they don't exist in the storage APIs. Have I misunderstood?

marcjohnson-kint:
> > > A client is allowed to push the bound-with flags back to the API in storage request but the flag's setting will be ignored.
> 
> > I don't think I understand this. Aren't these properties derived in the business logic APIs and as such, they don't exist in the storage APIs. Have I misunderstood?
> 
> That's right Marc, it's derived and doesn't exist i storage. 

Then can you help me why the storage request is mentioned in the description for this pull request?

> The flag is just accepted in PUTs for the convenience of the UI. I believe our UI may retrieve Instances from BL, modify them and PUT them back through BL. In that process the UI won't have to worry about removing the boundWith flag (which I think might be a UI breaking change even if not formally a breaking change perhaps).

Why would this property be a breaking change for the UI, given it can be sent or not?

nielserik:
> Why would this property be a breaking change for the UI, given it can be sent or not?

well, if  ui-inventory retrieves the Instance (now with the flag), modifies other properties, and then PUTs the Instance back, then the request would, perhaps surprisingly, fail unless the isBoundWith flag is being allowed in PUTs. 

This PUT is also what I was referring to with the term 'storage requests'.

marcjohnson-kint:
> > Why would this property be a breaking change for the UI, given it can be sent or not?
> 
> well, if ui-inventory retrieves the Instance (now with the flag), modifies other properties, and then PUTs the Instance back, then the request would, perhaps surprisingly, fail unless the isBoundWith flag is being allowed in PUTs.

I thought you said that that the `isBoundWith` property is being ignored by the business logic APIs so as to avoid this?
 
> This PUT is also what I was referring to with the term 'storage requests'.

Do you mean if a client took the business logic representation of the record and the made a PUT request direct to storage?

PR #344, opened 2021-05-28
MODINV-398: Optimistic locking (_version field, 409 status)

julianladisch:
mod-inventory-storage enabled optimistic locking for
instances/holdings/items.

If the _version property is provided, it must be returned unchanged
when updating the record using PUT, otherwise it will be rejected
with 409 status code.


marcjohnson-kint (commented):
As stated in Slack, I think the end of the development period for a release cycle (and now after the development freeze) is an unfortunate time to make (what are considered to be) breaking compatibility changes, as it does not really allow much time for resolving any downstream impacts and the cost of change increases given that releases of these modules are imminent (after the schema upgrade testing and release of storage modules).

v-shalaev (approved):

PR #351, opened 2021-06-04
MODINV-430: fix: blocked mirror for repositories: indexdata

Aliaksandr-Fedasiuk:
## Purpose
Blocked mirror for repositories: [indexdata (http://maven.indexdata.com/, default, releases+snapshots)]

## Approach
This problem is related to the maven version you are using in your pipeline, it seems that it doesnt accept http url for downloading dependencies, to solve this problem , you should change the docker image you are using to : maven:3.6.3-openjdk-11 and the pipeline will run successfully.

### DIFF
--- a/.rancher-pipeline.yml
+++ b/.rancher-pipeline.yml
@@ -7,7 +7,7 @@ stages:
   - name: Build
     steps:
       - runScriptConfig:
-          image: maven:3-adoptopenjdk-11
+          image: maven:3.6.3-openjdk-11
           shellScript: mvn package -DskipTests
   - name: Build Docker with DIND
     steps:


RuslanLavrov (approved):

julianladisch (changes requested):
The latest maven version prevents http maven repositories to fix this MitM (Machine in the Middle) security issue. https should be used instead.
Can you update the mod-inventory dependencies so that https is used?
See https://issues.folio.org/browse/MODINV-430

julianladisch:
Downgrading maven to 3.6.3 doesn't fix the security issue: https://maven.apache.org/docs/3.8.1/release-notes.html#cve-2021-26291

PR #355, opened 2021-06-08
MODINV-428: Propagate publicationPeriod property.

bohdan-suprun:
Propagate the `publicationPeriod` property for instance on update/create/get.
Increase `inventory` API version to `10.12`
Require `instance-storage` of `7.8` version

### DIFF
--- a/src/main/java/org/folio/inventory/support/JsonHelper.java
+++ b/src/main/java/org/folio/inventory/support/JsonHelper.java
@@ -1,11 +1,13 @@
 package org.folio.inventory.support;
 
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 
-import org.apache.commons.lang3.StringUtils;
+import java.util.Optional;
 
 import io.vertx.core.json.JsonObject;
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/inventory/support/JsonHelper.java
```diff
@@ -28,14 +30,23 @@ public static String getString(JsonObject representation, String propertyName) {
...
     }
   }
 
+  @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+  public static <T> void includeIfPresent(
+    JsonObject representation, String propertyName, Optional<T> value) {
+
+    if (representation != null && isNotBlank(propertyName) && value.isPresent()) {
+      representation.put(propertyName, value.get());
+    }
```
hjiebsco, line 47:
```suggestion
    includeIfPresent(representation, propertyName, value.isPresent() ? value : null);
```

src/main/java/org/folio/inventory/support/JsonHelper.java
```diff
@@ -28,14 +30,23 @@ public static String getString(JsonObject representation, String propertyName) {
...
+    JsonObject representation, String propertyName, T value) {
 
-    if (representation != null && StringUtils.isNotBlank(propertyName) && value != null) {
+    if (representation != null && isNotBlank(propertyName) && value != null) {
       representation.put(propertyName, value);
     }
   }
 
+  @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+  public static <T> void includeIfPresent(
```
hjiebsco, line 42: This method seems a bit redundant. If there is a static `PublicationPeriod.toJson(PublicationPeriod publicationPeriod)`, you can avoid adding this one.

hjiebsco (approved):

src/main/java/org/folio/inventory/support/JsonHelper.java
bohdan-suprun: Make sense, I've removed this method and made the toJson to be static.

pfilippovEPAM (approved):
Looks good to me

PR #357, opened 2021-06-11
Change version to 17.0.0 MODINV-420

marcjohnson-kint:
As module depends upon new major (3.0) version of `source-storage-records`. This was changed during the work for [MODINV-422](https://issues.folio.org/browse/MODINV-422).


rladdusaw (approved):

marcjohnson-kint:
@rladdusaw Thank you

PR #369, opened 2021-07-02
Support many related instances in single Instance (MODINV-446)

nielserik:
The default RMB limit of 10 records has been applying to the population of the parent and child relations arrays of an Instance when fetching a  single Instance by ID. 

With this PR this limit would be removed, ie to support displaying many more relations in the Instance details view. (UIIN-1545)

### DIFF
--- a/src/main/java/org/folio/inventory/resources/AbstractInstances.java
+++ b/src/main/java/org/folio/inventory/resources/AbstractInstances.java
@@ -8,6 +8,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -71,11 +72,11 @@ public abstract class AbstractInstances {
     CollectionResourceClient relatedInstancesClient = createInstanceRelationshipsClient(
       routingContext, context);
     CollectionResourceRepository relatedInstancesRepository = new CollectionResourceRepository(relatedInstancesClient);
-    List<String> instanceId = Arrays.asList(instance.getId());
+    List<String> instanceId = Collections.singletonList( instance.getId() );
     String query = createQueryForRelatedInstances(instanceId);
 
     CompletableFuture<Response> future = new CompletableFuture<>();
-    relatedInstancesClient.getMany(query, future::complete);
+    relatedInstancesClient.getMany(query, Integer.MAX_VALUE, 0, future::complete);
 
     return future.thenCompose(result ->
       updateInstanceRelationships(instance, relatedInstancesRepository, result));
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -71,11 +72,11 @@ public AbstractInstances(final Storage storage, final HttpClient client) {
     CollectionResourceClient relatedInstancesClient = createInstanceRelationshipsClient(
       routingContext, context);
     CollectionResourceRepository relatedInstancesRepository = new CollectionResourceRepository(relatedInstancesClient);
-    List<String> instanceId = Arrays.asList(instance.getId());
+    List<String> instanceId = Collections.singletonList( instance.getId() );
```
julianladisch, line 75:
```suggestion
    List<String> instanceId = List.of(instance.getId());
```

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -134,7 +135,7 @@ public AbstractInstances(final Storage storage, final HttpClient client) {
     CollectionResourceRepository precedingSucceedingTitlesRepository =
       new CollectionResourceRepository(precedingSucceedingTitlesClient);
 
-    List<String> instanceId = Arrays.asList(instance.getId());
+    List<String> instanceId = Collections.singletonList( instance.getId() );
```
julianladisch, line 138:
```suggestion
    List<String> instanceId = List.of(instance.getId());
```

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -520,7 +520,7 @@ private void getById(RoutingContext routingContext) {
     if (relatedInstancesClient != null) {
       CompletableFuture<Response> relatedInstancesFetched = new CompletableFuture<>();
 
-      relatedInstancesClient.getMany(query, relatedInstancesFetched::complete);
+      relatedInstancesClient.getMany( query, Integer.MAX_VALUE, 0, relatedInstancesFetched::complete);
```
julianladisch, line 523:
```suggestion
      relatedInstancesClient.getMany(query, Integer.MAX_VALUE, 0, relatedInstancesFetched::complete);
```

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -71,11 +72,11 @@ public AbstractInstances(final Storage storage, final HttpClient client) {
     CollectionResourceClient relatedInstancesClient = createInstanceRelationshipsClient(
       routingContext, context);
     CollectionResourceRepository relatedInstancesRepository = new CollectionResourceRepository(relatedInstancesClient);
-    List<String> instanceId = Arrays.asList(instance.getId());
+    List<String> instanceId = Collections.singletonList( instance.getId() );
     String query = createQueryForRelatedInstances(instanceId);
 
     CompletableFuture<Response> future = new CompletableFuture<>();
-    relatedInstancesClient.getMany(query, future::complete);
+    relatedInstancesClient.getMany(query, Integer.MAX_VALUE, 0, future::complete);
```
julianladisch, line 79: The magic values `Integer.MAX_VALUE, 0` should be avoided by adding a method like getAll(String query, Consumer<Response>) to CollectionResourceClient. This also avoids duplication of this code in Instances.java.

Javadoc should be added to CollectionResourceClient.getMany pointing out the most APIs be default have a limit of 10 records.

PR #370, opened 2021-07-03
Remove limit of 10 parents/children in single instance (MODINV-446)

nielserik:
When retrieving a single instance by ID,  the default RMB limit of 10 records has been in effect and ui-inventory, for example, could thus only display the first 10 parent and/or child relationships of an instance. 

With this PR this limitation would be removed. 

When retrieving multiple instances by query, there was already no such limit or parents/children of each instance in the result set.

The same issue is fixed for precedingSucceedingTitles.

A new getAll method is used for this.

The getMany without limit/offset parameters is no longer in use and is deleted.

The urlEncoding and cqlEncoding is fixed - for separation of concerns:
* When building a CQL query do cqlEncoding. For example in CqlQuery.
* When building a URL do urlEncoding.

CQL encoding and URL encoding is moved to the correct places. This reduces the code and improved readability.

### DIFF
--- a/src/main/java/org/folio/inventory/resources/AbstractInstances.java
+++ b/src/main/java/org/folio/inventory/resources/AbstractInstances.java
@@ -134,7 +134,7 @@ public abstract class AbstractInstances {
     CollectionResourceRepository precedingSucceedingTitlesRepository =
       new CollectionResourceRepository(precedingSucceedingTitlesClient);
 
-    List<String> instanceId = Collections.singletonList( instance.getId() );
+    List<String> instanceId = List.of(instance.getId());
     String query = createQueryForPrecedingSucceedingInstances(instanceId);
 
     CompletableFuture<Response> future = new CompletableFuture<>();
--- a/src/main/java/org/folio/inventory/resources/AbstractInstances.java
+++ b/src/main/java/org/folio/inventory/resources/AbstractInstances.java
@@ -7,8 +7,6 @@ import java.lang.invoke.MethodHandles;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -75,7 +73,7 @@ public abstract class AbstractInstances {
     String query = createQueryForRelatedInstances(instanceId);
 
     CompletableFuture<Response> future = new CompletableFuture<>();
-    relatedInstancesClient.getMany(query, Integer.MAX_VALUE, 0, future::complete);
+    relatedInstancesClient.getAll(query, future::complete);
 
     return future.thenCompose(result ->
       updateInstanceRelationships(instance, relatedInstancesRepository, result));
@@ -138,7 +136,7 @@ public abstract class AbstractInstances {
     String query = createQueryForPrecedingSucceedingInstances(instanceId);
 
     CompletableFuture<Response> future = new CompletableFuture<>();
-    precedingSucceedingTitlesClient.getMany(query, future::complete);
+    precedingSucceedingTitlesClient.getAll(query, future::complete);
 
     return future.thenCompose(result ->
       updatePrecedingSucceedingTitles(instance, precedingSucceedingTitlesRepository, result));
@@ -342,7 +340,7 @@ public abstract class AbstractInstances {
 
   protected String createQueryForPrecedingSucceedingInstances(List<String> instanceIds) {
     String idList = instanceIds.stream().distinct().collect(Collectors.joining(" or "));
-    return format("query=succeedingInstanceId==(%s)+or+precedingInstanceId==(%s)", idList, idList);
+    return format("succeedingInstanceId==(%s) or precedingInstanceId==(%s)", idList, idList);
   }
 
   protected <T> CompletableFuture<List<T>> allResultsOf(
@@ -355,7 +353,7 @@ public abstract class AbstractInstances {
 
   protected String createQueryForRelatedInstances(List<String> instanceIds) {
     String idList = instanceIds.stream().distinct().collect(Collectors.joining(" or "));
-    return format("subInstanceId==(%s)+or+superInstanceId==(%s)", idList, idList);
+    return format("subInstanceId==(%s) or superInstanceId==(%s)", idList, idList);
   }
 
   protected OkapiHttpClient createHttpClient(
--- a/src/main/java/org/folio/inventory/resources/AbstractInstances.java
+++ b/src/main/java/org/folio/inventory/resources/AbstractInstances.java
@@ -8,6 +8,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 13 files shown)


src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -354,7 +355,7 @@ protected String createQueryForPrecedingSucceedingInstances(List<String> instanc
 
   protected String createQueryForRelatedInstances(List<String> instanceIds) {
     String idList = instanceIds.stream().distinct().collect(Collectors.joining(" or "));
-    return format("query=subInstanceId==(%s)+or+superInstanceId==(%s)", idList, idList);
+    return format("subInstanceId==(%s)+or+superInstanceId==(%s)", idList, idList);
```
marcjohnson-kint, line 358: Why has the query= part of this been removed? Is it because the previous `getMany` method expects a raw HTTP query string?

marcjohnson-kint (commented):
Could a test be added / changed that demonstrates this change fixes the issue?

src/main/java/org/folio/inventory/resources/AbstractInstances.java
julianladisch: Yes: https://github.com/folio-org/mod-inventory/blob/master/src/main/java/org/folio/inventory/storage/external/CollectionResourceClient.java

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -70,11 +71,11 @@ public AbstractInstances(final Storage storage, final HttpClient client) {
     CollectionResourceClient relatedInstancesClient = createInstanceRelationshipsClient(
       routingContext, context);
     CollectionResourceRepository relatedInstancesRepository = new CollectionResourceRepository(relatedInstancesClient);
-    List<String> instanceId = Arrays.asList(instance.getId());
+    List<String> instanceId = Collections.singletonList( instance.getId() );
```
julianladisch, line 74:
```suggestion
    List<String> instanceId = List.of(instance.getId());
```

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -70,11 +71,11 @@ public AbstractInstances(final Storage storage, final HttpClient client) {
...
     String query = createQueryForRelatedInstances(instanceId);
 
     CompletableFuture<Response> future = new CompletableFuture<>();
-    relatedInstancesClient.getMany(query, future::complete);
+    relatedInstancesClient.getMany(query, Integer.MAX_VALUE, 0, future::complete);
```
julianladisch, line 78: The magic values `Integer.MAX_VALUE, 0` should be avoided by adding a method like getAll(String query, Consumer<Response>) to CollectionResourceClient. This also avoids duplication of this code in Instances.java.

Javadoc should be added to CollectionResourceClient.getMany pointing out the most APIs be default have a limit of 10 records.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -133,7 +134,7 @@ public AbstractInstances(final Storage storage, final HttpClient client) {
     CollectionResourceRepository precedingSucceedingTitlesRepository =
       new CollectionResourceRepository(precedingSucceedingTitlesClient);
 
-    List<String> instanceId = Arrays.asList(instance.getId());
+    List<String> instanceId = Collections.singletonList( instance.getId() );
```
julianladisch, line 137:
```suggestion
    List<String> instanceId = List.of(instance.getId());
```

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -536,7 +536,7 @@ private void getById(RoutingContext routingContext) {
     if (relatedInstancesClient != null) {
       CompletableFuture<Response> relatedInstancesFetched = new CompletableFuture<>();
 
-      relatedInstancesClient.getMany(query, relatedInstancesFetched::complete);
+      relatedInstancesClient.getMany( query, Integer.MAX_VALUE, 0, relatedInstancesFetched::complete);
```
julianladisch, line 539:
```suggestion
      relatedInstancesClient.getMany(query, Integer.MAX_VALUE, 0, relatedInstancesFetched::complete);
```

src/main/java/org/folio/inventory/resources/AbstractInstances.java
nielserik: Not that it matters to me, but Collections.singletonList was chosen to be consistent with the method used for this elsewhere in mod-inventory.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
nielserik: Would both of those recommendations apply to MultipleRecordsFetchClient.java, which also uses magic values and has no Javadoc at all.

nielserik:
> Could a test be added / changed that demonstrates this change fixes the issue?

Could there be a test added in MultipleRecordsFetchClient.java, which has the same fix for the limit?

marcjohnson-kint:
> > Could a test be added / changed that demonstrates this change fixes the issue?
> 
> Could there be a test added in MultipleRecordsFetchClient.java, which has the same fix for the limit?

If you would like to add a test to that area as well, that's ok with me. Or are you arguing that because that area does not have a test, this change does not have a test?

src/main/java/org/folio/inventory/resources/AbstractInstances.java
marcjohnson-kint: > The magic values `Integer.MAX_VALUE, 0` should be avoided by adding a method like getAll(String query, Consumer) to CollectionResourceClient. This also avoids duplication of this code in Instances.java.

I like the idea of creating a getAll method to make the intent to get all of the records explicit.

> Javadoc should be added to CollectionResourceClient.getMany pointing out the most APIs be default have a limit of 10 records.

I don't understand the use of Javadoc for this. I agree with the idea of a comment that explains why this uses a specific value for the limit to get all records.

nielserik:
> > > Could a test be added / changed that demonstrates this change fixes the issue?
> > 
> > 
> > Could there be a test added in MultipleRecordsFetchClient.java, which has the same fix for the limit?
> 
> If you would like to add a test to that area as well, that's ok with me.
> 
> Alternatively, are you arguing that because that area does not have a test, this change does not have a test?

I'm suggesting consistency. 

The commits in this PR use methods (Collections.singletonList), constants (nteger.MAX_VALUE), level of test coverage and Javadoc for setting instance relations in a single Instance, consistent with the methods, constants, test coverage and Javadoc that were applied when similarly setting instance relations in a set of Instances in 2020.  

For some reason it's met with a number of suggestions that all of that is done differently in this case. If the standard changed, maybe we need to raise an issue to align mod-inventory with the new standard across the board?

marcjohnson-kint:
> I'm suggesting consistency.

Consistency with what, the previous work done?

> The commits in this PR use methods (Collections.singletonList), constants (nteger.MAX_VALUE), level of test coverage and Javadoc for setting instance relations in a single Instance, consistent with the methods, constants, test coverage and Javadoc that were applied when similarly setting instance relations in a set of Instances in 2020.

> For some reason it's met with a number of suggestions that all of that is done differently in this case. 

I understand that you are frustrated by the quantity of feedback being offered when you were trying to help a team resolve a bug quickly.

The one thing I've asked for is a test that demonstrates the bug which can be used to check that the bug is fixed. I don't know if that is a FOLIO standard, it is how I suggest bug fixes are done.

My concern with not doing these kinds of things, because they weren't done before, is that leads to no incremental improvement (as the previous change can be used to justify not doing something).

> If the standard changed, maybe we need to raise an issue to align mod-inventory with the new standard across the board?

What standard are you referring to?

We could raise an issue to make wholesale changes. I'm happy if you want to do that. My experience is that those issues almost never get prioritised.

You are at liberty to ignore all of the feedback given on this pull request. As you are neither part of the Core Platform or Prokopovych teams, neither @julianladisch  or I have any authority over the changes you make.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
julianladisch: The underlying cause of the bug that this PR fixes is that many APIs have a default limit of 10.
Adding javadoc to getMany(String, Consumer) that warns about this default limit might prevent making this bug again.

julianladisch (approved):

marcjohnson-kint (commented):
I'm not really sure how to progress with this work

@julianladisch thank you for adding a test to demonstrate that the issue is resolved

I understand why you introduced the `getAll` method to avoid duplication of the maximum limit value.

I don't understand why the additional changes to CQL encoding have been included in this change. 

Please can you help me understand that better (especially as this is a bug fix that needs to be back-ported to 2021 R2).

PR #380, opened 2021-08-03
Exclude xerces:xmlParserAPIs from jar (org.xml.sax compile error)

julianladisch:
xerces:xmlParserAPIs is provided by JDK >= 9 and causes a compile error
(Eclipse doesn't suppress this compile error):
"The package org.xml.sax is accessible from more than one module: <unnamed>, java.xml"
https://stackoverflow.com/questions/55571046/eclipse-is-confused-by-imports-accessible-from-more-than-one-module


julianladisch:
Duplicate of #372

PR #386, opened 2021-09-01
MODINV-441 Remove commons-io dependency

AivarIUsupov:
Removing the commons-io dependency reduces the attack surface of the module and makes the life of the FOLIO security group easier.

### DIFF
--- a/src/main/java/org/folio/inventory/parsing/ModsParser.java
+++ b/src/main/java/org/folio/inventory/parsing/ModsParser.java
@@ -2,10 +2,10 @@ package org.folio.inventory.parsing;
 
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
-import org.apache.commons.io.IOUtils;
 import org.w3c.dom.Document;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
+import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 
 import javax.xml.parsers.DocumentBuilder;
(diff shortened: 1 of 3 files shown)


KaterynaSenchenko (approved):

marcjohnson-kint (approved):

src/main/java/org/folio/inventory/parsing/ModsParser.java
```diff
@@ -36,7 +37,7 @@ public ModsParser(CharacterEncoding characterEncoding) {
     DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
     factory.setNamespaceAware(false);
     DocumentBuilder builder = factory.newDocumentBuilder();
-    Document doc = builder.parse(IOUtils.toInputStream(xml, "UTF-8"));
+    Document doc = builder.parse(new InputSource(new StringReader(xml)));
```
SerhiiNosko, line 40: we need to think about closing resources, class StringReader has close method, do we need to invoke it?

AivarIUsupov: StringReader close javadoc
There's little value in closing a StringReader, since it doesn't hold an external resource (only Java-managed memory, not a file handle or native memory), so it's fine to omit it

SerhiiNosko: Thanks for checking

SerhiiNosko (approved):

Aliaksandr-Fedasiuk (approved):

VRohach (approved):

PR #390, opened 2021-09-10
MODINV-473 Support Holdings update viq QM flow

psmagin:
## Purpose
Support Holdings update via QM flow

## Approach
* update HoldingsUpdateDelegate
* update record in inventory based on RECORD_TYPE from payload


src/main/java/org/folio/inventory/dataimport/consumers/QuickMarcKafkaHandler.java
```diff
@@ -80,21 +88,57 @@ private void createProducer(KafkaConfig kafkaConfig, QMEventTypes eventType) {
...
+          .recover(throwable -> sendErrorEvent(params, eventPayload, throwable))
+          .map(ar -> record.key()),
+          th -> Future.failedFuture(th.getMessage())
         );
     }
     return Future.succeededFuture();
   }
 
+  private QMEventTypes getReplyEventType(Record.RecordType recordType) {
+    if (recordType == Record.RecordType.MARC_BIB) {
```
oleksandr-dekin, line 102: If we have` Record.RecordType.MARC_AUTHORITY` or `Record.RecordType.EDIFACT`, will be work normal for these types?
Maybe we should use `else if`

src/main/java/org/folio/inventory/dataimport/handlers/actions/HoldingsUpdateDelegate.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  public Future<HoldingsRecord> handle(Map<String, String> eventPayload, Record marcRecord, Context context) {
+    try {
+      JsonObject mappingRules = new JsonObject(eventPayload.get(MAPPING_RULES_KEY));
+      MappingParameters mappingParameters =
+        new JsonObject(eventPayload.get(MAPPING_PARAMS_KEY)).mapTo(MappingParameters.class);
+
+      JsonObject parsedRecord = retrieveParsedContent(marcRecord.getParsedRecord());
+      String holdingsId = marcRecord.getExternalIdsHolder().getHoldingsId();
```
oleksandr-dekin, line 45: Do we have such that the `getExternalIdsHolder` will be empty? Do we need to avoid NPE?

src/main/java/org/folio/inventory/dataimport/consumers/QuickMarcKafkaHandler.java
psmagin: quick marc doesn't work with authority or edifact records.

src/main/java/org/folio/inventory/dataimport/handlers/actions/HoldingsUpdateDelegate.java
psmagin: If it is empty, there is no chance to somehow handle this event. It should be there in any case. So, there is no difference between throwing NPE there or checking ExternalIdsHolder for NULL and throwing some other exception. Maybe it will be more clear. Tell me if I should handle this NULL-check.

oleksandr-dekin (approved):

PR #398, opened 2021-09-29
MODINV-498 Read-only fields for FOLIO holdings

Igor-Gorchakov:
**_Purpose_**
Create a list of read-only fields for FOLIO holdings. The list will be used to block the fields on UI application.

**_Approach_**
 - Create list of read-only fields for holdings (example InventoryConfigurationImpl) to return list of read only fields
 - Create separate endpoint: /inventory/config/holdings/blocked-fields
 - Update ModuleDescriptor-template.json
 - Write tests


src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -45,22 +35,38 @@
...
+import static org.folio.inventory.validation.InstancesValidators.refuseWhenHridChanged;
 
 public class Instances extends AbstractInstances {
   private static final String INSTANCES_CONTEXT_PATH = INSTANCES_PATH + "/context";
-  private static final String BLOCKED_FIELDS_CONFIG_PATH = INVENTORY_PATH + "/config/instances/blocked-fields";
-  private static final String BLOCKED_FIELDS_UPDATE_ERROR_MESSAGE = "Instance is controlled by MARC record, "
+  private static final String INSTANCE_BLOCKED_FIELDS_CONFIG_PATH = INVENTORY_PATH + "/config/instances/blocked-fields";
+  private static final String INSTANCE_BLOCKED_FIELDS_UPDATE_ERROR_MESSAGE = "Instance is controlled by MARC record, "
     + "these fields are blocked and can not be updated: ";
+  private static final String HOLDINGS_BLOCKED_FIELDS_CONFIG_PATH = INVENTORY_PATH + "/config/holdings/blocked-fields";
```
psmagin, line 68: It is strange to have holdings config in `Instances` class. Also, I don't see where these fields for holdings are really blocked the holdings update like it is done for instances: 
image

Igor-Gorchakov: Will do

psmagin (approved):

oleksandr-dekin (approved):

PR #410, opened 2021-10-19
MODINV-504 - Updated authority schema according to the new mapping rules

Igor-Gorchakov:
**_Purpose_**
The purpose is to update the schema accordingly to the mapping rules created in https://github.com/folio-org/mod-source-record-manager/pull/505/files
There is a need to simplify the schema to do not have complex objects (object with multiple fields) because the mapping processor can't seem to handle such mapping


MikhailFokanov (approved):
As we don't have an opportunity, to do it in clear way (with nested fields), this PR looks good to me.

KaterynaSenchenko (approved):

PR #424, opened 2021-11-15
MODINV-501 - Create Handler to generate Authorities from MARC Authority records

natali-zaitseva:
## Purpose
In the scope of handling MARC Authority creation via data-import we need to add a handler to map and create a corresponding Authority record and then send it to the mod-inventory-storage module.

## Approach

* added handler for MARC Authority records
* added external collection classes to work with inventory-storage
* added test
* updated raml-util to latest master version
* updated NEWS.md file


src/main/java/org/folio/inventory/dataimport/consumers/DataImportKafkaHandler.java
```diff
@@ -18,6 +18,7 @@
 import org.folio.inventory.dataimport.handlers.actions.CreateHoldingEventHandler;
 import org.folio.inventory.dataimport.handlers.actions.CreateInstanceEventHandler;
 import org.folio.inventory.dataimport.handlers.actions.CreateItemEventHandler;
+import org.folio.inventory.dataimport.handlers.actions.CreateMarcAuthoritiesEventHandler;
```
psmagin, line 21:
```suggestion
import org.folio.inventory.dataimport.handlers.actions.CreateMarcAuthorityEventHandler;
```
Maybe will be better to use singular?

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateMarcAuthoritiesEventHandler.java
```diff
@@ -0,0 +1,156 @@
...
+    return false;
+  }
+
+  private JsonObject prepareAuthority(DataImportEventPayload dataImportEventPayload) {
+    var authorityAsJson = new JsonObject(dataImportEventPayload.getContext().get(AUTHORITY.value()));
+    if (authorityAsJson.getJsonObject(AUTHORITY_PATH) != null) {
+      authorityAsJson = authorityAsJson.getJsonObject(AUTHORITY_PATH);
+    }
+    //Uncomment when
+//    authorityAsJson.put(SOURCE_KEY, MARC_FORMAT);
```
psmagin, line 145: When we should uncomment it? Do we have a story for this?

natali-zaitseva: yes MODINVSTOR-825, I will update the comment with ticket number

psmagin: But the ticket that you mentioned is for mod-inventory-storage. If we wait for the MODINVSTOR-825 competition, we also should have a mod-inventory ticket to uncomment this part of code.

natali-zaitseva: created an issue to not forgot about uncommenting this line https://issues.folio.org/browse/MODINV-576

psmagin (approved):

Igor-Gorchakov (approved):

oleksandr-dekin (approved):

RuslanLavrov (approved):

PR #428, opened 2021-11-24
Modinv-584: Improve logging for to be able to trace the path of each record and file_chunks mod-inventory

VRohach:
**Purpose**

We can change meaningless kafka header "correllationId" to "recordId" and "chunkId".

**Approach**
Improved log for handlers for showing ids.


src/main/java/org/folio/inventory/dataimport/consumers/DataImportKafkaHandler.java
```diff
@@ -61,6 +63,8 @@
 
   private static final Logger LOGGER = LogManager.getLogger(DataImportKafkaHandler.class);
   private static final String RECORD_ID_HEADER = "recordId";
+  private static final String CHUNK_ID_HEADER = "chunkId";
+  private static final String JOB_EXECUTION_ID_HEADER = "jobExecutionId";
```
SerhiiNosko, line 67: Here is job execution id header from lower case, and in MarcBibInstanceHridSetKafkaHandler its from upper case. In other classes the same, can we unify it in the same way(use the same case or extract as constant in some utility class)?

VRohach: @SerhiiNosko Yes, thank you! I missed to retrieve jobExecutionId from headers, not from eventPayload in DataImportKafkaHandler. I removed this constant from that handler.

Abdulkhakimov (approved):

Aliaksandr-Fedasiuk:
Could you review the ERROR level messages and add more parameters to them, if possible?

Aliaksandr-Fedasiuk (approved):

SerhiiNosko (approved):

AivarIUsupov (approved):

RuslanLavrov (approved):

PR #429, opened 2021-11-30
MODINV-583 Authority: Update metadata in schema

viacheslavkol:
MODINV-583 Authority: Update metadata in schema
https://issues.folio.org/browse/MODINV-583

### Purpose
Now the schema has a reference to metadata that is not convenient for mapping processor.

### Approach
- drop the reference to metadata in schema and add 4 new single properties - createdBy, createdDate, updatedBy, updatedDate
- update sample file
- add schema building to pom.xml


viacheslavkol:
It was decided not to change schema and to use regular metadata for Authority, thus closing this PR and deleting source branch.

PR #434, opened 2021-12-02
MODINV-580: Instance. Add new data element Administrative note

K-Felk:
Added support in the RAML files and in the mapping code.

Also added some code to one of the tests that creates a note 
and checks for its existence.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ## 18.x.x 2021-xx-xx
 
 * added Marc Authority handler for entities created via MARC Authority Data Import (MODINV-501)
+* added support for new administrative notes field (MODINV-580, MODINV-581, MODINV-582)
 
 ## 18.1.0 2021-10-18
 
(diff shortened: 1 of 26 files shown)


K-Felk:
Question on interface version:  the three issues that handle adminnotes in mod-inv each modify the json and should increment the interface version.  Do I need to increment it once of each issue?  In which case, it will probably end up being something like 11.4.  Or increment once for all three changes?  In which case it stays 11.2?

marcjohnson-kint:
> Question on interface version: the three issues that handle adminnotes in mod-inv each modify the json and should increment the interface version. Do I need to increment it once for each issue? In which case, it will probably end up being something like 11.4. Or increment once for all three changes? In which case it stays 11.2?

This is a grey area. It's likely that easier approach is acceptable.

As the work is being done separately, I think changing it for each change is the safest option (this is likely what I would do). 

If you would prefer not to do that, I think that is ok. Alternatively, all of the changes could be consolidated into a single pull request (with a single version change).

marcjohnson-kint (approved):

PR #435, opened 2021-12-03
MODINV-585: Pass on mod-inventory-storage HTTP 409 optimistic locking

julianladisch:


src/main/java/org/folio/inventory/support/http/server/ClientErrorResponse.java
```diff
@@ -26,4 +26,10 @@ public static void forbidden(HttpServerResponse response) {
     response.setStatusCode(403);
     response.end();
   }
+
+  public static void optimisticLocking(HttpServerResponse response, String reason) {
+    response.setStatusCode(409);
+    response.putHeader(HttpHeaders.CONTENT_TYPE, ContentType.TEXT_PLAIN);
+    response.end(reason);
+  }
 }
```
adamdickmeiss, line 35: Is the method only used in testing? If so, perhaps it should be in testing somehow.

adamdickmeiss (approved):

src/main/java/org/folio/inventory/support/http/server/ClientErrorResponse.java
julianladisch: Yes, it's only used in testing.
It parallels the other methods in ClientErrorResponse.
Moving optimisticLocking from ClientErrorResponse into ClientErrorResponseForTesting is not elegant and people will miss it if production code needs it when more APIs enable optimistic locking.

adamdickmeiss: Ok

PR #439, opened 2021-12-14
MODINV-599: Update Log4j to 2.16.0.

kaladay:
Resolve CVE-2021-44228 as per [MODINV-599](https://issues.folio.org/browse/MODINV-599).

Use 2.16.0 rather than 2.15.0, it contains additional changes related to the CVE and is also the version RMB is set to use as of [Release 33.1.3](https://github.com/folio-org/raml-module-builder/pull/1002).


marcjohnson-kint (approved):
Please add a news entry for this change (this should make back porting easier, as that will be back ported as well).

(I usually wouldn't advise this for a library version change on the mainline, however folks are paying attention to which releases have this change in it)

PR #440, opened 2021-12-14
MODINV-599: Update Log4j to 2.16.0 (Juniper).

kaladay:
Resolve CVE-2021-44228 as per [MODINV-599](https://issues.folio.org/browse/MODINV-599) for 17.0 version (Used by Juniper).

Use 2.16.0 rather than 2.15.0, it contains additional changes related to the CVE and is also the version RMB is set to use as of [Release 33.1.3](https://github.com/folio-org/raml-module-builder/pull/1002).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 17.0.8 2021-12-15
+
+* Update Log4j to 2.16.0. (CVE-2021-44228) (MODINV-599)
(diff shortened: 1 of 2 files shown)


kaladay:
Closing with the plans to re-create with the correct Ticket numbers.
Each release is using different tickets.

PR #446, opened 2021-12-20
MODINV-588 Initial changes for Instances deduplication with tests

AivarIUsupov:
## Purpose
Currently instance id equals to record id, that adds some confusion. Need to implement approach to generate different UUIDs for these entities and at the same to prevent duplicates processing.
The approach has been designed in scope of spike: [MODDATAIMP-500](https://issues.folio.org/browse/MODDATAIMP-500)  and needs the creation of new table for deduplication purposes. The working flow described in the diagrams below. Currently mod-inventory does not have a database schema and code to work with it, so need to add these capabilities in scope of current story. 

## Approach
- Added schema for mod-inventory. 
- Added liquibase for module's db version control.
- Added postgres client factory to work with db.
- Added dao/model/service layers with general solution for all further implementations.
- Covered new approach with tests (including embedded DB with schema initialization).

## Learning
[MODINV-588](https://issues.folio.org/browse/MODINV-588)


SerhiiNosko (approved):

NEWS.md
```diff
@@ -3,6 +3,7 @@
 * added Marc Authority handler for entities created via MARC Authority Data Import (MODINV-501)
 * added support for new administrative notes field (MODINV-580, MODINV-581, MODINV-582)
 * Update Log4j to 2.16.0. (CVE-2021-44228) (MODINV-599)
+* Implement deduplication for Instances (MODINV-588)
```
KaterynaSenchenko, line 6: please remove the news entry, for mod-inventory it is filled at the moment of release

KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/services/InstanceIdStorageService.java
```diff
@@ -0,0 +1,26 @@
...
+  public InstanceIdStorageService(EntityIdStorageDao entityIdStorageDao) {
+    this.entityIdStorageDao = entityIdStorageDao;
+  }
+
+  @Override
+  public Future<RecordToEntity> store(String recordId, String instanceId, String tenantId) {
+    RecordToEntity recordToInstance = RecordToEntity.builder().table(EntityTable.INSTANCE).recordId(recordId).entityId(instanceId).build();
+    LOGGER.info("Saving RecordToInstance relationship: {}", recordToInstance);
+    return entityIdStorageDao.saveRecordToEntityRelationship(recordToInstance, tenantId)
+      .map(recordToInstance);
```
RuslanLavrov, line 24: Hi @AivarIUsupov, propose to remove map() method call in order to return existing instance id from DB instead of new instanceId that we pass to the method in case of instance id already exists for specified recordId (duplicated event processing)
```suggestion
    return entityIdStorageDao.saveRecordToEntityRelationship(recordToInstance, tenantId);
```

NEWS.md
SerhiiNosko: Done

src/main/java/org/folio/inventory/services/InstanceIdStorageService.java
SerhiiNosko: Done

RuslanLavrov (approved):

Aliaksandr-Fedasiuk (approved):

PR #453, opened 2022-01-03
MODINV-589 Implement deduplication for Holdings

Shans-Kaluhin:
Currently mark holding id equals to record id, that adds some confusion. Need to implement approach to generate different UUIDs for just common inventory holdings and marc holdings, and at the same to prevent duplicates processing.

The approach has been designed in scope of spike: MODDATAIMP-500  and needs the creation of new table for deduplication purposes. This story need to be started after story: MODINV-588 because in MODINV-588 new DB schema for mod-inventory module should be created.   

Currently data-import supports 1 to 1 mapping between record and holding. In future when we need to support 1 to many mapping - table records_holdings can be extended with additional column that identifies particular holding such as sequence number. For marc holdings as option sequence number can be always equals to 1, because mark holdings will not support 1 to many, only inventory holdings will support.

The working flow described in the diagrams below. 
### Learn
https://issues.folio.org/browse/MODINV-589


src/main/java/org/folio/inventory/services/HoldingsIdStorageService.java
```diff
@@ -0,0 +1,25 @@
...
+
+  private final EntityIdStorageDao entityIdStorageDao;
+
+  public HoldingsIdStorageService(EntityIdStorageDao entityIdStorageDao) {
+    this.entityIdStorageDao = entityIdStorageDao;
+  }
+
+  @Override
+  public Future<RecordToEntity> store(String recordId, String instanceId, String tenantId) {
+    RecordToEntity recordToHoldings = RecordToEntity.builder().table(EntityTable.HOLDINGS).recordId(recordId).entityId(instanceId).build();
```
psmagin, line 21:
```suggestion
    RecordToEntity recordToHoldings = RecordToEntity.builder().table(EntityTable.HOLDINGS).recordId(recordId).entityId(holdingsId).build();
```

src/main/resources/liquibase/tenant/scripts/v-18.1.0/2022-01-01--13-00-create-records-holdings-table.xml
```diff
@@ -0,0 +1,18 @@
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.9.xsd">
+
+  <changeSet id="2021-12-10--13-00-create-records-instances-table" author="ShansKaluhin">
```
psmagin, line 7:
```suggestion
  <changeSet id="2021-12-10--13-00-create-records-holdings-table" author="ShansKaluhin">
```

psmagin (changes requested):

psmagin (approved):

SerhiiNosko (approved):

PR #454, opened 2022-01-04
MODINV-591 Implement deduplication for Authorities

Shans-Kaluhin:
Need to implement approach to generate different UUIDs for Marc Authorities, and at the same to prevent duplicates processing.

The approach has been designed in scope of spike: MODDATAIMP-500  and needs the creation of new table for deduplication purposes. This story need to be started after story: MODINV-588 because in MODINV-588 new DB schema for mod-inventory module should be created.   

Currently data-import supports 1 to 1 mapping between record and marc authority. In future when we need to support 1 to many mapping - table records_authorities can be extended with additional column that identifies particular authority such as sequence number.

The working flow described in the diagrams below. 

### Learn
https://issues.folio.org/browse/MODINV-591


src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateMarcAuthorityEventHandler.java
```diff
@@ -73,29 +78,36 @@ public CreateMarcAuthorityEventHandler(Storage storage, MappingMetadataCache map
...
-          future.complete(dataImportEventPayload);
-        })
-        .onFailure(e -> {
-          LOGGER.error("Failed to save new Authority by jobExecutionId: '{}' and recordId: '{}' and chunkId: '{}' ",jobExecutionId,
-            recordId, chunkId, e);
-          future.completeExceptionally(e);
+      String recordId = payloadContext.get(RECORD_ID_HEADER);
+      String chunkId = payloadContext.get(CHUNK_ID_HEADER);
+
+      Future<RecordToEntity> recordToAuthorityFuture = idStorageService.store(recordId, UUID.randomUUID().toString(), dataImportEventPayload.getTenant());
```
SerhiiNosko, line 84: As I remember authority UUID already generated on the earlier stage, so we need to extract it from payload and use it as parameter for IdStorageService instead of generating new UUID

psmagin: That's right. This should be fixed.

src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractAuthorityEventHandler.java
```diff
@@ -180,19 +182,19 @@ protected void prepareEvent(DataImportEventPayload payload) {
     payload.getContext().put(targetRecordType().value(), new JsonObject().encode());
   }
 
-  private String getChunkIdHeader(DataImportEventPayload payload) {
+  protected String getChunkIdHeader(DataImportEventPayload payload) {
     return payload.getContext() == null ? "-" : payload.getContext().get(CHUNK_ID_HEADER);
   }
 
-  private String getRecordIdHeader(DataImportEventPayload payload) {
+  protected String getRecordIdHeader(DataImportEventPayload payload) {
```
psmagin, line 189: Do we need the change of methods visibility?

Shans-Kaluhin: I use this methods in `CreateAuthorityEventHandler`.
image

I can implement deduplication in the `AbstractAuthorityEventHandler` but it will affect the `UpdateAuthorityEventHandler` and have to compare profileActions, like `ActionProfile.Action.CREATE == actionProfile()`

SerhiiNosko (approved):

psmagin (approved):

PR #474, opened 2022-02-11
MODINV-641: Can't provide instanceId to build hotlink for Item when trying to update Item and match by Item's field

VRohach:
**Overview**: The main issue of this ticket is that in cases when there is an update just only for Item-entity -> there is no information about an instance in UI. That's why hotlinks for "UPDATED" Item don't work.
**Approach**: 
If there is no data about Holding in the context at the ItemUpdateHandler, then the new code will go and retrieve a Holding by id (it exists in Item-entity). After that, this Holding(with instanceId) will be added to the payload for the build of all needed data for UI in the next steps of data-import.

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
@@ -31,6 +31,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.ActionProfile;
 import org.folio.DataImportEventPayload;
+import org.folio.MappingMetadataDto;
 import org.folio.dbschema.ObjectMapperTool;
 import org.folio.inventory.common.Context;
 import org.folio.inventory.common.api.request.PagingParameters;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
```diff
@@ -161,6 +170,27 @@ public UpdateItemEventHandler(Storage storage, MappingMetadataCache mappingMetad
     return future;
   }
 
+  private Future<DataImportEventPayload> addHoldingToPayloadIfNeeded(DataImportEventPayload dataImportEventPayload, Context context, Item updatedItem) {
+    Promise<DataImportEventPayload> promise = Promise.promise();
+    if (dataImportEventPayload.getContext().get(HOLDINGS.value()) == null || StringUtils.isBlank(dataImportEventPayload.getContext().get(HOLDINGS.value()))) {
```
SerhiiNosko, line 175: StringUtils.isBlank() also checks that input param is not equals to null, so its enought just to use this method

src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateItemEventHandler.java
```diff
@@ -161,6 +170,27 @@ public UpdateItemEventHandler(Storage storage, MappingMetadataCache mappingMetad
...
+          LOG.info("Successfully retrieved Holdings for the hotlink by id: {}", updatedItem.getHoldingId());
+          dataImportEventPayload.getContext().put(HOLDINGS.value(), Json.encodePrettily(success.getResult()));
+          promise.complete(dataImportEventPayload);
+        },
+        failure -> {
+          LOG.warn(format("Error retrieving Holdings for the hotlink by id %s cause %s, status code %s", updatedItem.getHoldingId(), failure.getReason(), failure.getStatusCode()));
+          promise.complete(dataImportEventPayload);
+        });
+    } else {
+      LOG.info("Holdings already exists in payload with for the hotlink");
```
SerhiiNosko, line 188: I think we can include holdings id to this log from updatedItem.getHoldingId()
and possibly make this log as debug but its up to you

Aliaksandr-Fedasiuk (approved):

Abdulkhakimov:
Add information about changes to NEWS.MD please

Abdulkhakimov (approved):

VRohach:
> Add information about changes to NEWS.MD please

Hi @Abdulkhakimov! As far as I know, we are never adding information to the mod-inventory module. (Because we are not code-owners)

Abdulkhakimov:
> > Add information about changes to NEWS.MD please
> 
> Hi @Abdulkhakimov! As far as I know, we are never adding information to the mod-inventory module. (Because we are not code-owners)

Forgot about it)

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #477, opened 2022-02-16
MODINV-645: Delete generate-documentation.sh files

julianladisch:
Running https://github.com/folio-org/mod-inventory/blob/master/generate-documentation.sh fails with this error:

```
Error: version validation
Unsupported RAML version: '#%RAML 1.0'
  in "../../ramls/inventory.raml", line 1, column 1:
    #%RAML 1.0
              ^
```

And reports security issues:

```
found 8 vulnerabilities (2 low, 2 moderate, 4 high) in 304 scanned packages
  1 vulnerability requires semver-major dependency updates.
  7 vulnerabilities require manual review. See the full report for details.
```

There is a new way how Jenkins generates the API documentation and publishes
them at https://dev.folio.org/reference/api/ therefore the files of
generate-documentation.sh can be deleted from the mod-inventory repository;
they are still in the git history.


marcjohnson-kint (commented):
@julianladisch I'm happy for these old scripts to be deleted. There is a lot of old stuff like this that needs to be removed e.g. the scripts for spinning up modules and registering them with Okapi.

PR #482, opened 2022-02-24
Fewer sockets in use for tests MODINV-650

adamdickmeiss:
This is a test-only fix. There seem to be no problem with
main usage of clients as various Verticle implementations
create one HttpClient. However, the code could be further be
simpfified by not createing them at all.. Just pass Vertx
which is already there in many cases. And let OkapiHttpClient
worry about WebClient creation.

### DIFF
--- a/src/main/java/org/folio/inventory/common/VertxAssistant.java
+++ b/src/main/java/org/folio/inventory/common/VertxAssistant.java
@@ -26,6 +26,10 @@ public class VertxAssistant {
     }
   }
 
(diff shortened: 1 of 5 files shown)


K-Felk (commented):
I pulled this and ran a clean install, and tests now pass.  Thank you!

marcjohnson-kint (commented):
@adamdickmeiss Thank you for making these changes.

I've ended up making similar changes in other modules due to how the tests execute and what that means for web client instances.

PR #485, opened 2022-03-07
MODDICORE-247 - Record matches are not decreased when additional match conditions are added to a job profile

RuslanLavrov:
## Purpose
Previously an error should be propagated when multiple inventory entities (instance etc) are matched.
Currently, it's necessary to process multiple match results by next sub-match profile if such exists

## Approach
* check whether current match profile contains a sub-match profile that linked for matches and populate matching result for further processing
* provide logic for entities loaders to process multiple matching results from the previous matching
* add tests

## Learning
[MODINV-652](https://issues.folio.org/browse/MODINV-652)

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
@@ -2,6 +2,7 @@ package org.folio.inventory.dataimport.handlers.matching.loaders;
 
 import io.vertx.core.Vertx;
 import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonArray;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.DataImportEventPayload;
@@ -14,15 +15,21 @@ import org.folio.processing.matching.loader.LoadResult;
 import org.folio.processing.matching.loader.MatchValueLoader;
 import org.folio.processing.matching.loader.query.LoadQuery;
 import org.folio.rest.jaxrs.model.EntityType;
+import org.folio.rest.jaxrs.model.ProfileSnapshotWrapper;
 
 import java.io.UnsupportedEncodingException;
+import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
+import static java.lang.String.format;
 import static org.folio.inventory.dataimport.handlers.matching.util.EventHandlingUtil.constructContext;
 
 public abstract class AbstractLoader<T> implements MatchValueLoader {
 
   private static final Logger LOG = LogManager.getLogger(AbstractLoader.class);
+
+  public static final String MULTI_MATCH_IDS = "MULTI_MATCH_IDS";
 
   private final Vertx vertx;
 
@@ -49,9 +56,16 @@ public abstract class AbstractLoader<T> implements MatchValueLoader {
             if (collection.totalRecords == 1) {
               loadResult.setValue(mapEntityToJsonString(collection.records.get(0)));
             } else if (collection.totalRecords > 1) {
-              String errorMessage = String.format("Found multiple records matching specified conditions. CQL query: [%s].\nFound records: %s", cql, Json.encodePrettily(collection.records));
-              LOG.error(errorMessage);
-              future.completeExceptionally(new MatchingException(errorMessage));
+              if (canProcessMultiMatchResult(eventPayload)) {
+                LOG.info("Found multiple records by CQL query: [{}]. Found records IDs: {}", cql, mapEntityListToIdsJson(collection.records));
+                loadResult.setEntityType(MULTI_MATCH_IDS);
+                loadResult.setValue(mapEntityListToIdsJson(collection.records));
+              } else {
+                String errorMessage = String.format("Found multiple records matching specified conditions. CQL query: [%s].%nFound records: %s", cql, Json.encodePrettily(collection.records));
+                LOG.error(errorMessage);
+                future.completeExceptionally(new MatchingException(errorMessage));
+                return;
+              }
             }
             future.complete(loadResult);
           },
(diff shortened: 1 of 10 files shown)


SerhiiNosko (approved):

AivarIUsupov (approved):

Abdulkhakimov (approved):

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
```diff
@@ -49,9 +59,16 @@ public AbstractLoader(Vertx vertx) {
...
             } else if (collection.totalRecords > 1) {
-              String errorMessage = String.format("Found multiple records matching specified conditions. CQL query: [%s].\nFound records: %s", cql, Json.encodePrettily(collection.records));
-              LOG.error(errorMessage);
-              future.completeExceptionally(new MatchingException(errorMessage));
+              if (canProcessMultiMatchResult(eventPayload)) {
+                LOG.info("Found multiple records by CQL query: [{}]. Found records IDs: {}", cql, mapEntityListToIdsJson(collection.records));
+                loadResult.setEntityType(MULTI_MATCH_IDS);
+                loadResult.setValue(mapEntityListToIdsJson(collection.records));
+              } else {
+                String errorMessage = String.format("Found multiple records matching specified conditions. CQL query: [%s].%nFound records: %s", cql, Json.encodePrettily(collection.records));
```
okolawole-ebsco, line 67: General question, do we care if customer information gets posted to logs?

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
```diff
@@ -68,16 +85,33 @@ public AbstractLoader(Vertx vertx) {
...
     return getEntityType() == entityType;
   }
 
+  protected String getConditionByMultiMatchResult(DataImportEventPayload eventPayload) {
+    String preparedIds = new JsonArray(eventPayload.getContext().remove(MULTI_MATCH_IDS))
+      .stream()
+      .map(Object::toString)
+      .collect(Collectors.joining(" OR "));
+
+    return format(" AND id == (%s)", preparedIds);
```
okolawole-ebsco, line 105: it appears that this method is constructing some SQL is that correct? Thinking this should originate in some other class that deals with SQL statements if it does.

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AuthorityLoader.java
```diff
@@ -52,4 +59,13 @@ protected String addCqlSubMatchCondition(DataImportEventPayload eventPayload) {
   protected String mapEntityToJsonString(Authority authority) {
     return Json.encode(authority);
   }
+
+  @Override
+  protected String mapEntityListToIdsJson(List<Authority> authorityList) {
```
okolawole-ebsco, line 64: General information.

It would be clearer if the method returned something like a `JsonObject` or similar so that it is enforced that a correct JSON object is returned. Since it is just a string, it could be anything! You are overriding a method so i dont expect any change to the API!

src/test/java/org/folio/inventory/eventhandlers/MatchAuthorityEventHandlerUnitTest.java
```diff
@@ -458,6 +466,93 @@ public void shouldMatchWithSubMatchByAuthorityOnHandleEventPayload(TestContext t
     });
   }
 
+  @Test
+  public void shouldMatchWithSubConditionBasedOnMultiMatchResultOnHandleEventPayload(TestContext testContext) throws UnsupportedEncodingException {
+    Async async = testContext.async();
+    List<String> multiMatchResult = List.of(
+      "bd39d7cc-f313-47ea-bf2d-abcac2b24a64", "e6dc8015-fcad-40cf-afa9-e817a605dc06");
```
okolawole-ebsco, line 473: Does the values of this list have to be the UUID listed? Could it be any UUID?

okolawole-ebsco (commented):
I mostly had general info!

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
RuslanLavrov: the records don't contain sensitive data, so suppose they can be posted

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
```diff
@@ -68,16 +85,33 @@ public AbstractLoader(Vertx vertx) {
     return future;
   }
 
+  private boolean canProcessMultiMatchResult(DataImportEventPayload eventPayload) {
+    List<ProfileSnapshotWrapper> childProfiles = eventPayload.getCurrentNode().getChildSnapshotWrappers();
```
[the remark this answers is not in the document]
RuslanLavrov: A child snapshot is ProfileSnapshotWrapper type. A profile snapshot represents m-ary tree of profiles of different types which can be linked to each other in a different order. ProfileSnapshotWrapper represents a node of tree. The `getChildSnapshotWrappers()` is api of ProfileSnapshotWrapper class and it returns child nodes which are ProfileSnapshotWrapper as well.  Also, ProfileSnapshotWrapper contains data, for example, relation type (for matches/for non-matches) which can not be stored in a profile itself and can be different depending on a tree where particular profile is used.

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
RuslanLavrov: this method returns CQL condition (not SQL) that is used by rest api of mod-inventory-storage

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AuthorityLoader.java
RuslanLavrov: agree, definitely method declaration with JsonObject would enforce to return correct json string representation. Just, in fact, the method is used only to get json string representation, so to avoid unused converting to JsonObject and then deserialization to string a less strict return type signature is used as a compromise.

src/test/java/org/folio/inventory/eventhandlers/MatchAuthorityEventHandlerUnitTest.java
RuslanLavrov: they can be any UUID. Updated test by random UUID to avoid confusion regarding the purpose of these values.

PR #487, opened 2022-03-12
MODDATAIMP-648 Show authority log after update

ALiemeshka:
### Purpose
Don't show authority log after update

### Approach
Run DI_INVENTORY_AUTHORITY_UPDATED even manualy

### Learning
[MODDATAIMP-648](https://issues.folio.org/browse/MODDATAIMP-648)

### DIFF
--- a/src/test/java/org/folio/inventory/dataimport/handlers/actions/UpdateAuthorityEventHandlerTest.java
+++ b/src/test/java/org/folio/inventory/dataimport/handlers/actions/UpdateAuthorityEventHandlerTest.java
@@ -42,6 +42,7 @@ import com.github.tomakehurst.wiremock.matching.UrlPathPattern;
 import io.vertx.core.Vertx;
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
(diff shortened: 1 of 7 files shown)


src/test/java/org/folio/inventory/dataimport/handlers/actions/UpdateAuthorityEventHandlerTest.java
```diff
@@ -114,7 +113,7 @@ public void setUp() throws IOException {
     MockitoAnnotations.openMocks(this);
     MappingManager.clearReaderFactories();
     MappingMetadataCache mappingMetadataCache = new MappingMetadataCache(vertx, vertx.createHttpClient(), 3600);
-    eventHandler = new UpdateAuthorityEventHandler(storage, mappingMetadataCache);
+    eventHandler = new UpdateAuthorityEventHandler(storage, mappingMetadataCache, publisher);
```
psmagin, line 116: You should check in test that event was really sent

psmagin (changes requested):

PR #491, opened 2022-03-16
MODINV-652 - Record matches are not decreased when additional match conditions are added to a job profile

RuslanLavrov:
## Purpose
It is necessary to expand the limit for records loading during a match profile processing. Currently a default limit value is used limit=10, that may not be enough for further processing multiple matching results by next linked match profile

## Learning
[MODINV-652](https://issues.folio.org/browse/MODINV-652)

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
@@ -34,6 +34,7 @@ public abstract class AbstractLoader<T> implements MatchValueLoader {
 
   public static final String MULTI_MATCH_IDS = "MULTI_MATCH_IDS";
   private static final String ERROR_LOAD_MSG = "Failed to load records cause: %s, status code: %s";
+  private static final int MULTI_MATCH_LOAD_LIMIT = 90;
 
   private final Vertx vertx;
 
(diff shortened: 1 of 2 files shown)


Abdulkhakimov (approved):

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
```diff
@@ -85,6 +88,21 @@ protected AbstractLoader(Vertx vertx) {
...
+   * Otherwise, for performance needs returns paging parameters with limit = 2, which is
+   * a minimum value that is necessary to get target record or identify whether multiple match result occurred.
+   *
+   * @param multiMatchResultParams - identifies whether to return paging parameters for multiple matching
+   * @return {@link PagingParameters}
+   */
+  private PagingParameters buildPagingParameters(boolean multiMatchResultParams) {
+    // currently, limit = 90 is used because of constraint for URL size that is used for processing multi-match result
+    // in scope of https://issues.folio.org/browse/MODDICORE-251 a new approach will be introduced for multi-matching result processing
+    return multiMatchResultParams ? new PagingParameters(90, 0) : new PagingParameters(2, 0);
```
VRohach, line 103: @RuslanLavrov I think it would be better to extract this value (90) to the specific constant via a logic-based name. What do you think?

VRohach (approved):

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/AbstractLoader.java
RuslanLavrov: Hi @VRohach, moved to constant

AivarIUsupov (approved):

PR #504, opened 2022-04-13
MODINV-678: The status of SRS MARC record is not updated after uploading MARC file with non-existing instanceId in 999ff$i field.

VRohach:
1. Error catching in cases when Instance+Holdings was NOT found added.

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/HoldingsUpdateDelegate.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/HoldingsUpdateDelegate.java
@@ -81,7 +81,7 @@ public class HoldingsUpdateDelegate {
     Promise<HoldingsRecord> promise = Promise.promise();
     holdingsRecordCollection.findById(holdingsId, success -> {
         if (success.getResult() == null) {
-          LOGGER.error(format("Can't find Holdings by id: %s ", holdingsId));
+          LOGGER.error("Can't find Holdings by id: {} ", holdingsId);
           promise.fail(new NotFoundException(format("Can't find Holdings by id: %s ", holdingsId)));
         } else {
           promise.complete(success.getResult());
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/HoldingsUpdateDelegate.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/HoldingsUpdateDelegate.java
@@ -17,6 +17,7 @@ import org.folio.HoldingsRecord;
 import org.folio.inventory.common.Context;
 import org.folio.inventory.dataimport.exceptions.OptimisticLockingException;
 import org.folio.inventory.domain.HoldingsRecordCollection;
+import org.folio.inventory.exceptions.NotFoundException;
 import org.folio.inventory.services.HoldingsCollectionService;
 import org.folio.inventory.storage.Storage;
 import org.folio.processing.mapping.defaultmapper.RecordMapper;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/actions/InstanceUpdateDelegate.java
```diff
@@ -73,7 +74,15 @@ private JsonObject retrieveParsedContent(ParsedRecord parsedRecord) {
 
   private Future<Instance> getInstanceById(String instanceId, InstanceCollection instanceCollection) {
     Promise<Instance> promise = Promise.promise();
-    instanceCollection.findById(instanceId, success -> promise.complete(success.getResult()),
+    instanceCollection.findById(instanceId, success -> {
+        if (success.getResult() == null) {
+          String errorMsg = format("Can't find Instance by id: %s ", instanceId);
+          LOGGER.error(errorMsg);
```
Abdulkhakimov, line 80: Maybe it would be better to use formating directly in Logger without setting new variable?

src/main/java/org/folio/inventory/dataimport/handlers/actions/InstanceUpdateDelegate.java
```diff
@@ -73,7 +74,15 @@ private JsonObject retrieveParsedContent(ParsedRecord parsedRecord) {
...
+          promise.fail(new NotFoundException(errorMsg));
+        } else {
+          promise.complete(success.getResult());
+        }
+      },
```
SerhiiNosko, line 85: @VRohach, what regarding holdings, items, authorities, do they already have this fix, or its not applicable for them?

src/main/java/org/folio/inventory/dataimport/handlers/actions/InstanceUpdateDelegate.java
VRohach: @Abdulkhakimov Done.

src/main/java/org/folio/inventory/dataimport/handlers/actions/InstanceUpdateDelegate.java
VRohach: @SerhiiNosko Thank you! Find and added to the HoldingsUpdateDelegate.

Abdulkhakimov (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/HoldingsUpdateDelegate.java
```diff
@@ -78,7 +79,14 @@ private JsonObject retrieveParsedContent(ParsedRecord parsedRecord) {
   private Future<HoldingsRecord> getHoldingsRecordById(String holdingsId,
                                                        HoldingsRecordCollection holdingsRecordCollection) {
     Promise<HoldingsRecord> promise = Promise.promise();
-    holdingsRecordCollection.findById(holdingsId, success -> promise.complete(success.getResult()),
+    holdingsRecordCollection.findById(holdingsId, success -> {
+        if (success.getResult() == null) {
+          LOGGER.error(format("Can't find Holdings by id: %s ", holdingsId));
```
SerhiiNosko, line 84: @VRohach for loggers we can rewrite like this: LOGGER.error("Can't find Holdings by id: {} ", holdingsId));

psmagin (approved):

SerhiiNosko (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/HoldingsUpdateDelegate.java
okolawole-ebsco, line 84: nitpick: Maybe you could create the exception first, then log it, fail the promise with it. This would avoid declaring this string more than once.

VRohach: @SerhiiNosko Changed.

VRohach: @okolawole-ebsco thank you! I changed it via @SerhiiNosko  comment. I think that this approach will be readable.

RuslanLavrov (approved):

PR #506, opened 2022-04-14
[MODINV-616] Afford related resources

wwelling:


src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -603,7 +694,7 @@ private CompletableFuture<Instance> fetchPrecedingSucceedingTitles(
   // Utilities
 
   private CqlQuery cqlMatchAnyByInstanceIds(List<String> instanceIds) {
-    return CqlQuery.exactMatchAny("instanceId", instanceIds);
+    return CqlQuery.exactMatchAny(RelatedInstance.INSTANCE_ID_KEY, instanceIds);
```
marcjohnson-kint, line 697: Whilst I understand the value is the same, I think that is by coincidence and the reuse of the constant could be confusing for folks, especially given that these concepts are already easy to mix up (given their similar names and structure)

(same for preceding / succeeding titles)

src/main/java/org/folio/inventory/domain/instances/RelatedInstance.java
```diff
@@ -0,0 +1,64 @@
...
+      return object.toString().equals(this.toString());
+    } else {
+      return false;
+    }
+  }
+
+  @Override
+  public int hashCode() {
+    return toString().hashCode();
+  }
```
marcjohnson-kint, line 47: What are these used for?

src/main/java/org/folio/inventory/resources/InstancesResponse.java
```diff
@@ -30,6 +32,17 @@ public boolean hasRecords () {
...
+  public Map<String, List<RelatedInstance>> getRelatedInstanceMap() {
+    return Collections.unmodifiableMap(relatedInstanceMap);
+  }
+
+  public InstancesResponse setRelatedInstanceMap(
+    Map<String, List<RelatedInstance>> relatedInstanceMap) {
+
+    this.relatedInstanceMap = relatedInstanceMap;
+    return this;
+  }
```
marcjohnson-kint, line 44: I'm glad this is currently a single map. I understand it might be necessary to have the parent / child distinction, like instance relationships, when it comes to matching records.

marcjohnson-kint (commented):
As requested, I've given this a quick review of the approach.

I think the approach fits with my understanding of the desired domain (as confusing as it is) and with how the other related ideas have been implemented.

Thank you for renaming the variables / parameters for the instance relationships code to disambiguate them from related instances.

I share your concerns with the quantity of code needed to accomplish this and with the amount of duplication across these related (no pun intended) ideas. I think it is a good candidate for technical debt in the module.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -56,17 +57,25 @@ public AbstractInstances(final Storage storage, final HttpClient client) {
     this.config = new InventoryConfigurationImpl();
   }
 
+  protected CompletableFuture<List<Response>> updateRelatedRecords(
+    RoutingContext routingContext, WebContext context, Instance instance) {
+
+    return updateRelatedInstances(instance, routingContext, context)
+      .thenCompose(r -> updateInstanceRelationships(instance, routingContext, context))
+      .thenCompose(r -> updatePrecedingSucceedingTitles(instance, routingContext, context));
+  }
```
marcjohnson-kint, line 66: Thank you for extracting this out, the future chains were getting a little unwieldy.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -76,12 +85,66 @@ protected CompletableFuture<List<Response>> updateInstanceRelationships(Instance
...
+    Map<String, RelatedInstance> existingRelatedInstances = new HashMap<>();
+
+    relatedInstancesList.stream().map(RelatedInstance::new).forEachOrdered(related ->
+      existingRelatedInstances.put(related.id, related)
+    );
+
+    Map<String, RelatedInstance> updatingRelatedInstances = new HashMap<>();
+    if (instance.getRelatedInstances() != null) {
+      instance.getRelatedInstances().forEach(updating -> {
+        String id = (updating.id == null ? UUID.randomUUID().toString() : updating.id);
```
marcjohnson-kint, line 106: I'm not sure this will work for related instances in the current form. 

At the moment the `id` of the relationship isn't in the business logic API representation. If the client abides by that, then the `id` received from the client will always be null. Which might mean that a new related instance record is created.

This might be why the instance relationships representation exposes an ID (which unfortunately, has some negative impacts too (refer to [MODINV-674](https://issues.folio.org/browse/MODINV-674?focusedCommentId=128833&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-128833) for more details).

src/test/java/api/RelatedInstanceTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+    var firstNodJson = nod(instanceId)
+        .put(Instance.TITLE_KEY, "An instance");
+
+    instancesClient.create(firstNodJson);
+
+    var thirdNodJson = nod(relatedInstanceId)
+        .put(Instance.TITLE_KEY, "A related instance")
+        .put(Instance.RELATED_INSTANCES_KEY, new JsonArray()
+            .add(createRelatedInstance(UUID.randomUUID(), relatedInstanceId, instanceId, relatedInstanceTypeId)));
```
marcjohnson-kint, line 77: This sends an `id` property which currently isn't defined in the API documentation

src/main/java/org/folio/inventory/resources/Instances.java
wwelling: Would you like I switch back to using string literals? There is one class in which causes a code smell due to the duplication of string literals.

src/main/java/org/folio/inventory/domain/instances/RelatedInstance.java
wwelling: Not sure if required at all. It was following the pattern from [InstanceRelationship](https://github.com/folio-org/mod-inventory/blob/master/src/main/java/org/folio/inventory/domain/instances/InstanceRelationship.java). Will gladly remove unless comparison is required of these RelatedInstance.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
wwelling: How should we address this issue? Should it also be addressed for instance relationships at this time?

src/test/java/api/RelatedInstanceTest.java
wwelling: Yes, there is a bit of mismatch between the Instance schema for relatedInstances and what is actual. Will require multiple representations of related instance; one that is placed in relatedInstances on the Instance and another for the entity retrieved from mod-inventory-storage related-instances endpoint.

I am concerned of this difference and not sure what is the best approach to handle. Keep in mind there is already a slight mutation of the related instances placed on the instance response in which the instanceId and relatedInstanceId are swapped is not matching the instance id in the expected "directionality".

src/main/java/org/folio/inventory/resources/Instances.java
marcjohnson-kint: > Would you like I switch back to using string literals? There is one class in which causes a code smell due to the duplication of string literals.

I think this is an issue with static analysis.

I think it's ok to use a constant. I would prefer the constant was localised to that class and name more generally than specifically for related instances.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
marcjohnson-kint: > How should we address this issue? 

If it's possible to hide the `id` of the related instance records (which are an artefact of the storage / business logic split rather than the domain) then I think we should.

This will mean the logic needed to detect how a related instance translates to storage will need to be different from instance relationships.

> Should it also be addressed for instance relationships at this time?

I don't think so. This would be a breaking change and folks haven't decided what to do about this. I don't think we should mix these two pieces of work together.

src/test/java/api/RelatedInstanceTest.java
marcjohnson-kint: > Yes, there is a bit of mismatch between the Instance schema for relatedInstances and what is actual. 

What do you mean by actual? Is it the storage representation of related instances?

> Will require multiple representations of related instance; one that is placed in relatedInstances on the Instance and another for the entity retrieved from mod-inventory-storage related-instances endpoint.

Yes, I think that is inherent in trying to hide the artefacts of 

> I am concerned of this difference and not sure what is the best approach to handle.

What concerns do you have about the difference?

I think there needs to be a mapping. I would expect there is a mapping in instance relationships. Am I wrong about that? Maybe that is why the `id` is exposed and the records are split between parent and child.

> Keep in mind there is already a slight mutation of the related instances placed on the instance response in which the instanceId and relatedInstanceId are swapped is not matching the instance id in the expected "directionality".

I understand that. That is the nature of the domain model and the way the storage / business logic split is achieved.

I don't know of a way to avoid it.

ramls/instance.json
```diff
@@ -27,6 +27,36 @@
...
+          "relatedInstanceTypeId": {
+            "description": "Id of the related instance type",
+            "type": "string"
+          }
+        }, 
+        "additionalProperties": false,
+        "required": [
+          "relatedInstanceId",
+          "relatedInstanceTypeId"
+        ]
```
marcjohnson-kint, line 57: Should instance ID also be required?

As I understand it, both are needed for a related instance to make sense. Since we are exposing that aspect and not using the host instance's ID, I think we need it to be provided here.

src/test/java/api/RelatedInstanceTest.java
```diff
@@ -0,0 +1,308 @@
...
+  public void testRelatedInstance() {
+    String id = UUID.randomUUID().toString();
+    String instanceId = UUID.randomUUID().toString();
+    String relatedInstanceId = UUID.randomUUID().toString();
+    String relatedInstanceTypeId = UUID.randomUUID().toString();
+    RelatedInstance relatedInstance = new RelatedInstance(
+        id,
+        instanceId,
+        relatedInstanceId,
+        relatedInstanceTypeId);
```
marcjohnson-kint, line 35: Why are the parameters formatted this way?

src/test/java/api/RelatedInstanceTest.java
```diff
@@ -0,0 +1,308 @@
...
+import org.junit.Test;
+
+import api.support.ApiTests;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+public class RelatedInstanceTest extends ApiTests {
+
+  @Test
+  public void testRelatedInstance() {
```
marcjohnson-kint, line 26: What does this test about a related instance?

src/test/java/api/RelatedInstanceTest.java
```diff
@@ -0,0 +1,308 @@
...
+
+        assertThat(secondInstanceSecondRelatedInstanceJson.getString(RelatedInstance.RELATED_INSTANCE_ID_KEY),
+            is(anotherRelatedInstanceId.toString()));
+    } else {
+        assertThat(secondInstanceFirstRelatedInstanceJson.getString(RelatedInstance.RELATED_INSTANCE_ID_KEY),
+            is(anotherRelatedInstanceId.toString()));
+
+        assertThat(secondInstanceSecondRelatedInstanceJson.getString(RelatedInstance.RELATED_INSTANCE_ID_KEY),
+            is(instanceId.toString()));
+    }
```
marcjohnson-kint, line 197: Why is this branching needed?

Is it to match up multiple related instance to the correct expectations?

src/test/java/api/RelatedInstanceTest.java
```diff
@@ -0,0 +1,308 @@
...
+
+import org.folio.inventory.domain.instances.Instance;
+import org.folio.inventory.domain.instances.RelatedInstance;
+import org.junit.Test;
+
+import api.support.ApiTests;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+public class RelatedInstanceTest extends ApiTests {
```
marcjohnson-kint, line 23: I think there also needs to be a test for when a related instance is removed.

src/test/java/api/RelatedInstanceTest.java
```diff
@@ -0,0 +1,308 @@
...
+    assertEquals(relatedInstanceTypeId, inverseRelatedInstance.relatedInstanceTypeId);
+
+    assertEquals(relatedInstance.id, inverseRelatedInstance.id);
+    assertEquals(relatedInstance.instanceId, inverseRelatedInstance.relatedInstanceId);
+    assertEquals(relatedInstance.relatedInstanceId, inverseRelatedInstance.instanceId);
+    assertEquals(relatedInstance.relatedInstanceTypeId, inverseRelatedInstance.relatedInstanceTypeId);
+  }
+
+  @Test
+  public void canCreateInstanceWithRelatedInstances() throws Exception {
```
marcjohnson-kint, line 65: Personally, I would use the sneaky throws trait rather than indicate this method throws a general exception (which provides almost no information to the reader)

Ideally, the sneaky throws trait could be put on only the methods invoked in the test that throw checked exceptions. That way other tests can benefit from this as well.

```suggestion
  @SneakyThrows
  public void canCreateInstanceWithRelatedInstances() {
```

marcjohnson-kint (commented):
Given the concerns raised about the implementation logic during our call, I'm not going to review that thoroughly as I am assuming that it mostly mimics that of the instance relationships except with a single array.

I think it could be good if the tests could be reworked a bit to remove the branching statements and some of the noise of handling HTTP and JSON where possible.

ramls/instance.json
wwelling: Yes it should. Thank you.

src/test/java/api/RelatedInstanceTest.java
wwelling: I am not sure. What is the appropriate style?

src/test/java/api/RelatedInstanceTest.java
wwelling: Tests the static method that swap instanceId and relatedInstanceId appropriately.

src/test/java/api/RelatedInstanceTest.java
wwelling: Because random UUID will be in a random order in the list.

src/test/java/api/RelatedInstanceTest.java
wwelling: Thanks. Will add.

src/test/java/api/RelatedInstanceTest.java
wwelling: Nice. Learn something new every day. Thanks.

src/test/java/api/RelatedInstanceTest.java
wwelling: Added.

src/test/java/api/RelatedInstanceTest.java
wwelling: I have tried ordering the collection before setting the list on the instance, but it appears serialization is causing reorder.

ramls/instance.json
```diff
@@ -27,6 +27,38 @@
...
+            "type": "string"
+          },
+          "relatedInstanceTypeId": {
+            "description": "Id of the related instance type",
+            "type": "string"
+          }
+        }, 
+        "additionalProperties": false,
+        "required": [
+          "id",
```
marcjohnson-kint, line 55: I'm not sure if `id` can be required. When a related instance is new, I don't think the UI (or client) will generate the `id`.

src/test/java/api/RelatedInstanceTest.java
```diff
@@ -0,0 +1,342 @@
...
+
+import api.support.ApiTests;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import lombok.SneakyThrows;
+
+public class RelatedInstanceTest extends ApiTests {
+
+  @Test
+  public void testRelatedInstanceFrom() {
```
marcjohnson-kint, line 27: I think this is a unit test for mapping related instances from JSON to a instance of a class.

Is that correct?

If so, I think that should be in a separate class (in a namespace applicable to the code that has the mapping code) to the API tests.

marcjohnson-kint (approved):
Following our earlier conversation, I think we defer an further refinement / improvement of the readability of the API tests to a separate pull request / activity

PR #507, opened 2022-04-15
MODINV-577 Match logic for VRN matches to Inventory records

viacheslavkol:
## Purpose
To implement the matching logic invoked when a user selects Vendor reference number (VRN) as the matchpoint from an incoming MARC Bibliographic record to an existing Inventory Instance, Holdings, or Item
## Approach
- add preloading field
- extend OrdersPreloaderHelper to support VRN matches
- implement unit tests

## Learning
[MODINV-577](https://issues.folio.org/browse/MODINV-577)

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/matching/preloaders/OrdersPreloaderHelper.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/matching/preloaders/OrdersPreloaderHelper.java
@@ -9,6 +9,7 @@ import static org.folio.inventory.support.CqlHelper.buildMultipleValuesCqlQuery;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import io.vertx.core.json.JsonArray;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/matching/preloaders/OrdersPreloaderHelper.java
```diff
@@ -19,6 +20,9 @@
 
 public class OrdersPreloaderHelper {
 
+    private static final String ORDER_LINES_CQL_PATTERN = "purchaseOrder.workflowStatus==Open AND %s";
+    private static final String VRN_VALUE_CQL_PATTERN = "\"\\\"refNumber\\\":\\\"%s\\\"\"";
```
psmagin, line 24: In documentation is specified another way to query this type of values: https://github.com/folio-org/raml-module-builder/blob/master/README.md#cql-matching-array-elements

`property =/@type1 value`

order-storage has the same structure in schema.json 
image


please test this type of query

psmagin (approved):

Shans-Kaluhin (approved):

DieHardJava (approved):

PR #508, opened 2022-04-22
MODINV-684: Fix "_version" is not mandatory field

Aliaksandr-Fedasiuk:
## Purpose
OL field "_version" is mandatory only for cases when OL is enabled.

## Approach
Added a condition to check this case

## Learning
[MODINV-684](https://issues.folio.org/browse/MODINV-684)

### DIFF
--- a/src/main/java/org/folio/inventory/support/InstanceUtil.java
+++ b/src/main/java/org/folio/inventory/support/InstanceUtil.java
@@ -1,8 +1,7 @@
 package org.folio.inventory.support;
 
-import java.util.ArrayList;
-import java.util.List;
-
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
 import org.folio.ChildInstance;
 import org.folio.ParentInstance;
 import org.folio.Tags;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/inventory/support/InstanceUtil.java
```diff
@@ -100,4 +99,16 @@ public static JsonObject mergeInstances(JsonObject existing, JsonObject mapped)
...
     return mergedInstanceAsJson;
   }
+
+  /**
+   * Returns the value if s can be parsed as Integer. Returns null if s is null.
+   *
+   * @throws NumberFormatException if s is not null and cannot be parsed as Integer.
+   */
+  private static Integer asIntegerOrNull(String s) {
+    if (s == null) {
```
SerhiiNosko, line 109: Better to use StringUtils.isBlank() here

Abdulkhakimov: We can rewrite method as bellow: 
return StringUtils.isBlank(s) ? null : Integer.parseInt(s)

Aliaksandr-Fedasiuk: I think checking on "null" is enough because this field will be added automatically by trigger and we do not need several additional checks from StringUtils when we need only to check if this field is present or not.

Aliaksandr-Fedasiuk: > We can rewrite method as bellow: return StringUtils.isBlank(s) ? null : Integer.parseInt(s)
For me is not a good decision because I need to add a comment to this check such as I added to the method.

okolawole-ebsco (approved):

PR #509, opened 2022-04-22
MODSOURMAN-782. Upgrade Kafka version

SerhiiNosko:
Make kafka-clients version consistent as in mod-data-import, mod-source-record-manager.
Currently mod-inventory uses 2.6 kafka version from transitive dependency, when:
mod-data-import uses 3.1.0 https://github.com/folio-org/mod-data-import/blob/master/pom.xml#L28
mod-source-record-manager uses 3.1.0: https://github.com/folio-org/mod-source-record-manager/blob/master/mod-source-record-manager-server/pom.xml#L19

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,6 +39,11 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>${kafkaclients.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.testcontainers</groupId>
       <artifactId>postgresql</artifactId>
       <version>${testcontainers.version}</version>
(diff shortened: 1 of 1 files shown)


Aliaksandr-Fedasiuk (approved):

pom.xml
```diff
@@ -255,11 +260,12 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <testcontainers.version>1.15.3</testcontainers.version>
-    <vertx.version>4.2.5</vertx.version>
```
okolawole-ebsco, line 258: Is this version compatible with RMB? Should we update RMB to the version that depends on 4.2.6?

SerhiiNosko: Hi @okolawole-ebsco , we don't have RMB here in mod-inventory

RuslanLavrov (approved):

AivarIUsupov (approved):

pom.xml
okolawole-ebsco: Great point!

okolawole-ebsco (approved):

PR #516, opened 2022-05-03
[MODINV-616] Update permissions for related instances

wwelling:


damien-git (approved):
These changes fix the issues with mod-orders' integration tests.

PR #518, opened 2022-05-20
MODINV-686 - Add source record preparation before mapping to instance for instance update

RuslanLavrov:
## Purpose
it is necessary to perform preparation of record from payload before mapping to instance to avoid changing instance fields controlled by MARC record fields which are protected by protection settings while processing action profile for instance update.

## Approach
* add "source" instance field check to the ReplaceInstanceEventHandler event handler
 * if the "source" instance field is not equal to MARC, then proceed with event processing as is
 * if the "source" field equals MARC then
   * obtain MARC record related to the target instance by the instance Id from mod-source-record-storage (/source-storage/records/\{instanceId}/formatted?idType=INSTANCE)
   * perform record parsed content preparation using the util method from data-import-processing-core library and use the prepared parsed content for instance mapping
   * ensure a payload of instance update result message (DI_INVENTORY_INSTANCE_UPDATED_READY_FOR_POST_PROCESSING) with prepared record content to be saved in the mod-source-record-storage
* update and add tests

## Learning
[MODINV-686](https://issues.folio.org/browse/MODINV-686)

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
@@ -103,8 +103,8 @@ public class ReplaceInstanceEventHandler extends AbstractInstanceEventHandler {
       String chunkId = dataImportEventPayload.getContext().get(CHUNK_ID_HEADER);
       mappingMetadataCache.get(jobExecutionId, context)
         .compose(parametersOptional -> parametersOptional
-          .map(mappingMetadata -> prepareAndExecuteMapping(dataImportEventPayload, new JsonObject(mappingMetadata.getMappingRules()), new JsonObject(mappingMetadata.getMappingParams())
-            .mapTo(MappingParameters.class), instanceToUpdate))
+          .map(mappingMetadata -> prepareAndExecuteMapping(dataImportEventPayload, new JsonObject(mappingMetadata.getMappingRules()),
+            Json.decodeValue(mappingMetadata.getMappingParams(), MappingParameters.class), instanceToUpdate))
           .orElseGet(() -> Future.failedFuture(format(MAPPING_PARAMETERS_NOT_FOUND_MSG, jobExecutionId,
             recordId, chunkId))))
         .compose(e -> {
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
```diff
@@ -189,15 +194,45 @@ private JsonObject prepareTargetInstance(DataImportEventPayload dataImportEventP
...
+        return Future.succeededFuture();
+      });
+  }
+
+  private Future<Record> getRecordByInstanceId(DataImportEventPayload dataImportEventPayload,
+                                               String instanceId) {
+    SourceStorageRecordsClient client = new SourceStorageRecordsClient(dataImportEventPayload.getOkapiUrl(),
+      dataImportEventPayload.getTenant(), dataImportEventPayload.getToken());
+
+      return client.getSourceStorageRecordsFormattedById(instanceId, "INSTANCE").compose(resp -> {
```
SerhiiNosko, line 229: Can we use constant INSTANCE.value() here?

Aliaksandr-Fedasiuk (approved):

SerhiiNosko (approved):

AivarIUsupov (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
```diff
@@ -252,8 +283,8 @@ private void getActualInstanceAndReInvokeCurrentHandler(Instance instance, Insta
       })
       .exceptionally(e -> {
         eventPayload.getContext().remove(CURRENT_RETRY_NUMBER);
-        LOGGER.error(format("Cannot get actual Instance by id: %s", e.getCause()));
-        promise.fail(format("Cannot get actual Instance by id: %s", e.getCause()));
+        LOGGER.error("Cannot get actual Instance by id: {}", instance.getId(), e);
```
okolawole-ebsco, line 286: Why this change?

okolawole-ebsco (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
RuslanLavrov: the main purpose of the changes is to pass to logger the exception itself instead of its cause, which actually can be a `null`. In addition, logger parametrized message has been applied instead of string formating.

PR #520, opened 2022-05-23
MODINV-550: Bound-with-parts gets an empty collection

K-Felk:
The cause of this problem was that the instance retrieval code that sets the 
bound-with flag does not check to see if holdings records attached to
the instance record actually have items before doing an item record lookup.
This results in it sending an empty list to the code in items that fills out the 
item records, which also has no guard to see if it's actually being passed 
item records.  The item endpoint therefore tries to look up non-existent 
records.  Most of the code just passes over the empty records, but the 
bound-with lookup actually inserts an empty value into the bound-with 
endpoint lookup, resulting in a cql query error.  

I put two guards in place to address this.  One is in the instance code that 
does the bound-with lookup, and checks to make sure there are item records 
to look up before actually doing the lookup.  the second is in the item code, 
and checks any incoming query for item record data to make sure it's actually 
being passed item UUIDs before attempting to look them up.  This second guard is
necessary because I noticed that other parts of the code trigger this problem as well.
Any search for item records that does not match any items will trigger the error,
for the same reason the instance lookup does-the code does not check to see if 
the search actually retrieved anything before passing the results to the lookup code.

### DIFF
--- a/src/main/java/org/folio/inventory/resources/Instances.java
+++ b/src/main/java/org/folio/inventory/resources/Instances.java
@@ -441,7 +441,7 @@ public class Instances extends AbstractInstances {
              .find( holdingsRecordIds, this::cqlMatchAnyByHoldingsRecordIds)
              .thenCompose(
                items -> {
-                 if (items.size() == 0) {
+                 if (items.isEmpty()) {
                   return CompletableFuture.completedFuture(Collections.emptyList());
                  }
                  List<String> itemIds = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


marcjohnson-kint (commented):
Thank you for the thorough analysis and description.

I'm a little surprised this reports 100% coverage given that there don't seem to be new tests for the scenario the new branches are intended to address.

If possible please add a test to demonstrate the issue.

marcjohnson-kint (approved):

K-Felk:
> If possible please add a test to demonstrate the issue.

I can't think of one that makes sense.  There's no observable change in the outcome of any of the operations involved.  A search against the items endpoint that matches no items before would have gotten you an empty result set (as it should).  That is the same result you get now, I just made that decision more explicit and moved it up earlier in the code so that it would not cause error messages to be logged.  Likewise with instance retrieval, the code would mark the bound-with flag as false, which is the correct value if there are no item records attached.  That is the same thing it does now.

PR #521, opened 2022-05-27
MODINV-715 Update instance import overwrites the update date for instance status when profile is not editing the instance status

Abdulkhakimov:
### Purpose
Before merging existing and maped instances (one of step preparation before mapping) we didn't map status updated date. As result it was overwriten in the instance replace operation where we don't set Instance status term in mapping profile.

### Learning
[MODINV-715](https://issues.folio.org/browse/MODINV-715)

### DIFF
--- a/src/main/java/org/folio/inventory/support/InstanceUtil.java
+++ b/src/main/java/org/folio/inventory/support/InstanceUtil.java
@@ -2,6 +2,7 @@ package org.folio.inventory.support;
 
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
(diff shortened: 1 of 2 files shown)


SerhiiNosko (approved):

RuslanLavrov (approved):

okolawole-ebsco:
@Abdulkhakimov is it possible to write a new test or update an existing test for this?

Abdulkhakimov:
> @Abdulkhakimov is it possible to write a new test or update an existing test for this?

Hi @okolawole-ebsco, I've updated existing test.

Aliaksandr-Fedasiuk (approved):

PR #527, opened 2022-07-12
MODINV-674: Cannot add tag when unlinked preceding title

VRohach:
1. Zeroing fields-logic removed.
2. Tests fixed.


src/test/java/api/InstancesApiExamples.java
```diff
@@ -36,39 +38,21 @@
...
-import static org.folio.inventory.domain.instances.Instance.TAG_LIST_KEY;
+import static org.folio.inventory.domain.instances.Instance.*;
 import static org.folio.inventory.domain.instances.PublicationPeriod.publicationPeriodToJson;
-import static org.hamcrest.CoreMatchers.hasItem;
-import static org.hamcrest.CoreMatchers.hasItems;
-import static org.hamcrest.CoreMatchers.is;
-import static org.hamcrest.CoreMatchers.notNullValue;
-import static org.hamcrest.CoreMatchers.nullValue;
-import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.CoreMatchers.*;
```
marcjohnson-kint, line 48: Whist FOLIO has no official standards on imports, in order to reduce the amount of unnecessary flux when different folks order them, some of us tend to use Eclipse import ordering, which does not allow wildcard import statements ([FOLIO-1667|https://issues.folio.org/browse/FOLIO-1667] has more information)

src/test/java/support/matchers/ResponseMatchers.java
```diff
@@ -13,17 +13,69 @@
...
+          .appendText("Response has 422 status and 'message' - ")
+          .appendValue(expectedMessage);
+      }
+
+      @Override
+      protected void describeMismatchSafely(Response response,
+        Description mismatchDescription) {
+        mismatchDescription.appendText("Status: ").appendValue(
+            response.getStatusCode())
+          .appendText(", body: ");
```
marcjohnson-kint, line 55: If I was going to split this statement over multiple lines, I would break it up on the method rather than parameters

(similarly for the other examples)

```suggestion
        mismatchDescription.appendText("Status: ")
          .appendValue( response.getStatusCode())
          .appendText(", body: ");
```

src/test/java/support/matchers/ResponseMatchers.java
```diff
@@ -53,17 +105,21 @@ protected boolean matchesSafely(Response response) {
...
-        mismatchDescription.appendText("Status: ").appendValue(response.getStatusCode())
+      protected void describeMismatchSafely(Response response,
+        Description mismatchDescription) {
+        mismatchDescription.appendText("Status: ").appendValue(
+            response.getStatusCode())
           .appendText(", body: ");
 
-        if (response.getContentType().startsWith(ContentType.APPLICATION_JSON)) {
+        if (response.getContentType().startsWith(
+          ContentType.APPLICATION_JSON)) {
```
marcjohnson-kint, line 122: As this is too long for a single line, I'd be tempted to extract a method like `isJsonContent`

src/test/java/api/InstancesApiExamples.java
```diff
@@ -460,37 +444,45 @@ public void canUpdateAnExistingInstance()
...
+            .put("identifierTypeId", "913300b2-03ed-469a-8179-c1092c991227")
+            .put("value", "0273-656X"))
+        ));
+
     JsonObject updateInstanceRequest = newInstance.copy()
       .put(TAGS_KEY, new JsonObject().put(TAG_LIST_KEY, new JsonArray().add(tagNameTwo)))
       .put(PUBLICATION_PERIOD_KEY, publicationPeriodToJson(new PublicationPeriod(2000, 2012)))
-      .put(PRECEDING_TITLES_KEY, procedingTitles)
+      // Deliberately send update request with preceding title without an ID
+      .put(PRECEDING_TITLES_KEY, precedingTitles)
```
marcjohnson-kint, line 485: The code on line 473 above, seems to deliberately include the `id` property in this representation.

That does not fit with this comment which states that it is deliberately not including the `id`.

Am I missing something in how these line up together?

(the tags UI currently sends the full representation with an amendment to the tags. As I understand it, the situation where the ID is not present is the general edit, which is being changed concurrently)

src/test/java/api/InstancesApiExamples.java
VRohach: @marcjohnson-kint  Thank you for this comment. I miss it in the setting of IDE after moving from another device. Removed wildcards.

src/test/java/support/matchers/ResponseMatchers.java
VRohach: @marcjohnson-kint  Good point! Changed.

src/test/java/support/matchers/ResponseMatchers.java
VRohach: @marcjohnson-kint agree. A brand new method was added.

src/test/java/api/InstancesApiExamples.java
VRohach: @marcjohnson-kint Yes, you are right. I removed this comment from the codebase. Thank you!

VRohach:
@marcjohnson-kint Could you please double-check and approve if Ok?

KaterynaSenchenko (approved):

marcjohnson-kint (approved):

PR #529, opened 2022-07-20
MODINV-674: Cannot add tag when unlinked preceding title - additional fix.

VRohach:
Previous changes **didn't fix** this logic completely. UI devs added transferring "id"-field for the precedingTitles/succeedingTitles.
And UI-request for the "precedingTitles" looks like: 

	"precedingTitles": [
			{
				"id": "c25edfb5-c3bf-4b71-8005-a4e154de6678",
				"title": "Houston oil directory",
				"identifiers": []
			}
	],

**BUT** on back-end side we will have another structure:

	"precedingTitles": [
			{
				"id": "c25edfb5-c3bf-4b71-8005-a4e154de6678",
				"title": "Houston oil directory",
                                 "hrid": null,
                                 "precedingInstanceId": null,
                                 "succedingInstanceId":"c25edfb5-c3bf-4b71-8005-a4e154de6679",
				  "identifiers": []
			}
	],

So, there are additional fields. 

To summarize, there will be an error, when existingInstance will have specific id in "succedingInstanceId", BUT updatedInstance **will NOT have it,** there will be _null_, and comparison between blocked fields will be **failed**.
This field was filled when the instance was created and stored in DB in mod-inventory-storage, that's why this field appeared in existingInstance.

So, to fix it, it seems like we **need to zero** this specific field from comparison. After the comparison is executed, this field will be filled anyway. Because there is a particular method(updatePrecedingSucceedingTitles) to retrieve it from DB and fill in updatedInstance.


KaterynaSenchenko (approved):

RuslanLavrov (approved):

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -269,6 +269,13 @@ private void updateInstance(Instance instance, RoutingContext rContext, WebConte
   private boolean areInstanceBlockedFieldsChanged(Instance existingInstance, Instance updatedInstance) {
     JsonObject existingInstanceJson = JsonObject.mapFrom(existingInstance);
     JsonObject updatedInstanceJson = JsonObject.mapFrom(updatedInstance);
+
+    // We still need zeroing "succeedingInstanceId" in precedingTitles and "precedingInstanceId" in succeedingTitles
+    // because these fields are not provided for/from UI requests. We just ignore these fields on comparing as a blocked fields.
+    // Anyway they are filled after this comparing while fetching from DB in "updatePrecedingSucceedingTitles"-method.
+    zeroingField(existingInstanceJson.getJsonArray(Instance.PRECEDING_TITLES_KEY), PrecedingSucceedingTitle.SUCCEEDING_INSTANCE_ID_KEY);
+    zeroingField(existingInstanceJson.getJsonArray(Instance.SUCCEEDING_TITLES_KEY), PrecedingSucceedingTitle.PRECEDING_INSTANCE_ID_KEY);
```
marcjohnson-kint, line 277: Why are they not provided by the UI?

Is it due to the representation of these structures being different in the business logic API (because those `id` properties are always the `id` of the current instance?

VRohach: Hi @marcjohnson-kint !

I think UI doesn't have it because 'mod-inventory' returns this structure even if we just search Instance in "inventory" app:

```
  "precedingTitles" : [ {
    "id" : "b26d82fa-e648-46c4-be47-ea1b4db32333",
    "title" : "Houston oil directory",
    "identifiers" : [ ]
  } ],
```

marcjohnson-kint (approved):

PR #531, opened 2022-07-22
MODINV-725: Upgrade Users interface to 16.0

K-Felk:


descriptors/ModuleDescriptor-template.json
```diff
@@ -591,7 +591,7 @@
     },
     {
       "id": "users",
-      "version": "15.0"
+      "version": "15.0, 16.0"
```
marcjohnson-kint, line 594: I believe versions are space separated rather than comma separated. It might be that both are supported. It could be worth checking.
```suggestion
      "version": "15.0 16.0"
```

jeremythuff (approved):

rmathew1011 (approved):

PR #533, opened 2022-08-22
MODINV-731: Delete instance/item by CQL

julianladisch:
mod-inventory-storage has changed the DELETE all records API for instances, holdings and items. It requires a query parameter with the CQL query that selects the records to delete.

mod-inventory should also require this query parameter and pass it to mod-inventory-storage APIs.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Now supports interface users 15.0, 16.0 (MODINV-725)
 * Extend instance contributors schema with Authority ID ([MODINV-729](https://issues.folio.org/browse/MODINV-729))
+* Delete instance/item by CQL ([MODINV-731](https://issues.folio.org/browse/MODINV-731))
+* Provides inventory 12.0, supports instance-storage 9.0, holdings-storage 6.0, item-storage 10.0.
 
 ## 18.2.0 2022-06-27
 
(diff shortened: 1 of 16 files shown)


psmagin (approved):

src/main/java/org/folio/inventory/domain/AsynchronousCollection.java
```diff
@@ -11,9 +11,15 @@
 import org.folio.inventory.exceptions.InternalServerErrorException;
 
 public interface AsynchronousCollection<T> {
-  void empty(Consumer<Success<Void>> completionCallback,
+  void empty(String cqlQuery,
```
marcjohnson-kint, line 14: The name empty no longer makes sense given that the CQL query parameter allows for only part of the collection to be affected.

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "inventory",
-      "version": "11.5",
+      "version": "12.0",
```
marcjohnson-kint, line 7: This change should have been done together with a major implementation version change (to 19.0.0 I believe). 

Please raise a follow up pull request to do so.

NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 * Now supports interface users 15.0, 16.0 (MODINV-725)
 * Extend instance contributors schema with Authority ID ([MODINV-729](https://issues.folio.org/browse/MODINV-729))
+* Delete instance/item by CQL ([MODINV-731](https://issues.folio.org/browse/MODINV-731))
+* Provides inventory 12.0, requires instance-storage 9.0, item-storage 10.0, holdings-storage 2.0 ... 6.0.
```
marcjohnson-kint, line 6: Please split these entries onto separate lines in the news, as that is the existing convention and easier to read.

PR #536, opened 2022-09-07
Remove suite for external collection failure tests MODINV-739

marcjohnson-kint:
### Context

Using test suites to try to reduce the overheads of spinning up infrastructure for multiple tests has led to unfortunate coupling between the tests and failures when IDEs etc don't understand the relationship between the tests and suites. It is also unlikely they actually save much time in test execution.

Thus when possible these suites should be removed.

### Approach
* Move the external collection tests to use WireMock (rather than a hand rolled fake)
* Remove the base classes for sharing tests between these tests
* Remove the suite

#### TODO
These tests have a lot of duplication which suggests a flaw in the design of the system or scope of the tests.

Much of the reason for this is that the `external storage module collection` class is an abstract base class relying on subclasses that are all tests separately. Extracting responsibilities from this class and using composition could lead to better testing of mapping and failure handling separately to each record type.


src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleHoldingCollection.java
```diff
@@ -1,22 +1,19 @@
...
   extends ExternalStorageModuleCollection<Holding>
   implements HoldingCollection {
 
-  ExternalStorageModuleHoldingCollection(Vertx vertx,
-                                         String baseAddress,
-                                         String tenant,
-                                         String token,
-                                         HttpClient client) {
+  ExternalStorageModuleHoldingCollection(String baseAddress, String tenant,
+    String token, HttpClient client) {
```
K-Felk, line 16: the line breaks here seem strange.  I actually prefer the old formatting, as it makes the arguments easier to see.

K-Felk (approved):

src/main/java/org/folio/inventory/storage/external/ExternalStorageModuleHoldingCollection.java
marcjohnson-kint: What do you mean by strange?

I used to prefer the old formatting too (except for the horizontal alignment). However many folks don't and it takes up a significant amount of vertical space.

marcjohnson-kint: What would be your ideal formatting for these parameters?

K-Felk: If we can't stack them, could they be on their own line?

K-Felk: I don't think this is an important enough issue to hold up the PR.  I get your point about the fact that we have several different approaches to formatting arguments.

marcjohnson-kint: > I don't think this is an important enough issue to hold up the PR. 

I wasn't holding it up.

> I get your point about the fact that we have several different approaches to formatting arguments.

It's a long standing challenge that hasn't really been addressed (probably in part because my own opinion has varied over time)

PR #537, opened 2022-09-14
Remove Suite for External Collection Tests MODINV-739

marcjohnson-kint:
### Context

Using test suites to try to reduce the overheads of spinning up infrastructure for multiple tests has led to unfortunate coupling between the tests and failures when IDEs etc don't understand the relationship between the tests and suites. It is also unlikely they actually save much time in test execution.

Thus when possible these suites should be removed.

### Approach
* Introduce a base class that achieves the same steps as the suite
* Move all of the existing test classes to the base class
* Remove the suite

#### TODO
The base class introduced in this change retains the undesirable coupling between these tests. It should be possible to remove this later.

These tests have a lot of duplication which suggests a flaw in the design of the system or scope of the tests.

Much of the reason for this is that the `external storage module collection` class is an abstract base class relying on subclasses that are all tests separately. Extracting responsibilities from this class and using composition could lead to better testing of mapping and failure handling separately to each record type.


src/test/java/org/folio/inventory/storage/external/ExternalItemCollectionExamples.java
```diff
@@ -28,16 +24,16 @@
 import org.junit.Before;
 import org.junit.Test;
 
-public class ExternalItemCollectionExamples {
+import lombok.SneakyThrows;
```
rmathew1011, line 27: Could we place this import statement outside the class.

src/test/java/org/folio/inventory/storage/external/ExternalInstanceCollectionExamples.java
```diff
@@ -27,22 +24,22 @@
 import org.junit.Before;
 import org.junit.Test;
 
-public class ExternalInstanceCollectionExamples {
+import lombok.SneakyThrows;
```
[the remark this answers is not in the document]
marcjohnson-kint: Yeah, I'll change this. Thank you for identifying it.

I didn't consciously do this. I didn't even know this was valid.

src/test/java/org/folio/inventory/storage/external/ExternalItemCollectionExamples.java
marcjohnson-kint: Hmm, having investigated this. The import is outside of the class.

I think this is an artefact of how the diff is presented. The diff includes both the addition of the import and a change to the class line. The line above the import in this presentation is the removal of the old class definition.

The code in the branch is:
```
import lombok.SneakyThrows;

public class ExternalItemCollectionExamples extends ExternalStorageTests {
```

K-Felk (approved):

PR #539, opened 2022-09-28
Remove Dublin Core Context from Instances MODINV-734

marcjohnson-kint:
*Scope*

Remove the experimental Dublin Core support that was implemented a long long time ago (in a project far far far away).


K-Felk (approved):
Seems pretty straightforward.

PR #550, opened 2022-11-01
Support "unlimited" bound-with parts in GET item by ID (MODINV-752)

nielserik:
- Partitions the request for referenced holdings and instances, using MultipleRecordsFetchClient.

### DIFF
--- a/src/main/java/org/folio/inventory/resources/Items.java
+++ b/src/main/java/org/folio/inventory/resources/Items.java
@@ -46,6 +46,8 @@ import org.folio.inventory.services.MoveItemIntoStatusService;
 import org.folio.inventory.storage.Storage;
 import org.folio.inventory.storage.external.Clients;
 import org.folio.inventory.storage.external.CollectionResourceClient;
+import org.folio.inventory.storage.external.CqlQuery;
+import org.folio.inventory.storage.external.MultipleRecordsFetchClient;
 import org.folio.inventory.support.CqlHelper;
 import org.folio.inventory.support.ItemUtil;
 import org.folio.inventory.support.JsonArrayHelper;
@@ -676,7 +678,7 @@ public class Items extends AbstractInventoryResource {
 
         allFutures.add(
           setBoundWithTitlesOnItem( item,
-            boundWithPartsClient, holdingsClient, instancesClient));
+            boundWithPartsClient, routingContext));
 
         CompletableFuture<Void> allDoneFuture = allOf(allFutures);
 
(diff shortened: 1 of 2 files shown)


src/test/java/api/BoundWithTests.java
```diff
@@ -222,6 +222,32 @@ public void canRetrieveBoundWithItemByHoldingsRecordId() throws InterruptedExcep
   }
 
   @Test
+  public void canFetchBoundWithItemWithManyParts() throws InterruptedException, TimeoutException, ExecutionException
+  {
+    IndividualResource instance1 = instancesStorageClient.create( InstanceSamples.smallAngryPlanet( UUID.randomUUID() ).put("title", "Instance 1") );
+    IndividualResource holdings1 = holdingsStorageClient.create(new HoldingRequestBuilder()
+      .forInstance(instance1.getId()).permanentlyInMainLibrary().withCallNumber( "HOLDINGS 1" ));
```
marcjohnson-kint, line 229: I'm curious, where does the (sometimes) putting spaces around parameters in brackets come from?

src/test/java/api/BoundWithTests.java
```diff
@@ -222,6 +222,32 @@ public void canRetrieveBoundWithItemByHoldingsRecordId() throws InterruptedExcep
   }
 
   @Test
+  public void canFetchBoundWithItemWithManyParts() throws InterruptedException, TimeoutException, ExecutionException
+  {
```
marcjohnson-kint, line 226: Personally, I would use SneakyThrows to avoid the checked exceptions. And include the opening bracket on the same line as the end of the definition.

src/test/java/api/BoundWithTests.java
```diff
@@ -222,6 +222,32 @@ public void canRetrieveBoundWithItemByHoldingsRecordId() throws InterruptedExcep
...
+    }
+    Response itemResponse = okapiClient.get(ApiTestSuite.apiRoot()+
+        "/inventory/items/"+item.getId())
+      .toCompletableFuture().get(5, SECONDS);
+
+    assertThat("Item has boundWithTitles array with 200 titles",
+      itemResponse.getJson().getJsonArray( "boundWithTitles" ).size(), is(200));
+  }
+
+    @Test
```
marcjohnson-kint, line 250: Please align the indentation of the annotation with the method definition

```suggestion
  @Test
```

src/test/java/api/BoundWithTests.java
```diff
@@ -222,6 +222,32 @@ public void canRetrieveBoundWithItemByHoldingsRecordId() throws InterruptedExcep
...
+
+    for (int i=2; i<=200; i++) {
+      IndividualResource instance = instancesStorageClient.create( InstanceSamples.smallAngryPlanet( UUID.randomUUID() ).put("title", "Instance " + i) );
+      IndividualResource holdings = holdingsStorageClient.create(new HoldingRequestBuilder()
+        .forInstance(instance.getId()).permanentlyInMainLibrary().withCallNumber( "HOLDINGS " + i ));
+      boundWithPartsStorageClient.create(
+        makeObjectBoundWithPart( item.getJson().getString("id"), holdings.getJson().getString( "id" ) ));
+    }
+    Response itemResponse = okapiClient.get(ApiTestSuite.apiRoot()+
+        "/inventory/items/"+item.getId())
```
marcjohnson-kint, line 243:
```suggestion
    Response itemResponse = okapiClient.get(ApiTestSuite.apiRoot() +
        "/inventory/items/" + item.getId())
```

src/main/java/org/folio/inventory/resources/Items.java
```diff
@@ -871,35 +873,46 @@ private boolean circulationNoteChanged(CirculationNote newNote, CirculationNote
...
-      } else {
-        return fetchHoldingsByForeignKeys( boundWithParts, holdingsClient ).thenCompose(
-          holdingsRecords -> fetchInstancesByForeignKeys( holdingsRecords, instancesClient ).thenCompose(
-            instances -> {
-              JsonArray boundWithTitles = buildBoundWithTitlesArray( holdingsRecords, instances );
-              item.withBoundWithTitles( boundWithTitles ).withIsBoundWith( true );
-              return CompletableFuture.completedFuture( null );
-            } ) );
-      }});
+    return getBoundWithPartsForItemFuture( item, boundWithPartsClient )
```
marcjohnson-kint, line 886: I think it could be good to start extracting some of this behaviour out of the items class and into separate classes, as the class has gotten rather large (over 1000 lines)

nielserik: All right.

src/test/java/api/BoundWithTests.java
nielserik: Probably from a previous set of IDE code formatting rules.

src/test/java/api/BoundWithTests.java
nielserik: All right, not familiar with SneakyThrows.

marcjohnson-kint: It's an [annotation](https://projectlombok.org/features/SneakyThrows) provided by lombok that (kinda) turns checked exceptions into runtime exceptions.

I don't advocate for it in production code. In tests, where any uncaught exception fails the test, it can get rid of a bunch of noise in the tests.

src/main/java/org/folio/inventory/resources/Items.java
marcjohnson-kint: How was this resolved?

PR #554, opened 2022-11-11
[MODINV-627]Improving code coverage by log messages for the INFO level

KyryloShulzhenko:
In the task [MODINVSTOR-866](https://issues.folio.org/browse/MODINVSTOR-866) we met situation with specific DI behavior on the one Villanova test env. We couldn't reproduce this case on the other hosted and local envs. Current INFO log messages have not allowed us to understand the main reason of this situation. We need to improve code coverage by log messages for the INFO level for all modules of DI.

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/consumers/DataImportKafkaHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/consumers/DataImportKafkaHandler.java
@@ -122,8 +122,10 @@ public class DataImportKafkaHandler implements AsyncRecordHandler<String, String
           .orElse(CompletableFuture.failedFuture(new EventProcessingException(format("Job profile snapshot with id '%s' does not exist", jobProfileSnapshotId)))))
         .whenComplete((processedPayload, throwable) -> {
           if (throwable != null) {
+            LOGGER.error(throwable.getMessage());
             promise.fail(throwable);
           } else if (DI_ERROR.value().equals(processedPayload.getEventType())) {
+            LOGGER.warn("Failed to process data import event payload: {}", processedPayload.getEventType());
             promise.fail("Failed to process data import event payload");
           } else {
             promise.complete(record.key());
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractAuthorityEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractAuthorityEventHandler.java
@@ -4,6 +4,7 @@ import static io.vertx.core.json.JsonObject.mapFrom;
 import static java.lang.String.format;
 
 import static org.folio.inventory.dataimport.handlers.matching.util.EventHandlingUtil.constructContext;
+import static org.folio.inventory.dataimport.util.LoggerUtil.logParametersEventHandler;
 import static org.folio.rest.jaxrs.model.ProfileSnapshotWrapper.ContentType.ACTION_PROFILE;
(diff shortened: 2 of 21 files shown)


src/main/java/org/folio/inventory/dataimport/consumers/DataImportKafkaHandler.java
```diff
@@ -122,8 +122,10 @@ public Future<String> handle(KafkaConsumerRecord<String, String> record) {
           .orElse(CompletableFuture.failedFuture(new EventProcessingException(format("Job profile snapshot with id '%s' does not exist", jobProfileSnapshotId)))))
         .whenComplete((processedPayload, throwable) -> {
           if (throwable != null) {
+            LOGGER.error(throwable.getMessage());
             promise.fail(throwable);
           } else if (DI_ERROR.value().equals(processedPayload.getEventType())) {
+            LOGGER.error("Failed to process data import event payload: {}", processedPayload.getEventType());
```
RomanChernetskyi, line 128: we considered using 'WARN' level for exceptions that we handle

KyryloShulzhenko: done

RomanChernetskyi (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractAuthorityEventHandler.java
```diff
@@ -64,6 +64,7 @@ protected AbstractAuthorityEventHandler(Storage storage, MappingMetadataCache ma
 
   @Override
   public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload payload) {
+    LOGGER.trace("handle:: parameter payload: {}", payload);
```
Aliaksandr-Fedasiuk, line 67: Does it makes sense to add LOGGER.debug("handle:: parameter payload is: {}", payload == null ? "null" : ''not null");

src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateMarcHoldingsEventHandler.java
```diff
@@ -90,16 +90,20 @@ public UpdateMarcHoldingsEventHandler(Storage storage,
 
   @Override
   public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload payload) {
+    LOGGER.trace("handle:: parameter payload: {}", payload);
     CompletableFuture<DataImportEventPayload> future = new CompletableFuture<>();
     try {
       if (!isExpectedPayload(payload)) {
+        LOGGER.warn("Payload {} is not expected", payload);
```
Aliaksandr-Fedasiuk, line 97: Print full payload in TRACE. In WARN - print eventType or jobExecutionId

Aliaksandr-Fedasiuk:
A lot of places where we have only TRACE level for DataImportEventPayload info. I think we also can add DEBUG level with eventType and jobExecutionId for cases where DataImportEventPayload is not null.

src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractAuthorityEventHandler.java
RuslanLavrov: Hello guys, we don't need here check for payload ==null, because if paylod is null, this method/handler will not be called

Aliaksandr-Fedasiuk (approved):

VRohach (approved):
Everything is OK for me. Good work!

PR #562, opened 2022-12-23
MODINV-765/742: Implement mechanism for checking the last action and the if the Order action profiles exists

VRohach:
## Purpose
Create a mechanism inside mod-inv that will search if the current action is the last one in the action-profiles tree and if there is an action profile for Create Order inside the current tree.

## Approach
New OrderHelperService was created. It has a method that checks if there is an Order CREATE action profile in snapshotProfileWrapper-tree, and if the current action profile is the last one. If true, then send all needed data to the current payload for Order post-processing. If there is some kind of error, then send "DI_ERROR"-event.

This method will trigger from ALL Handlers for Instance/Holdings/Item on action Created after all logic was executed successfully.
## Learning
https://issues.folio.org/browse/MODINV-765 AND https://issues.folio.org/browse/MODINV-742


VRohach:
This PR completed after: https://github.com/folio-org/data-import-raml-storage/pull/268 will be merged.

RomanChernetskyi (approved):

KaterynaSenchenko (approved):

RuslanLavrov (approved):

Aliaksandr-Fedasiuk (approved):

PR #589, opened 2023-04-28
MODINV-800  Unify API response in case of ValidationException (http 422)

pawelmol:
Changed behavior of creating instance to properly handle ValidationException

### DIFF
--- a/src/main/java/org/folio/inventory/resources/Instances.java
+++ b/src/main/java/org/folio/inventory/resources/Instances.java
@@ -165,7 +165,11 @@ public class Instances extends AbstractInstances {
           log.warn(
             format("Failed to create self link for instance: %s", e.toString()));
         }
-      }).exceptionally(doExceptionally(routingContext));
+        }).exceptionally(error -> {
+          handleFailure(getKnownException(error), routingContext);
+
+          return null;
+        });
   }
 
   private void update(RoutingContext rContext) {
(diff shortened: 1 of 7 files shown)


Shans-Kaluhin (approved):

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -165,7 +165,11 @@ private void create(RoutingContext routingContext) {
           log.warn(
             format("Failed to create self link for instance: %s", e.toString()));
         }
-      }).exceptionally(doExceptionally(routingContext));
+        }).exceptionally(error -> {
+          handleFailure(getKnownException(error), routingContext);
+
+          return null;
+        });
```
psmagin, line 172: Does this need only in this API endpoint? Do other endpoints are unified?

Could you please add some tests for this?

pawelmol: Well, this was the only endpoint for instances which needed to be unified .

For PUT/POST items errors are redirected from mod-inventory-storage (FailureResponseConsumer.serverError(routingContext.response())),

For PUT/POST holdings requests are send directly from frontend to holdings-storage/holdings, so there is no need to unify

For PUT/POST instances errors are now translated from mod-inventory-storage in the same way (handleFailure(getKnownException(ex), rContext);)

pawelmol: Items and Holdings endpoints are forwarding validation errors correctly so I made changes for Instances only.

Added test for create and update an Instance canFrowardInstanceUpdateValidationErrorFromStorage, canFrowardInstanceCreateValidationErrorFromStorage

Changed ma previous solution due to above test failuares.

psmagin (approved):

PR #598, opened 2023-06-30
MODINV-813 : Fix deleting the value from the "Call number" field from"MARC Holdings".

ArtHov:
Fix deleting the value from the "Call number" field from"MARC Holdings" by a separate update of directly impacted fields.


src/test/java/org/folio/inventory/eventhandlers/UpdateHoldingsQuickMarcEventHandlerTest.java
```diff
@@ -45,7 +45,9 @@ public class UpdateHoldingsQuickMarcEventHandlerTest {
 
   private static final String MAPPING_RULES_PATH = "src/test/resources/handlers/holdings-rules.json";
   private static final String INSTANCE_PATH = "src/test/resources/handlers/holdings.json";
+  private static final String INSTANCE_WITH_CELL_NUMBER_PATH = "src/test/resources/handlers/holdings-with-call-number.json";
   private static final String RECORD_PATH = "src/test/resources/handlers/holdings-record.json";
+  private static final String RECORD_WITHOUT_CELL_NUMBER_PATH = "src/test/resources/handlers/holdings-record-without-call-number.json";
```
psmagin, line 50: Please fix the naming of constants

src/test/java/org/folio/inventory/eventhandlers/UpdateHoldingsQuickMarcEventHandlerTest.java
```diff
@@ -132,6 +143,46 @@ public void shouldProcessEvent() {
     Assert.assertEquals("http://localhost", argument.getValue().getOkapiLocation());
   }
 
+  @Test
+  public void shouldProcessEvent1() throws IOException {
```
psmagin, line 147: please provide better naming for the test

src/test/java/org/folio/inventory/eventhandlers/UpdateHoldingsQuickMarcEventHandlerTest.java
```diff
@@ -132,6 +143,46 @@ public void shouldProcessEvent() {
...
+    record = new JsonObject(TestUtil.readFileFromPath(
+      RECORD_WITHOUT_CELL_NUMBER_PATH));
+    List<HoldingsType> holdings = new ArrayList<>();
+    holdings.add(new HoldingsType().withName("testingnote$a").withId("fe19bae4-da28-472b-be90-d442e2428eadx"));
```
psmagin, line 151: Please create a constant for id as it used several times

src/test/resources/handlers/holdings-record-without-call-number.json
psmagin, line 1: I think it could be clean-uped to not have so much fields, and left only required and needed for test

psmagin (changes requested):

psmagin (approved):

PR #601, opened 2023-07-10
B20.0 MODINV 832

nielserik:


nielserik:
This is a P1 bug fix that was also merged to master a few minutes ago.

PR #605, opened 2023-07-25
MODINV-817: Shadow Instances - add support for Consortium source values

Aliaksandr-Fedasiuk:
## Purpose
Shadow Instances are instances that live in a local tenant and share corresponding identifier with another instance in the central tenant of a consortium. Library staff will be able to creating holdings and items under a Shadow Instance but are not able to modify the Shadow Instance.

## Approach
Blocking creating Instances by API with other source values except MARC, FOLIO, CONSORTIUM-MARC, and CONSORTIUM-FOLIO.

## Learning
[MODINV-817](https://issues.folio.org/browse/MODINV-817)

### DIFF
--- a/src/main/java/org/folio/inventory/validation/InstanceSourceTypeValidator.java
+++ b/src/main/java/org/folio/inventory/validation/InstanceSourceTypeValidator.java
@@ -18,13 +18,14 @@ public final class InstanceSourceTypeValidator {
   private InstanceSourceTypeValidator() {}
 
   public static CompletableFuture<Instance> refuseWhenIncorrectSource(Instance instance) {
-
-    final ValidationError incorrectSourceError = new ValidationError(
-      "Instance source is incorrect", "source", instance.getSource());
-
     return isSourceTypeCorrect(instance.getSource())
       ? completedFuture(instance)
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/inventory/validation/InstanceSourceTypeValidator.java
```diff
@@ -0,0 +1,34 @@
...
+
+public final class InstanceSourceTypeValidator {
+
+  private static final String[] INSTANCE_SOURCE_TYPES = {"MARC", "FOLIO", "CONSORTIUM-MARC", "CONSORTIUM-FOLIO"};
+
+  private InstanceSourceTypeValidator() {}
+
+  public static CompletableFuture<Instance> refuseWhenIncorrectSource(Instance instance) {
+
+    final ValidationError incorrectSourceError = new ValidationError(
```
yaroslav-epam, line 22: it seems we could inline the creation of `ValidationError` to line#27, thus it won't be created in case of the correct source type

yaroslav-epam (approved):

RuslanLavrov (approved):

PR #609, opened 2023-08-15
[MODINV-854] Adjust multiple Holdings update to send DI_ERROR only when no Holdings Updated

RomanChernetskyi:
### Approach
Adjust multiple Holdings update to send DI_ERROR only when no Holdings Updated, otherwise complete with partial error
### Jira
https://issues.folio.org/browse/MODINV-854

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateHoldingEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateHoldingEventHandler.java
@@ -121,7 +121,6 @@ public class UpdateHoldingEventHandler implements EventHandler {
           MappingParameters mappingParameters = Json.decodeValue(mappingMetadataDto.getMappingParams(), MappingParameters.class);
           MappingManager.map(dataImportEventPayload, new MappingContext().withMappingParameters(mappingParameters));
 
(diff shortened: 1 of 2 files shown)


RuslanLavrov (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/UpdateHoldingEventHandler.java
```diff
@@ -140,33 +139,32 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload d
...
-              future.completeExceptionally(ar.cause());
-            }
-          });
+          CompositeFuture.all(updatedHoldingsRecordFutures)
+            .onSuccess(ar -> {
+              if (!expiredHoldings.isEmpty()) {
+                processOLError(dataImportEventPayload, future, holdingsRecordCollection, expiredHoldings.get(0), errors);
+              }
+              String errorsAsStringJson = Json.encode(errors);
+              if (!updatedHoldingsRecord.isEmpty() || errors.size() == 0) {
```
RuslanLavrov, line 158:
```suggestion
              if (!updatedHoldingsRecord.isEmpty() || errors.isEmpty() == 0) {
```

PR #613, opened 2023-08-29
MODINV-862: Deserialize multiple holdings as an array

yaroslav-epam:
[MODINV-862](https://issues.folio.org/browse/MODINV-862) propblem 2.

**Depends on PR:** https://github.com/folio-org/data-import-processing-core/pull/305.

**Description:**  deserialize multiple holdings as an array.

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/ItemLoader.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/ItemLoader.java
@@ -10,6 +10,7 @@ import java.util.stream.Collectors;
 
 import io.vertx.core.Vertx;
 import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
 import org.folio.DataImportEventPayload;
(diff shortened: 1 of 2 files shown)


VRohach (approved):

RomanChernetskyi (approved):

RuslanLavrov (approved):

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/ItemLoader.java
```diff
@@ -62,7 +63,8 @@ protected String addCqlSubMatchCondition(DataImportEventPayload eventPayload) {
         JsonObject itemAsJson = new JsonObject(eventPayload.getContext().get(EntityType.ITEM.value()));
         cqlSubMatch = format(" AND id == \"%s\"", itemAsJson.getString("id"));
       } else if (isNotEmpty(eventPayload.getContext().get(EntityType.HOLDINGS.value()))) {
-        JsonObject holdingAsJson = new JsonObject(eventPayload.getContext().get(EntityType.HOLDINGS.value()));
+        JsonObject holdingAsJson = new JsonArray(eventPayload.getContext().get(EntityType.HOLDINGS.value()))
+          .getJsonObject(0);
```
okolawole-ebsco, line 67: @yaroslav-epam  In current karate test failure, this line errors sometime with a index out of range exception. That is preventing DI_INVENTORY_ITEM_NOT_MATCHED to be published by mod-inventory to allow a discarded message to show in journal logs.

we should change this to account for empty array scenarios.

okolawole-ebsco: The exception is not being logged as well so we make modifications to allow proper exception propagation.

PR #615, opened 2023-09-04
[MODINV-816] Ad Hoc Shadow Instance Creation During Data Import

RomanChernetskyi:
### Purpose
When a "Create Holding" action is required in a job profile and the Shadow Instance where the holding would live is not present, DI will initiate the consortium sharing process to create the Shadow Instance before creating the holding. Creation of the Shadow Instance could apply to other scenarios where a Shadow Instance needs to be acted upon but it is not present in the local tenant.
### Approach
Made changes in Create Holdings handler:
Determine if Instance that is present in the event payload context is present in local tenant (as local or shadow), send GET request to inventory-storage by Instance UUID, if it is not present - send GET request to /user-tenant?limit=1 and retrieve "centralTenantId", send GET request with "centralTenantId" at tenant header at "/consortia" endpoint to get **consortiumId**, send POST to /consortia/**consortiumId**/sharing/instances  to pull shared Instance on local tenant as a Shadow copy . Remaining process of Holdings creation stays as is.


src/main/java/org/folio/inventory/consortium/services/ConsortiumService.java
```diff
@@ -0,0 +1,35 @@
...
+import org.folio.inventory.consortium.entities.SharingInstance;
+
+public interface ConsortiumService {
+  /**
+   * Retrieves centralTenantId, consortiumId and runs job at mod-consortia that creates shadow instance for locale tenant
+   * @param context                    - Context for creating shadow instance.
+   * @param instanceId                 - id of shadow instance to create.
+   * @return - future of sharingInstance
+   */
+  Future<SharingInstance> createShadowInstance(Context context, String instanceId);
```
tatsiana-tarhonskaya, line 14: Looks like consortia dependency should be added to module descriptor

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+                return CompletableFuture.failedFuture(new NotFoundException(message));
+              }
+              String centralTenantId = userTenants.getJsonObject(0).getString("centralTenantId");
+              LOGGER.debug("getCentralTenantId:: Found centralTenantId: {}", centralTenantId);
+              return CompletableFuture.completedFuture(centralTenantId);
+            } else {
+              String message = String.format("Error retrieving centralTenantId by tenant id: %s, status code: %s, response message: %s",
+                context.getTenantId(), httpResponse.getStatusCode(), httpResponse.getBody());
+              LOGGER.warn(String.format("getCentralTenantId:: %s", message));
+              return CompletableFuture.failedFuture(new EventProcessingException(message));
```
tatsiana-tarhonskaya, line 71: Not sure here, but EventProcessingException looks like some exception during event processing. So if this service is going to be reused somewhere else, like within rest endpoint, it might look misleading

RomanChernetskyi: Thanks, done

src/main/java/org/folio/inventory/consortium/services/ConsortiumService.java
```diff
@@ -0,0 +1,35 @@
+package org.folio.inventory.consortium.services;
+
+import io.vertx.core.Future;
+import org.folio.inventory.common.Context;
+import org.folio.inventory.consortium.entities.SharingInstance;
+
+public interface ConsortiumService {
+  /**
+   * Retrieves centralTenantId, consortiumId and runs job at mod-consortia that creates shadow instance for locale tenant
```
tatsiana-tarhonskaya, line 9: Consortia permissions should be added to module descriptor

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+import org.folio.inventory.support.http.client.OkapiHttpClient;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+
+import static org.folio.inventory.dataimport.handlers.matching.util.EventHandlingUtil.constructContext;
+
+public class ConsortiumServiceImpl implements ConsortiumService{
```
tatsiana-tarhonskaya, line 25: formatting

RomanChernetskyi: Fixed, thanks

VRohach (approved):

src/main/java/org/folio/inventory/consortium/services/ConsortiumService.java
tatsiana-tarhonskaya: Discussed that user should possess these permissions

src/main/java/org/folio/inventory/consortium/entities/SharingInstance.java
```diff
@@ -0,0 +1,60 @@
+package org.folio.inventory.consortium.entities;
+
+import java.util.UUID;
+
+public class SharingInstance {
```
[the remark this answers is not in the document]
RomanChernetskyi: Thanks, done

KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/consortium/exceptions/ConsortiumException.java
```diff
@@ -0,0 +1,10 @@
+package org.folio.inventory.consortium.exceptions;
+
+/**
+ * Exception that used for consortium process
+ */
+public class ConsortiumException extends RuntimeException{
```
tatsiana-tarhonskaya, line 6: Formatting

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+          }));
+    return Future.fromCompletionStage(completableFuture);
+  }
+
+  public Future<SharingInstance> shareInstance(Context context, String consortiumId, SharingInstance sharingInstance) {
+    CompletableFuture<SharingInstance> completableFuture = createOkapiHttpClient(context)
+      .thenCompose(client ->
+        client.post(context.getOkapiLocation() + String.format(SHARE_INSTANCE_ENDPOINT, consortiumId), Json.encode(sharingInstance))
+          .thenCompose(httpResponse -> {
+            if (httpResponse.getStatusCode() == HttpStatus.SC_OK && !httpResponse.getJson().getString("status").equals(Status.ERROR.toString())) {
```
tatsiana-tarhonskaya, line 107: is "status" always present in response? Might be better to compare with Status.ERROR

src/main/java/org/folio/inventory/consortium/entities/Status.java
```diff
@@ -0,0 +1,7 @@
+package org.folio.inventory.consortium.entities;
+
+public enum Status {
```
okolawole-ebsco, line 3: If this is referring to status of instance sharing, i think this class should be called SharingStatus. The name is very generic right now.

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
okolawole-ebsco: If the sharing status is "IN_PROGRESS", this code will assume the sharing is successful. Is my statement correct?

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
okolawole-ebsco, line 1: In general, I would like retrieval of consortium id and central tenant id to not happen for every message. This values is not likely to change ever but would incur HTTP call costs each time. Consider having a cache that expires its items in 5 mins for any particular context. 
We can leave this class as is, and optimize after some performance tests have been executed.

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateHoldingEventHandler.java
```diff
@@ -118,13 +123,15 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload d
...
                 holdingAsJson.put("sourceId", FOLIO_SOURCE_ID);
-                fillInstanceIdIfNeeded(dataImportEventPayload, holdingAsJson);
+                fillInstanceIdIfNeeded(instanceId, holdingAsJson);
               }
 
               LOGGER.trace(format("handle:: Mapped holdings: %s", holdingsList.encode()));
               dataImportEventPayload.getContext().put(HOLDINGS.value(), holdingsList.encode());
               return List.of(Json.decodeValue(payloadContext.get(HOLDINGS.value()), HoldingsRecord[].class));
             })
+            .compose(holdingsToCreate -> ConsortiumUtil.createShadowInstanceIfNeeded(consortiumService, storage.getInstanceCollection(context),
```
okolawole-ebsco, line 133: We should only attempt to create a shadow user if ECS is enabled for the tenant. If not, non-ECS flow will incur more than neccessary costs. Consider storing the central tenant Id, consortium id for a tenant if ECS is enabled. If ECS is not enabled, the store could also reflect that for the requested tenant.

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateHoldingEventHandler.java
```diff
@@ -168,27 +175,31 @@ private void prepareEvent(DataImportEventPayload dataImportEventPayload) {
     dataImportEventPayload.setCurrentNode(dataImportEventPayload.getCurrentNode().getChildSnapshotWrappers().get(0));
   }
 
-  private void fillInstanceIdIfNeeded(DataImportEventPayload dataImportEventPayload, JsonObject holdingAsJson) {
+  private void fillInstanceIdIfNeeded(String instanceId, JsonObject holdingAsJson) {
```
okolawole-ebsco, line 178: The previous version of this call allowed instance id to be filled without making an extra HTTP call later. I think we should leave room for both.
- if non-ECS, get from payload
- if ECS, get from payload then verify existence

src/main/java/org/folio/inventory/dataimport/handlers/actions/InstanceUpdateDelegate.java
```diff
@@ -53,7 +53,7 @@ public Future<Instance> handle(Map<String, String> eventPayload, Record marcReco
       var mappedInstance = recordMapper.mapRecord(parsedRecord, mappingParameters, mappingRules);
       InstanceCollection instanceCollection = storage.getInstanceCollection(context);
 
-      return getInstanceById(instanceId, instanceCollection)
+      return ConsortiumUtil.findInstanceById(instanceId, instanceCollection)
```
okolawole-ebsco, line 56: Seems weird that ConsortiumUtil is used to get an instance even when the tenant is non-ECS. I think `getInstanceById` should be here. Inside `getInstanceById` we would determine if to use the consortiumUtil or the instanceCollection.

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
RomanChernetskyi: @okolawole-ebsco Your statement is correct, in this case, when we are pulling instance from the central tenant sharing status should be "COMPLETE" or "ERROR". https://github.com/folio-org/mod-consortia/blob/master/src/main/java/org/folio/consortia/service/impl/SharingInstanceServiceImpl.java#L101

src/main/java/org/folio/inventory/consortium/exceptions/ConsortiumException.java
RomanChernetskyi: Thanks, done

src/main/java/org/folio/inventory/consortium/entities/Status.java
RomanChernetskyi: Thanks, done

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
```diff
@@ -0,0 +1,131 @@
...
+  private static final String SHARE_INSTANCE_ENDPOINT = "/consortia/%s/sharing/instances";
+  private static final String CONSORTIA_ENDPOINT = "/consortia";
+  private static final String SHARING_INSTANCE_ERROR = "Error during sharing Instance for sourceTenantId: %s, targetTenantId: %s, instanceIdentifier: %s, status code: %s, response message: %s";
+  private final HttpClient httpClient;
+
+  public ConsortiumServiceImpl(HttpClient httpClient) {
+    this.httpClient = httpClient;
+  }
+
+  public Future<SharingInstance> createShadowInstance(Context context, String instanceId) {
```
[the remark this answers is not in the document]
RomanChernetskyi: Thanks, done

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
RomanChernetskyi: Hi @okolawole-ebsco I created ticket to add cache for consortium id and central tenant id https://issues.folio.org/browse/MODINV-872

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+    SharingInstance sharingInstance = new SharingInstance();
+    sharingInstance.setSourceTenantId(consortiumConfiguration.getCentralTenantId());
+    sharingInstance.setInstanceIdentifier(UUID.fromString(instanceId));
+    sharingInstance.setTargetTenantId(context.getTenantId());
+    return shareInstance(centralTenantContext, consortiumConfiguration.getConsortiumId(), sharingInstance);
+  }
+
+  @Override
+  public Future<Optional<ConsortiumConfiguration>> getConsortiumConfiguration(Context context) {
+    CompletableFuture<Optional<ConsortiumConfiguration>> completableFuture = createOkapiHttpClient(context)
```
okolawole-ebsco, line 49: Is there any way the okapi client can use vertx future instead of CompletableFuture?

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+              String message = String.format("Error retrieving centralTenantId by tenant id: %s, status code: %s, response message: %s",
+                context.getTenantId(), httpResponse.getStatusCode(), httpResponse.getBody());
+              LOGGER.warn(String.format("getCentralTenantId:: %s", message));
+              return CompletableFuture.failedFuture(new ConsortiumException(message));
+            }
+          }));
+    return Future.fromCompletionStage(completableFuture);
+  }
+
+  // Returns completed future if the sharing status is "IN_PROGRESS" or "COMPLETE"
```
okolawole-ebsco, line 73: returns "successful" future, not just completed.

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateHoldingEventHandler.java
```diff
@@ -118,13 +123,22 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload d
...
               LOGGER.trace(format("handle:: Mapped holdings: %s", holdingsList.encode()));
               dataImportEventPayload.getContext().put(HOLDINGS.value(), holdingsList.encode());
               return List.of(Json.decodeValue(payloadContext.get(HOLDINGS.value()), HoldingsRecord[].class));
             })
+            .compose(holdingsToCreate -> consortiumService.getConsortiumConfiguration(context)
+              .compose(consortiumConfigurationOptional -> {
+                if (consortiumConfigurationOptional.isPresent()) {
+                  return ConsortiumUtil.createShadowInstanceIfNeeded(consortiumService, storage.getInstanceCollection(context),
+                    context, getInstanceId(dataImportEventPayload), consortiumConfigurationOptional.get()).map(holdingsToCreate)
+                    .map(holdingsToCreate);
```
okolawole-ebsco, line 138: looks like `map(holdingsToCreate)` is called twice. Is this neccessary?

okolawole-ebsco (approved):

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
RomanChernetskyi: We cannot user vert.x future, as far as OkapiHttpClient uses CompletionStage on post/get requests

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateHoldingEventHandler.java
RomanChernetskyi: Removed, thanks

src/main/java/org/folio/inventory/consortium/services/ConsortiumServiceImpl.java
RomanChernetskyi: Done, Thanks

PR #620, opened 2023-09-13
MODINV-863:  Add necessary interface into required interfaces in mod-inventory md

Maksat-Galymzhan:
## Purpose
Add necessary interface into required interfaces in mod-inventory md

## Approach
[MODINV-863](https://issues.folio.org/browse/MODINV-863)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -592,6 +592,10 @@
     {
       "id": "bound-with-parts-storage",
       "version": "1.0"
+    },
+    {
+      "id": "authority-storage",
+      "version": "2.0"
     }
   ],
   "optional": [


descriptors/ModuleDescriptor-template.json
```diff
@@ -592,6 +592,10 @@
     {
       "id": "bound-with-parts-storage",
       "version": "1.0"
+    },
+    {
+      "id": "authority-storage",
+      "version": "1.1"
```
KaterynaSenchenko, line 598: Hi @Maksat-Galymzhan, please find which module provides this interface and what is the latest version. I have suspicion that it was upgraded in the current dev cycle (Poppy release)

Maksat-Galymzhan: you are right,  there is a new version which is [2.0](https://github.com/folio-org/mod-entities-links/blob/master/pom.xml)

KaterynaSenchenko (approved):

PR #621, opened 2023-09-15
[MODSOURCE-646] Set central tenant to context at post processing during consortium flow

RomanChernetskyi:
### Purpose
Set central tenant to context at post processing during consortium flow
### Approach
Set central tenant to context for post processing if event payload contains flag


src/main/java/org/folio/inventory/dataimport/handlers/actions/MarcBibMatchedPostProcessingEventHandler.java
```diff
@@ -39,6 +43,7 @@ public class MarcBibMatchedPostProcessingEventHandler implements EventHandler {
   private static final String ERROR_HOLDING_MSG = "Error loading inventory holdings for MARC BIB";
   private static final String MATCHED_MARC_BIB_KEY = "MATCHED_MARC_BIBLIOGRAPHIC";
   private final Storage storage;
+  public static final String CENTRAL_TENANT_ID = "CENTRAL_TENANT_ID";
```
yaroslav-epam, line 46: Looks like the constant might be private

yaroslav-epam (approved):

okolawole-ebsco (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/MarcBibMatchedPostProcessingEventHandler.java
RomanChernetskyi: Thanks, done

src/main/java/org/folio/inventory/dataimport/handlers/actions/MarcBibModifiedPostProcessingEventHandler.java
```diff
@@ -32,6 +32,7 @@
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.folio.ActionProfile.FolioRecord.INSTANCE;
 import static org.folio.DataImportEventTypes.DI_SRS_MARC_BIB_RECORD_MODIFIED_READY_FOR_POST_PROCESSING;
+import static org.folio.inventory.dataimport.handlers.actions.MarcBibMatchedPostProcessingEventHandler.getTenant;
```
RuslanLavrov, line 35: propose to move MarcBibMatchedPostProcessingEventHandler.getTenant() method to the EventHandlingUtil class since we use it in both handlers MarcBibMatchedPostProcessingEventHandler and MarcBibModifiedPostProcessingEventHandler

RuslanLavrov (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/MarcBibModifiedPostProcessingEventHandler.java
RomanChernetskyi: Thanks, done

PR #622, opened 2023-09-22
[MODINV-868] Perform MARC To Instance Matching in Local Tenant & Central Tenant

RomanChernetskyi:


src/main/java/org/folio/inventory/dataimport/handlers/matching/AbstractMatchEventHandler.java
```diff
@@ -63,12 +70,58 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload d
...
+        if (consortiumConfiguration.isPresent() && !consortiumConfiguration.get().getCentralTenantId().equals(context.getTenantId())) {
+          LOGGER.debug("matchCentralTenantIfNeeded:: Start matching on central tenant with id: {}", consortiumConfiguration.get().getCentralTenantId());
+          String localMatchedInstance = dataImportEventPayload.getContext().get(getEntityType().value());
+          preparePayloadBeforeConsortiumProcessing(dataImportEventPayload, consortiumConfiguration.get(), mappingMetadataDto, matchingParametersRelations);
+          return MatchingManager.match(dataImportEventPayload)
+            .thenCompose(isMatchedConsortium -> {
+              dataImportEventPayload.setTenant(context.getTenantId());
+              if (isMatchedConsortium && isMatchedLocal) {
+                LOGGER.warn("matchCentralTenantIfNeeded:: Found multiple results during matching on local tenant: {} and central tenant: {} ",
+                  consortiumConfiguration.get().getCentralTenantId(), context.getTenantId());
```
okolawole-ebsco, line 101: I think there is a case where a shared instance is matched in the central tenant and the corresponding shadow instance in matched in the local tenant. So it is possible to have multiple results and still be fine. What do you think?

src/main/java/org/folio/inventory/dataimport/handlers/matching/AbstractMatchEventHandler.java
```diff
@@ -63,12 +70,58 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload d
...
+                return CompletableFuture.failedFuture(new MatchingException(String.format(FOUND_MULTIPLE_ENTITIES, context.getTenantId(), consortiumConfiguration.get().getCentralTenantId())));
+              }
+              if (StringUtils.isEmpty(dataImportEventPayload.getContext().get(getEntityType().value()))) {
+                LOGGER.debug("matchCentralTenantIfNeeded:: Matched on local tenant: {}", context.getTenantId());
```
okolawole-ebsco, line 105: should this log line be shown when we match on local tenant?
Thinking we would see log lines that shows there is a match on central+local but no log for match on local only.

src/main/java/org/folio/inventory/dataimport/handlers/matching/AbstractMatchEventHandler.java
RomanChernetskyi: Hi, Olamide, thanks, this is a good point, I added a check if the Instance matched locally is a shadow instance, if yes we do not start the consortium matching process

okolawole-ebsco: As discussed in our meeting today, more changes are going to be made?

RomanChernetskyi: Added this changes, please review

src/main/java/org/folio/inventory/dataimport/handlers/matching/AbstractMatchEventHandler.java
```diff
@@ -86,4 +147,5 @@ public boolean isEligible(DataImportEventPayload dataImportEventPayload) {
 
   protected abstract String getNotMatchedEventType();
 
+  protected abstract boolean isConsortiumAvailable();
```
KaterynaSenchenko, line 150:
```suggestion
  protected abstract boolean isConsortiumActionAvailable();
```
let's make it a be more clear

KaterynaSenchenko (approved):

RuslanLavrov (approved):

PR #627, opened 2023-09-28
MODINV-866: Add deduplication mechanism for sharing Instance.

VRohach:
## Purpose
This needs to avoid event duplication for mod-inventory module in the scope of Instance-sharing.

## Approach
There was added a table named "events_shared_instances" with only one column - "event_id" as a primary key.
When ConsortiumInstanceSharingHandler started, the eventId from Kafka-record will be saved inside new DB-table.
If there is the same eventId - handle it via specific PGException Error Code and ignore the other logic inside ConsortiumInstanceSharingHandler.
So, deduplication is based on eventId from Kafka-event.
Moreover, this mechanism is easy-to-extend in the future. (If we will need some deduplications based ONLY ON eventId, NOT recordId+entityId). Just add new tableName to the enum and create a table for it.
New Services/DAOs/Entities were added.
New Tests added. Existed tests are fixed.
NEWS updated.
This mechanism was tested on the rancher env via Kafka-ui tool. (duplicated event were sent)

## Learning
JIRA: https://issues.folio.org/browse/MODINV-866


src/main/resources/liquibase/tenant/scripts/v-20.1.0/2023-09-28--14-00-create-events-shared-instances-table.xml
```diff
@@ -0,0 +1,15 @@
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+    http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.0.xsd">
+
+  <changeSet id="2023-09-28--14-00-create-events-shared-instances-table" author="VolodymyrRohach">
+    <createTable tableName="events_shared_instances">
+      <column name="event_id" type="uuid">
```
okolawole-ebsco, line 9: please add date fields to the table

VRohach: @okolawole-ebsco  Good point. Added.

KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/common/dao/EventIdStorageDaoImpl.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.logging.log4j.Logger;
+import org.folio.inventory.domain.relationship.EventTable;
+import org.folio.inventory.domain.relationship.EventToEntity;
+
+import java.util.UUID;
+
+public class EventIdStorageDaoImpl implements EventIdStorageDao {
+  private static final Logger LOGGER = LogManager.getLogger(EventIdStorageDaoImpl.class);
+
+  private static final String INSERT_FUNCTION = "INSERT INTO {schemaName}.{tableName} VALUES ($1::uuid, current_timestamp) RETURNING *;";
```
okolawole-ebsco, line 17: Due to the nature of this table, it can get intense(high cpu) when allowing the database to set the "current date". Under high load it is better to generate the date from the app tier and just send the data to the database.

VRohach: @okolawole-ebsco  As far as I remember, in most places, we left this datetime-generation to the DB side. 
This is good point for the future (maybe to move this generation to the Java-code, not functions/trigger on the DB-side...?)
So yes, I improved it here - moved this generation to the codebase.
Could you please double-check it?

Maksat-Galymzhan (approved):

RuslanLavrov (approved):

okolawole-ebsco (approved):

src/main/java/org/folio/inventory/domain/relationship/EventTable.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Enum which stores pointer on DB-tableNames inside.
+ */
+@Getter
+public enum EventTable {
+  SHARED_INSTANCE("events_shared_instances");
+
+  private final String tableName;
+
+  EventTable(String tableName) {
```
yaroslav-epam, line 14: since Lombok is used `@RequiredArgsConstructor` could be applied

yaroslav-epam (approved):

PR #633, opened 2023-10-05
[MODINV-879] Unlink local MARC authorities from shared MARC bib SRS records

RomanChernetskyi:
### Purpose
In a consortium environment, fields in local MARC bib records may be linked to local MARC authority records. However, Instances with source = MARC may also be promoted to be shared to the rest of the consortium. When the Instance with source = MARC is shared, we need to unlink the controlled fields in the MARC bibliographic record from the local MARC authority records.

### Approach
Get Link between instance and authorities from mod-entities-link, unlink local authorities, remove $9 with local authorities UUID from marcbib fields configured at linking rules


src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
```diff
@@ -26,29 +42,32 @@
...
+  private final EntitiesLinksService entitiesLinksService;
+  private final Storage storage;
   private final Vertx vertx;
 
-  public MarcInstanceSharingHandlerImpl(InstanceOperationsHelper instanceOperations, Vertx vertx) {
+  public MarcInstanceSharingHandlerImpl(InstanceOperationsHelper instanceOperations, Storage storage, Vertx vertx) {
     this.vertx = vertx;
     this.instanceOperations = instanceOperations;
     this.restDataImportHelper = new RestDataImportHelper(vertx);
+    this.entitiesLinksService = new EntitiesLinksServiceImpl(vertx, vertx.createHttpClient());
```
okolawole-ebsco, line 56: I am not too knowledgeable about the vertx http client. Should there be only one client for the app? should a client recreated somewhere else be used?

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
```diff
@@ -74,6 +93,63 @@ public Future<String> publishInstance(Instance instance, SharingInstance sharing
       });
   }
 
+  private Future<Record> unlinkAuthorityLinksIfNeeded(Record marcRecord, String instanceId, Context context, Storage storage) {
```
okolawole-ebsco, line 96: should we change the name to `detachLocalAuthorityLinksIfNeeded` since there isn't a javadoc to explain further?

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
```diff
@@ -74,6 +93,63 @@ public Future<String> publishInstance(Instance instance, SharingInstance sharing
...
+                                                String instanceId,Context context, AuthorityRecordCollection authorityRecordCollection) {
+
+    return getLocalAuthoritiesIdsList(entityLinks, authorityRecordCollection)
+      .compose(localAuthoritiesIds -> {
+        if (localAuthoritiesIds.isEmpty()) {
+          return Future.succeededFuture(marcRecord);
+        }
+        List<String> fields = linkingRules.stream().map(LinkingRuleDto::getBibField).toList();
+        LOGGER.debug("unlinkLocalAuthorities:: Unlinking from tenant: {} local authorities: {}", context.getTenantId(), localAuthoritiesIds);
+        if (MarcRecordUtil.removeSubfieldsThatContainsValues(marcRecord, fields, '9', localAuthoritiesIds)) {
```
okolawole-ebsco, line 119: Please add comments to this if block. I am not sure why shared authority links are in play.

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
```diff
@@ -74,6 +93,63 @@ public Future<String> publishInstance(Instance instance, SharingInstance sharing
...
+          List<Link> sharedAuthorityLinks = getSharedAuthorityLinks(entityLinks, localAuthoritiesIds);
+          return entitiesLinksService.putInstanceAuthorityLinks(context, instanceId, sharedAuthorityLinks).map(marcRecord);
+        }
+        LOGGER.warn(format("unlinkLocalAuthorities:: Error during remove of 9 subfields from record: %s", marcRecord.getId()));
```
okolawole-ebsco, line 123: This should log an error message from the removal of 9 subfields. "Why did it fail"

src/main/java/org/folio/inventory/consortium/services/EntitiesLinksService.java
```diff
@@ -0,0 +1,14 @@
+package org.folio.inventory.consortium.services;
+
+import io.vertx.core.Future;
+import org.folio.Link;
+import org.folio.LinkingRuleDto;
+import org.folio.inventory.common.Context;
+
+import java.util.List;
+
+public interface EntitiesLinksService {
```
okolawole-ebsco, line 10: nit: I don't think this interface is consortium specific. Could be put in a package outside of "consortium"

src/main/java/org/folio/inventory/consortium/util/ConsortiumUtil.java
```diff
@@ -31,4 +39,14 @@ public static Future<Optional<SharingInstance>> createShadowInstanceIfNeeded(Con
         return Future.failedFuture(throwable);
       });
   }
+
+  public static CompletableFuture<OkapiHttpClient> createOkapiHttpClient(Context context, HttpClient httpClient) {
```
okolawole-ebsco, line 43: could this use vertx future instead of java.util.concurrent.CompletableFuture?

src/main/java/org/folio/inventory/consortium/util/MarcRecordUtil.java
okolawole-ebsco, line 1: Was this class copied from somewhere else?

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
RomanChernetskyi: Moved initialization of HTTP client to consumer, so we have one client, thanks

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
RomanChernetskyi: Renamed, thanks

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
RomanChernetskyi: Added a comment

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
RomanChernetskyi: We are logging an error at **MarcRecordUtil.removeSubfieldsThatContainsValues** is it okay, or should we throw an exception from the util class and log an error message at **unlinkLocalAuthorities** method?

src/main/java/org/folio/inventory/consortium/services/EntitiesLinksService.java
RomanChernetskyi: Moved from consortium package, thanks

src/main/java/org/folio/inventory/consortium/util/ConsortiumUtil.java
RomanChernetskyi: OkapiHttpClient uses CompletableFuture under the hood, so using vert.x Future could cause some difficulties composing Future and CompletableFuture

src/main/java/org/folio/inventory/consortium/util/MarcRecordUtil.java
RomanChernetskyi: Yes, some logic was copied from [AdditionalFieldsUtil](https://github.com/folio-org/mod-source-record-storage/blob/master/mod-source-record-storage-server/src/main/java/org/folio/services/util/AdditionalFieldsUtil.java)

KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
okolawole-ebsco: I am asking for the exception from the util class to be logged as part of the message.

src/main/java/org/folio/inventory/consortium/util/ConsortiumUtil.java
okolawole-ebsco: I see. The difficulties are just moved elsewhere since mod-inventory ought to be a vert.x app.

okolawole-ebsco (approved):

PR #636, opened 2023-10-16
MODINV-886 - Problem with Inventory Single Record Imports for UPDATES in member tenant, for shared instances

RuslanLavrov:
## Purpose
Lack of permission causes an error on record deletion on local tenant. As a result local marc record that should be deleted/marked as deleted is considered during records matching that in turn causes multiple matching error (and error during single record imports for record updates).
The module gets an error response on record deletion but the module logs say that marc record has been deleted.

## Approach
* add response status verification from request for record deletion in order to complete result and to write log message accordingly
* add and update test

## Learning
[MODINV-886](https://issues.folio.org/browse/MODINV-886)

Requires: https://github.com/folio-org/mod-consortia/pull/129

### DIFF
--- a/src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
+++ b/src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
@@ -197,31 +197,22 @@ public class MarcInstanceSharingHandlerImpl implements InstanceSharingHandler {
   Future<String> deleteSourceRecordByRecordId(String recordId, String instanceId, String tenantId, SourceStorageRecordsClient client) {
     LOGGER.info("deleteSourceRecordByRecordId :: Delete source record with recordId={} for instance by InstanceId={} from tenant {}",
       recordId, instanceId, tenantId);
-    Promise<String> promise = Promise.promise();
-    client.deleteSourceStorageRecordsById(recordId).onComplete(responseResult -> {
-      try {
-        if (responseResult.failed()) {
-          LOGGER.error("deleteSourceRecordByRecordId:: Error deleting source record with recordId={} by InstanceId={} from tenant {}",
-            recordId, instanceId, tenantId, responseResult.cause());
-          promise.fail(responseResult.cause());
-        } else if(responseResult.result().statusCode() != SC_NO_CONTENT) {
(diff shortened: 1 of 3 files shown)


KaterynaSenchenko (approved):

tatsiana-tarhonskaya (approved):

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
```diff
@@ -75,7 +76,7 @@ public Future<String>  publishInstance(Instance instance, SharingInstance sharin
           .compose(result -> {
             if ("COMMITTED".equals(result)) {
               // Delete source record by instance ID if the result is "COMMITTED"
```
tatsiana-tarhonskaya, line 78: Looks like this comment should to be updated to 'by record id'. Or just removed

src/main/java/org/folio/inventory/consortium/handlers/MarcInstanceSharingHandlerImpl.java
```diff
@@ -193,25 +194,30 @@ Future<Record> getSourceMARCByInstanceId(String instanceId, String sourceTenant,
...
     Promise<String> promise = Promise.promise();
     client.deleteSourceStorageRecordsById(recordId).onComplete(responseResult -> {
       try {
         if (responseResult.failed()) {
-          LOGGER.error("deleteSourceRecordByInstanceId:: Error deleting source record with recordId={} by InstanceId={} from tenant {}",
+          LOGGER.error("deleteSourceRecordByRecordId:: Error deleting source record with recordId={} by InstanceId={} from tenant {}",
             recordId, instanceId, tenantId, responseResult.cause());
           promise.fail(responseResult.cause());
+        } else if(responseResult.result().statusCode() != SC_NO_CONTENT) {
+          String msg = format("Error deleting source record with recordId=%s by InstanceId=%s from tenant %s, responseStatus=%s, body=%s",
```
okolawole-ebsco, line 208: How does this change, solve multiple matches in SRS? is it supposed to solve multiple matches?

RuslanLavrov: The PR changes are aimed at correcting the log message in accordance with the response to the request for record deletion. The lack of permission itself, which eventually causes multiple matching, has been fixed in https://github.com/folio-org/mod-consortia/pull/129

PR #642, opened 2023-10-31
[MODINV-907] Avoid NPE during create of item after match by holding identifier

RomanChernetskyi:
### Purpose
When you want to create an item connected to an existing holdings, you match to that holdings hrid and have an action to create item. If you try to create it Item will be errored with NullException in JSON log

### Approach
Not try to remove holding identifiers from context if them does not exist there


Maksat-Galymzhan (approved):

yaroslav-epam (approved):

KaterynaSenchenko (approved):

tatsiana-tarhonskaya (approved):

JavokhirAbdullayev (approved):
looks good! but, i would suggest to use static imports

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

PR #658, opened 2023-12-15
MODINV-906: fix Item creation when Statistical Code value is empty

yaroslav-epam:
**Approach**
Check whether Statistical Code value is empty, if yes then call inventory-storage with an empty list of statistical code ids in the payload, otherwise follow the previous logic.

**Tested locally**
image

### DIFF
--- a/src/main/java/org/folio/inventory/support/ItemUtil.java
+++ b/src/main/java/org/folio/inventory/support/ItemUtil.java
@@ -1,5 +1,6 @@
 package org.folio.inventory.support;
 
+import static java.util.Collections.emptyList;
 import static org.folio.inventory.domain.converters.EntityConverters.converterForClass;
 import static org.folio.inventory.support.JsonArrayHelper.toList;
 import static org.folio.inventory.support.JsonArrayHelper.toListOfStrings;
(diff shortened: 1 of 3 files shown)


VRohach (approved):

src/main/java/org/folio/inventory/support/ItemUtil.java
```diff
@@ -192,8 +193,7 @@ public static Item jsonToItem(JsonObject itemRequest) {
     List<String> formerIds = toListOfStrings(
       itemRequest.getJsonArray(Item.FORMER_IDS_KEY));
 
-    List<String> statisticalCodeIds = toListOfStrings(
-      itemRequest.getJsonArray(Item.STATISTICAL_CODE_IDS_KEY));
+    List<String> statisticalCodeIds = getStatisticalCodeIds(itemRequest.getJsonArray(Item.STATISTICAL_CODE_IDS_KEY));
```
KaterynaSenchenko, line 196: would that work only for Items? we need to make sure that the same logic is applied for stat codes of Instance and Holdings

yaroslav-epam: this PR can be closed. Created PR https://github.com/folio-org/data-import-processing-core/pull/323 with fix  in di-core where the mapping logic is shared for Items, Instances and Holdings

PR #661, opened 2024-01-03
MODINV-918 Consortial: Local HRID gets added as 035 when instance is shared

JavokhirAbdullayev:
### Purpose
Consortial: Local HRID gets added as 035 when instance is shared [MODINV-918](https://issues.folio.org/browse/MODINV-918) 

### Approach 
remove 001 field from marc record before sending it to the data-import


JavokhirAbdullayev:
jenkins suspended, will reopen

KaterynaSenchenko (approved):

RuslanLavrov (approved):

PR #663, opened 2024-01-08
MODINV-883 Set instance record as deleted

JavokhirAbdullayev:
### Purpose
Set instance record as deleted [MODINV-883](https://issues.folio.org/browse/MODINV-883)

### Approach
add endpoint for soft deletion 
add tests

### TODO
make some refactors after soft delete by instance id will be added in mod-srs in another ticket


Aliaksandr-Fedasiuk (approved):
Please update NEWS.md

yaroslav-epam (approved):

Maksat-Galymzhan (approved):

PR #666, opened 2024-01-10
[MODINV-901] Allow to link local instance to shared instance

RomanChernetskyi:
### Purpose 
Allow to link local instance to shared instance

### Approach
Create shadow instance at member tenant if user trying to link local instance to shared instance


KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/Launcher.java
```diff
@@ -1,8 +1,11 @@
 package org.folio.inventory;
 
+import io.vertx.core.Context;
```
yaroslav-epam, line 3: unused imports

yaroslav-epam (approved):

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/inventory/exceptions/BadRequestException.java
```diff
@@ -0,0 +1,10 @@
+package org.folio.inventory.exceptions;
+
+import org.apache.http.HttpStatus;
+import org.folio.inventory.support.http.ContentType;
+
+public class BadRequestException extends AbstractInventoryException {
```
VRohach, line 6: Add documentation here pls.

src/main/java/org/folio/inventory/resources/AbstractInstances.java
```diff
@@ -118,10 +126,41 @@ private CompletableFuture<List<Response>> updateInstanceRelationships(Instance i
...
+        if (throwable instanceof NotFoundException) {
+          log.warn("confirmInstanceExistAtLocalMember:: Trying to link shared and local instance, instanceId: {}", instanceId);
+          return Future.failedFuture(new BadRequestException(LINKING_SHARED_AND_LOCAL_INSTANCE_ERROR));
+        }
+        return Future.failedFuture(throwable);
+      }).toCompletionStage().toCompletableFuture();
+  }
+
+  private static String getRelatedInstanceId(Instance instance, InstanceRelationship relationship) {
+    return !relationship.superInstanceId.equals(instance.getId()) ? relationship.superInstanceId : relationship.subInstanceId;
```
VRohach, line 163: Can superInstanceId can be null here? If yes, there should be added a check here to avoid NPE.

JavokhirAbdullayev (approved):
please update news md too

src/main/java/org/folio/inventory/exceptions/BadRequestException.java
RomanChernetskyi: Done, thanks

src/main/java/org/folio/inventory/resources/AbstractInstances.java
RomanChernetskyi: It cannot be null, thanks

PR #673, opened 2024-01-23
[MODINVSTOR-1154] - Add new field "Display summary" for the item schema

azizbekxm:
### Purpose
https://issues.folio.org/browse/MODINVSTOR-1154
### Approach
added new field to item schema

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -31,6 +31,10 @@
       "type": "boolean",
       "description": "Records the fact that the record should not be displayed in a discovery system"
     },
+    "displaySummary": {
+      "description": "Display summary about the item.",
+      "type": "string"
+    },
     "title": {
       "type": "string",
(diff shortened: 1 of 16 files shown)


ramls/item.json
```diff
@@ -31,6 +31,10 @@
       "type": "boolean",
       "description": "Records the fact that the record should not be displayed in a discovery system"
     },
+    "displaySummary": {
```
imerabishvili, line 34: Do we need to update samples?

azizbekxm: I think we can't introduce change to samples

SerhiiNosko (approved):

SerhiiNosko:
@azizbekxm we should increase minor version of API with adding new field

KaterynaSenchenko (approved):

RuslanLavrov (approved):

imerabishvili (approved):

PR #682, opened 2024-02-02
MODINV-951 Fix error when importing without 005 field

dmytrokrutii:


dmytrokrutii:
A forced push was performed to change the author's email in the last 3 commits

PR #685, opened 2024-02-07
[MODINV-961] Call delete endpoint in SRS when Instance is marked for deletion

RomanChernetskyi:
### Purpose
Call delete endpoint in SRS when Instance is marked for deletion
### Approach 
Update implementation of /inventory/instances/{id}/mark-deleted endpoint to send DELETE request to /source-storage/records/{id} with idType=INSTANCE instead of suppressFromDiscovery call.

### DIFF
--- a/src/main/java/org/folio/inventory/resources/Instances.java
+++ b/src/main/java/org/folio/inventory/resources/Instances.java
@@ -66,6 +66,7 @@ public class Instances extends AbstractInstances {
   private static final String BLOCKED_FIELDS_UPDATE_ERROR_MESSAGE = "Instance is controlled by MARC record, these fields are blocked and can not be updated: ";
   private static final String ID = "id";
   private static final String INSTANCE_ID = "instanceId";
+  public static final String INSTANCE_ID_TYPE = "INSTANCE";
 
   public Instances(final Storage storage, final HttpClient client, final ConsortiumService consortiumService) {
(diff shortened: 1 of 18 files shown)


KaterynaSenchenko (approved):

Maksat-Galymzhan (approved):

JavokhirAbdullayev (approved):

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -211,6 +211,25 @@ private void updateSuppressFromDiscoveryFlag(WebContext wContext, Instance updat
     }
   }
 
+  private void deleteSourceStorageRecord(WebContext wContext, String instanceId) {
+    try {
+      SourceStorageRecordsClient srsClient = getSourceStorageRecordsClient(wContext);
+      srsClient.deleteSourceStorageRecordsById(instanceId, INSTANCE_ID_TYPE, httpClientResponse -> {
+        if (httpClientResponse.result().statusCode() == HttpStatus.HTTP_NO_CONTENT.toInt()) {
+          log.info(format("Source storage record was successfully deleted in SRS. InstanceID: %s", instanceId));
```
okolawole-ebsco, line 219: For the future, this could look like below
`log.info("Source storage record was successfully deleted in SRS. InstanceID: {}", instanceId);`

PR #692, opened 2024-02-20
MODINV-944 OOM issue in mod-inventory

JavokhirAbdullayev:
### Purpose
OOM issue in mod-inventory


src/main/java/org/folio/inventory/dataimport/consumers/MarcBibUpdateKafkaHandler.java
```diff
@@ -122,6 +122,9 @@ private void sendEventToKafka(LinkUpdateReport linkUpdateReport, List<KafkaHeade
     try {
       var kafkaRecord = createKafkaProducerRecord(linkUpdateReport, kafkaHeaders);
       producer.send(kafkaRecord)
+        .<Void>mapEmpty()
+        .eventually(v -> producer.flush())
+        .eventually(v -> producer.close())
```
okolawole-ebsco, line 127: Why close this producer since the it is supposed to be owned by this class? Seems to me the producer will be closed some time after the first message, subsequent messages will fail because the producer will already be closed. Do you agree?

okolawole-ebsco: This approach is good. We will need to create a new shared producer each time. The current code looks to only create the shared producer once.

KaterynaSenchenko (approved):

okolawole-ebsco (approved):

PR #694, opened 2024-02-26
MODINV-969: The duplicate record is created after updating action

Maksat-Galymzhan:
## Purpose
 Two records are displayed in the table and SRS MARC column shows No action.
 ## Approach
 Added generation to incoming record and fixed status in srm module.
 
 ## Learning
 [MODINV-969](https://folio-org.atlassian.net/browse/MODINV-969)

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
@@ -300,6 +300,7 @@ public class ReplaceInstanceEventHandler extends AbstractInstanceEventHandler { 
 
           if (instance.getSource().equals(MARC.getValue())) {
             incomingRecord.setMatchedId(existingRecord.getMatchedId());
+            incomingRecord.setGeneration(null);
             String updatedIncomingRecord = Json.encode(incomingRecord);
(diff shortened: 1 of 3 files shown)


JavokhirAbdullayev (approved):

KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
```diff
@@ -299,6 +300,10 @@ private Future<Void> prepareRecordForMapping(DataImportEventPayload dataImportEv
 
           if (instance.getSource().equals(MARC.getValue())) {
             incomingRecord.setMatchedId(existingRecord.getMatchedId());
+            if (nonNull(existingRecord.getGeneration())) {
+              int incrementedGeneration = existingRecord.getGeneration();
+              incomingRecord.setGeneration(++incrementedGeneration);
```
RuslanLavrov, line 305: propose to set record's generation to null so that the generation will be calculated in mod-source-record storage by the logic of the endpoint used in this handler for record update

RuslanLavrov: discussed with @Maksat-Galymzhan that during processing action to update instance for the case when a job profile contains action profiles for updating marc-bib and for updating instance the SRS logic for generation calculation doesn't account previous record's generation which was calculated during processing action for marc-bib update,

RuslanLavrov (approved):

PR #697, opened 2024-02-28
MODINV-839 support holdings record pointing to many multi-title items (bound-withs) 

nielserik:
**Purpose**
Support use case where there is a large number (100+) bound-withs that all contain the same title. 

**Issue**
The API '/inventory/items-by-holdings-id' is dedicated to a UI Inventory use case, where a holdings record accordion
lists all items associated with the given holdings record, whether those items are regular, single-title items or bound-with/multi-title items.

The regular single-title items have simply been looked up in '/item-storage/items' by their holdings record ID (foreign key). IDs for the multi-title items have been looked up in the bound_with_part table by holdings record ID, and the items have then been looked up in /item-storage/items by the resulting list of item IDs. 

However, it turns out that there are use cases where some titles are a part of so many multi-title items that the look up by a list of item IDs results in a query string that violates the restrictions on the length of query strings. 

**Approach**
This PR is about refactoring the code to use partitioned queries instead of ID lists for looking up the multi-title items -- in order to support this use case.  Using partitioned queries additionally means the sorting and paging cannot be done in the database but must be performed by the module.


src/main/java/org/folio/inventory/resources/ItemsByHoldingsRecordId.java
```diff
@@ -21,179 +26,214 @@
 import java.lang.invoke.MethodHandles;
 import java.net.MalformedURLException;
 import java.net.URL;
-import java.util.Arrays;
-import java.util.List;
+import java.util.*;
```
KaterynaSenchenko, line 29: please do not use wildcard imports, thanks

KaterynaSenchenko (approved):

PR #702, opened 2024-03-13
MODINV-936 - Adjust Match event handlers to take into account multiple search results from previous match

RuslanLavrov:
## Purpose
to take into account multiple search results from previous marc-bib matching during subsequent instances or holdings matching processing

## Approach
* proceed import process in case multiple search results from previous marc-bib matching occur if the next profile is a match 
profile responsible for inventory entities matching. Change logic of loading instances and holdings to use multiple search results from marc-bib matching. Add a new match event handler that would be an entry point for processing instance/holdings/item/marc-bib matching and would be responsible for calling appropriate dedicated match event handler depending on entity type that is looked for. That in turn would allow to avoid sending an event with marc-bib multiple match results and process it by subsequent match profile for instance/holding matching.
* added and updated tests

## Learning
[MODINV-936](https://issues.folio.org/browse/MODINV-936)


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/inventory/dataimport/handlers/matching/AbstractMarcMatchEventHandler.java
```diff
@@ -240,6 +279,16 @@ private Future<Optional<Record>> getRecordById(RecordIdentifiersDto recordIdenti
...
   }
 
+  private void populatePayloadWithInstancesIdentifiers(RecordsIdentifiersCollection recordsIdentifiersCollection,
+                                                       DataImportEventPayload payload) {
+    List<String> instancesIds = recordsIdentifiersCollection.getIdentifiers()
+      .stream()
+      .map(RecordIdentifiersDto::getExternalId)
+      .toList();
+
+    payload.getContext().put(INSTANCES_IDS_KEY, Json.encode(instancesIds));
```
VRohach, line 289: @RuslanLavrov  Can you please specify - where this "INSTANCES_IDS_KEY" will be removed?

RuslanLavrov: @VRohach, as we were discussing earlier, this data will be deleted during construction search queries in scope of instance/holdings matching process. Added comment, but not sure whether suggested comment is relevant since this data represents a result of this handler execution, and how the result of particular handler execution will be processed is not responsibility of this handler itself. Just curious why suggested comment has not been added in the similar case here)?
https://github.com/folio-org/mod-inventory/pull/619/files#diff-302cbdbf38e8c258327303ee746d277e58ea9884049b352a7dcee82c27f52d06R129
https://github.com/folio-org/mod-source-record-storage/pull/569/files#diff-ef1ff9bf04347de98e733094a1cc759e4944ec85ac414944a92485371da60331R340

PR #708, opened 2024-04-04
MODINV-949 Update Data Import logic to normalize OCLC 035 values

JavokhirAbdullayev:
### Purpose
[MODINV-949](https://folio-org.atlassian.net/browse/MODINV-949)  Update Data Import logic to normalize OCLC 035 values

### Approach
update AdditionalFieldsUtil class

image

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
@@ -111,6 +111,7 @@ public class CreateInstanceEventHandler extends AbstractInstanceEventHandler {
                 MappingParameters mappingParameters = Json.decodeValue(mappingMetadata.getMappingParams(), MappingParameters.class);
                 AdditionalFieldsUtil.updateLatestTransactionDate(targetRecord, mappingParameters);
                 AdditionalFieldsUtil.move001To035(targetRecord);
+                AdditionalFieldsUtil.normalize035(targetRecord);
                 payloadContext.put(EntityType.MARC_BIBLIOGRAPHIC.value(), Json.encode(targetRecord));
                 return prepareAndExecuteMapping(dataImportEventPayload, new JsonObject(mappingMetadata.getMappingRules()), mappingParameters);
               })
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
@@ -41,8 +41,7 @@ import static org.folio.ActionProfile.FolioRecord.INSTANCE;
 import static org.folio.ActionProfile.FolioRecord.MARC_BIBLIOGRAPHIC;
 import static org.folio.DataImportEventTypes.DI_INVENTORY_INSTANCE_CREATED;
 import static org.folio.DataImportEventTypes.DI_INVENTORY_INSTANCE_CREATED_READY_FOR_POST_PROCESSING;
-import static org.folio.inventory.dataimport.util.AdditionalFieldsUtil.SUBFIELD_I;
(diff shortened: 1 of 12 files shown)


src/main/java/org/folio/inventory/dataimport/util/AdditionalFieldsUtil.java
```diff
@@ -261,6 +272,99 @@ public static void move001To035(Record srcRecord) {
...
+      Set<String> normalized035Subfields = formatOclc(subfields);
+
+      updateOclcSubfield(srsRecord, normalized035Subfields);
+    }
+  }
+
+  public static Set<String> formatOclc(List<Subfield> subFields) {
+    Set<String> processedSet = new LinkedHashSet<>();
+
+    Pattern pattern = Pattern.compile("\\(OCoLC\\)(tfe|ocm|ocn)?0*(\\d+)");
```
VRohach, line 287: @JavokhirAbdullayev  Let's move it into constant.

src/test/java/org/folio/inventory/dataimport/util/AdditionalFieldsUtilTest.java
```diff
@@ -336,6 +344,69 @@ public void shouldAdd035If001NotEqual003() {
     Assert.assertEquals(expectedParsedContent, parsedRecord.getContent());
   }
 
+  @RunWith(Parameterized.class)
+  public static class OclcFieldNormalizationTest {
```
VRohach, line 348: @JavokhirAbdullayev  Is there any way to add more tests for other corner-cases for these new methods in the AdditionalFieldsUtilTest?

JavokhirAbdullayev: there should be only one public method, the others should be private, just pushed changes, thanks, nested test class has 4 test cases to check various scenarios

JavokhirAbdullayev: added additinal tests

dmytrokrutii (approved):

VRohach (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/CreateInstanceEventHandler.java
```diff
@@ -111,6 +111,7 @@ public CompletableFuture<DataImportEventPayload> handle(DataImportEventPayload d
                 MappingParameters mappingParameters = Json.decodeValue(mappingMetadata.getMappingParams(), MappingParameters.class);
                 AdditionalFieldsUtil.updateLatestTransactionDate(targetRecord, mappingParameters);
                 AdditionalFieldsUtil.move001To035(targetRecord);
+                AdditionalFieldsUtil.normalize035(targetRecord);
```
KaterynaSenchenko, line 114: this function should also be called when Instance is updated

JavokhirAbdullayev: good catch, thanks

KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/dataimport/handlers/actions/InstanceUpdateDelegate.java
```diff
@@ -43,7 +44,7 @@ public Future<Instance> handle(Map<String, String> eventPayload, Record marcReco
     try {
       JsonObject mappingRules = new JsonObject(eventPayload.get(MAPPING_RULES_KEY));
       MappingParameters mappingParameters = new JsonObject(eventPayload.get(MAPPING_PARAMS_KEY)).mapTo(MappingParameters.class);
-
+      AdditionalFieldsUtil.normalize035(marcRecord);
```
RuslanLavrov, line 47: looks like, we need to add this method call to ReplaceInstanceEventHandler for being called when instance is updated during data import

PR #712, opened 2024-04-13
MODINV-949 Update Data Import logic to normalize OCLC 035 values

JavokhirAbdullayev:


Maksat-Galymzhan (approved):

src/main/java/org/folio/inventory/dataimport/util/AdditionalFieldsUtil.java
```diff
@@ -273,7 +274,7 @@ public static void move001To035(Record srcRecord) {
   }
 
   public static void normalize035(Record srsRecord) {
-    List<Subfield> subfields = get035SubfieldValues(srsRecord, TAG_035, TAG_035_SUB);
+    List<Subfield> subfields = get035OclcSubfieldValues(srsRecord, TAG_035, TAG_035_SUB);
```
Aliaksandr-Fedasiuk, line 277: In my opinion, the name get035SubfieldOclcValues would be clearer because 035O can be misread without IDE.

src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
```diff
@@ -215,6 +215,7 @@ private void processInstanceUpdate(DataImportEventPayload dataImportEventPayload
             if (instanceToUpdate.getSource().equals(MARC.getValue())) {
               setExternalIds(targetRecord, instance);
               AdditionalFieldsUtil.remove035FieldWhenRecordContainsHrId(targetRecord);
+              AdditionalFieldsUtil.normalize035(targetRecord);
```
RuslanLavrov, line 218: please, place this method call before instance mapping so that the normalized value will be set to instance

RuslanLavrov (approved):

PR #716, opened 2024-04-22
MODINV-986: InstanceIngress events consumption

PBobylev:


src/main/java/org/folio/inventory/handler/InstanceIngestEventHandler.java
```diff
@@ -0,0 +1,238 @@
...
+import org.folio.kafka.AsyncRecordHandler;
+import org.folio.kafka.exception.DuplicateEventException;
+import org.folio.processing.exceptions.EventProcessingException;
+import org.folio.processing.mapping.MappingManager;
+import org.folio.processing.mapping.defaultmapper.processor.parameters.MappingParameters;
+import org.folio.processing.mapping.mapper.MappingContext;
+import org.folio.rest.jaxrs.model.EntityType;
+import org.folio.rest.jaxrs.model.Record;
+
+// temporary copy of CreateInstanceEventHandler
```
PBobylev, line 52: Ignore this class for now, it will be implemented as a next step

src/main/java/org/folio/inventory/handler/InstanceIngestEventHandler.java
```diff
@@ -0,0 +1,21 @@
...
+public class InstanceIngestEventHandler implements AsyncRecordHandler<String, String> {
+
+  private static final Logger LOGGER = LogManager.getLogger(InstanceIngestEventHandler.class);
+
+  @Override
+  public Future<String> handle(KafkaConsumerRecord<String, String> kafkaConsumerRecord) {
+    // to extract and re-use common logic from CreateInstanceEventHandler
+    // 1. Change event; 2. Re-use all except: source type to be changed to BIBFRAME, DI event not to be sent
+    LOGGER.info("to be replaced with actual code in Step 2 of MODINV-986");
+    return null;
```
aleksei-pronichev, line 18: suggest not to return null, even temporarily.
Just return simple empty string future

src/main/java/org/folio/inventory/support/KafkaConsumerVerticle.java
```diff
@@ -0,0 +1,144 @@
...
+
+  protected String getCacheEnvVariable(JsonObject config, String variableName) {
+    var cacheExpirationTime = config.getString(variableName);
+    if (isBlank(cacheExpirationTime)) {
+      cacheExpirationTime = CACHE_EXPIRATION_DEFAULT;
+    }
+    return cacheExpirationTime;
+  }
+
+  protected int getMaxDistributionNumber() {
```
aleksei-pronichev, line 126: why so many protected methods? suggest to move to the private scope most of them

src/main/java/org/folio/inventory/support/KafkaConsumerVerticle.java
```diff
@@ -0,0 +1,144 @@
...
+        .build();
+      getLogger().info("kafkaConfig: {}", kafkaConfig);
+      return kafkaConfig;
+    });
+  }
+
+  protected HttpClient getHttpClient() {
+    return ofNullable(httpClient).orElseGet(() -> {
+      httpClient = vertx.createHttpClient();
+      return httpClient;
```
aleksei-pronichev, line 98: is it lazy init in most of fields? 
could we avoid it and don't have runtime problems?

src/main/java/org/folio/inventory/MarcHridSetConsumerVerticle.java
```diff
@@ -78,35 +40,8 @@ public void start(Promise<Void> startPromise) {
   }
 
   @Override
-  public void stop(Promise<Void> stopPromise) {
-    CompositeFuture.join(marcBibConsumerWrapper.stop(), marcHoldingsConsumerWrapper.stop())
-      .onComplete(ar -> stopPromise.complete());
-  }
-
-  private int getLoadLimit() {
-    return Integer.parseInt(System.getProperty("inventory.kafka.MarcBibInstanceHridSetConsumer.loadLimit", "5"));
```
pkjacob, line 87: The `getLoadLimit` method in `KafkaConsumerVerticle` will not work in this case.
Property value used here is `inventory.kafka.MarcBibInstanceHridSetConsumer.loadLimit`. However, the 
`getLoadLimit` in `KafkaConsumerVerticle` will resolve in a different value `inventory.kafka.MarcHridSetConsumer.loadLimit`

src/main/java/org/folio/inventory/support/KafkaConsumerVerticle.java
```diff
@@ -0,0 +1,144 @@
...
+      .loadLimit(getLoadLimit())
+      .globalLoadSensor(new GlobalLoadSensor())
+      .subscriptionDefinition(getSubscriptionDefinition(getKafkaConfig().getEnvId(), eventType))
+      .build();
+    consumerWrappers.add(kafkaConsumerWrapper);
+    return kafkaConsumerWrapper;
+  }
+
+  protected JsonObject getConfig() {
+    return ofNullable(config).orElseGet(() -> {
```
pkjacob, line 74: The parent class `AbstractVerticle` already has a `config()` method. Should we use that instead of creating this new method?

src/main/java/org/folio/inventory/handler/InstanceIngestEventHandler.java
PBobylev: Ignore this class for now, it will be implemented as a next step under the same PR

src/main/java/org/folio/inventory/MarcHridSetConsumerVerticle.java
PBobylev: Yes, for this class it doesn't follow common pattern. Thank you! Fixed

src/main/java/org/folio/inventory/support/KafkaConsumerVerticle.java
PBobylev: I think we should not use super method, because all 4 existed ConsumerVerticles used to have it as:
`JsonObject config = vertx.getOrCreateContext().config();`
and we basically do the same or return previously got config

src/main/java/org/folio/inventory/support/KafkaConsumerVerticle.java
PBobylev: because of inheritance. How could we make it private in case we can't extract corresponding logic to parent class?

PBobylev: discussed, agreed it's fine

src/main/java/org/folio/inventory/support/KafkaConsumerVerticle.java
PBobylev: discussed and agreed it's ok since it was before

pkjacob (approved):

PBobylev:
> please update news md file with a link to jira ticket to be able to track changes

Sure, thank you for the notice

PR #717, opened 2024-04-22
MODINV-1006 Avoid null values in the ElectronicAccess object

dmytrokrutii:
## Purpose
Do not send nulls when the values are empty in electronicAccess field

## Approach
Find null values inside electronicAccess object, and remove them

## Learning
[MODINV-1006](https://folio-org.atlassian.net/browse/MODINV-1006)

### DIFF
--- a/src/main/java/org/folio/inventory/support/JsonHelper.java
+++ b/src/main/java/org/folio/inventory/support/JsonHelper.java
@@ -11,6 +11,7 @@ import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/inventory/support/JsonHelper.java
```diff
@@ -53,4 +70,37 @@ private String readFile(InputStream is) throws IOException {
...
+      }
+    }
+
+    keysToRemove.forEach(itemObject::remove);
+    return itemObject;
+  }
+
+  private static List<Iterable<?>> toNotNullList(Collection<?> collection) {
+    return collection.stream()
+      .filter(item -> item != null)
```
Maksat-Galymzhan, line 93: You can use as an alternative lambda expression. For example,
filter(Objects::nonNull)

Maksat-Galymzhan (approved):

KaterynaSenchenko (approved):

JavokhirAbdullayev (approved):
please update news

src/main/java/org/folio/inventory/support/JsonHelper.java
```diff
@@ -41,6 +47,18 @@ public JsonObject getJsonFileAsJsonObject(String filePath) throws IOException {
     return new JsonObject(readFile(is));
   }
 
+  public static void putNotNullValues(JsonObject representation, String propertyName, Object obj) {
```
VRohach, line 50: @dmytrokrutii  Could you please add documentation for this public method?

RuslanLavrov (approved):

PR #721, opened 2024-05-14
MODINV-1023: Fix mod-inventory OOM issue.

VRohach:
Change creation MappingMetadataCache: 1 Instance for all verticles.


src/main/java/org/folio/inventory/dataimport/cache/MappingMetadataCache.java
```diff
@@ -107,4 +107,11 @@ private CompletableFuture<Optional<MappingMetadataDto>> loadMappingMetadata(Stri
       });
   }
 
+  public static synchronized MappingMetadataCache getInstance(Vertx vertx, HttpClient httpClient, long cacheExpirationTime) {
```
okolawole-ebsco, line 110: consider not accepting `cacheExpirationTime` as input. Once the singleton is created, getting an instance with a different cache expiration time would not have any effect on the already instantiated cache with an immutable cache expiration time.

VRohach: @okolawole-ebsco So where can I get the `cacheExpirationTime ` if not provided as input? I mean for using this method (which is used in the Verticles) `getCacheEnvVariable(config, "inventory.mapping-metadata-cache.expiration.time.seconds")` - in the `getInstance()`-method, it seems like I need to have a `config`. Maybe I missed smth...

okolawole-ebsco: you could pass the config instead. it looks like config is derived from `JsonObject config = vertx.getOrCreateContext().config();` so you could retrieve it wherever `Vertx vertx` is available.

VRohach: @okolawole-ebsco  I changed this code. Can you please review it? Also I encapsulated METADATA_EXPIRATION_TIME inside the cache, it seems like it would be better.

okolawole-ebsco: @VRohach I made another comment.

okolawole-ebsco (approved):

KaterynaSenchenko (approved):

Maksat-Galymzhan (approved):

JavokhirAbdullayev (approved):

Aliaksandr-Fedasiuk (approved):

PR #729, opened 2024-05-31
MODINV-986: InstanceIngress events consumption

PBobylev:


src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
```diff
@@ -0,0 +1,240 @@
...
+import org.folio.rest.jaxrs.model.InstanceIngressEvent;
+import org.folio.rest.jaxrs.model.ParsedRecord;
+import org.folio.rest.jaxrs.model.RawRecord;
+import org.folio.rest.jaxrs.model.Record;
+import org.folio.rest.jaxrs.model.Snapshot;
+
+public class CreateInstanceIngressEventHandler extends CreateInstanceEventHandler implements InstanceIngressEventHandler {
+
+  private static final Logger LOGGER = LogManager.getLogger(CreateInstanceIngressEventHandler.class);
+  private static final String BIBFRAME = " (bibframe)";
```
pkjacob, line 60: No space before brackets
`private static final String BIBFRAME = "(bibframe)";`

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
```diff
@@ -0,0 +1,240 @@
...
+            );
+          event.getEventPayload()
+            .withAdditionalProperty(MARC_BIBLIOGRAPHIC.value(), marcBibRecord);
+
+          LOGGER.info("Manipulating fields of a Record from InstanceIngressEvent with id '{}'", event.getId());
+          var mappingParameters = Json.decodeValue(mappingMetadata.getMappingParams(), MappingParameters.class);
+          AdditionalFieldsUtil.updateLatestTransactionDate(marcBibRecord, mappingParameters);
+          AdditionalFieldsUtil.move001To035(marcBibRecord);
+          AdditionalFieldsUtil.normalize035(marcBibRecord);
+          AdditionalFieldsUtil.addFieldToMarcRecord(marcBibRecord, TAG_035, SUBFIELD_A, event.getEventPayload().getSourceRecordIdentifier() + BIBFRAME);
```
pkjacob, line 123: Here we have to prepend the string BIBFRAME
```AdditionalFieldsUtil.addFieldToMarcRecord(marcBibRecord, TAG_035, SUBFIELD_A, BIBFRAME + event.getEventPayload().getSourceRecordIdentifier());```

File: src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java [hunk 3]
```diff
@@ -0,0 +1,240 @@
...
+    return Optional.empty();
+  }
+
+  private JsonObject prepareInstance(org.folio.Instance instance, SourceType sourceType) {
+    var instanceAsJson = JsonObject.mapFrom(instance);
+    instanceAsJson.put(SOURCE_KEY, sourceType.value());
+    if (isNull(instanceAsJson.getString(INSTANCE_TYPE_ID_KEY))) {
+      //q is that ok?
+      LOGGER.warn("No instance type provided in mapped instance with id '{}', setting 'unspecified'", instance.getId());
+      instanceAsJson.put(INSTANCE_TYPE_ID_KEY, "30fffe0e-e985-4144-b2e2-1e8179bdb41f");
```
pkjacob, line 173: What is this hardcoded ID? ( "30fffe0e-e985-4144-b2e2-1e8179bdb41f"). 
I guess this is the ID for type "unspecified". I guess this value will be different in different tenant databases and we cannot hardcode this in the code.

May be change the marc4ld library to always set and instancetype in the MARC record.

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
```diff
@@ -0,0 +1,240 @@
...
+    super(storage, precedingSucceedingTitlesHelper, mappingMetadataCache, idStorageService, null, httpClient);
+    this.context = context;
+    this.instanceCollection = storage.getInstanceCollection(context);
+  }
+
+  @Override
+  public CompletableFuture<InstanceIngressEvent> handle(InstanceIngressEvent event) {
+    try {
+      LOGGER.info("Processing InstanceIngressEvent with id '{}' for instance creation", event.getId());
+      if (isNull(event.getEventPayload()) || isNull(event.getEventPayload().getSourceRecordObject())) {
```
aleksei-pronichev, line 79: suggest to move it to methods to separate business from techical
`if (isEmptyEvent(event))` {
    return createError(event.getId)
}

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
```diff
@@ -0,0 +1,240 @@
...
+          return failIfErrors(uuidErrors, event.getId())
+            .orElseGet(() -> Future.succeededFuture(mappedInstance));
+        });
+    } catch (Exception e) {
+      return Future.failedFuture(e);
+    }
+  }
+
+  private static Optional<Future<Instance>> failIfErrors(List<String> errors, String eventId) {
+    if (!errors.isEmpty()) {
```
aleksei-pronichev, line 159: CollectionUtils.isNotEmpty
but bettor to inverse it
```java
if (isEmpty(errors) {
    return Optional.empty()
}
//some logic without bracket on the top level of line
```

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
aleksei-pronichev, line 173: move hard code to constant as minimum

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
```diff
@@ -0,0 +1,240 @@
...
+      ? Future.succeededFuture(instance)
+      : Future.failedFuture(format("Failed to add instance id '%s' to record with id '%s'", instance.getId(), srcRecord.getId()));
+  }
+
+  protected Future<Instance> saveRecordInSrsAndHandleResponse(InstanceIngressEvent event, Record srcRecord, Instance instance) {
+    LOGGER.info("Saving record in SRS and handling a response for an Instance with id '{}':", instance.getId());
+    Promise<Instance> promise = Promise.promise();
+    getSourceStorageRecordsClient(context.getOkapiLocation(), context.getToken(), context.getTenantId())
+      .postSourceStorageRecords(srcRecord)
+      .onComplete(ar -> {
```
aleksei-pronichev, line 214: suggest to move it to separate method

src/main/java/org/folio/inventory/instanceingress/handler/InstanceIngressUpdateEventHandler.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.concurrent.CompletableFuture;
+import org.folio.rest.jaxrs.model.InstanceIngressEvent;
+
+public class InstanceIngressUpdateEventHandler implements InstanceIngressEventHandler {
+
+  @Override
+  public CompletableFuture<InstanceIngressEvent> handle(InstanceIngressEvent instanceIngressEvent) {
+    // to be implemented in MODINV-1008
+    return null;
```
aleksei-pronichev, line 11: suggest to return empty action future instead of null

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
PBobylev: correct, fixed, thank you!

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
PBobylev: correct, fixed, thank you!

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
PBobylev: yes, have tested the case when our library puts 008 anyway - and it's successful case, we don't need this hardcode anymore. Removed

PBobylev: it was temporal, removed

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
PBobylev: fixed, thank you

src/main/java/org/folio/inventory/instanceingress/handler/InstanceIngressUpdateEventHandler.java
PBobylev: what's the matter? the next task in our team will be MODINV-1008.
But not a big deal, changed

src/main/java/org/folio/inventory/DataImportConsumerVerticle.java
```diff
@@ -97,89 +75,30 @@ public class DataImportConsumerVerticle extends AbstractVerticle {
...
-
-    List<Future<KafkaConsumerWrapper<String, String>>> futures = EVENT_TYPES.stream()
-      .map(eventType -> createKafkaConsumerWrapper(kafkaConfig, eventType, dataImportKafkaHandler))
-      .collect(Collectors.toList());
+    var futures = EVENT_TYPES.stream()
+      .map(eventType -> {
+        var consumerWrapper = super.createConsumer(eventType.value());
+        return consumerWrapper.start(dataImportKafkaHandler, ConsumerWrapperUtil.constructModuleName())
+          .map(consumerWrapper);
+      })
```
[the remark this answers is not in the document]
PBobylev: We are not here for changing existed logic or code refactoring more than needed. 
The code from the four existed XxxConsumerVerticle classes was extracted to parent AbstractVerticle, to be re-used then in a new InstanceIngressConsumerVerticle.

Here was a stream of EVENT_TYPES doing two things at once:
1) create a consumer
2) start it and return the future of this start operation (inside of createKafkaConsumerWrapper)

So we have exactly the same logic now, just the createConsumer operation is moved to parent class.

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
PBobylev: sure, changed, thank you!

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
PBobylev: It's a copy-paste of AbstractInstanceEventHandler#saveRecordInSrsAndHandleResponse, I suggest to leave it as is, to keep it similar and comparable with the original

src/main/java/org/folio/inventory/DataImportConsumerVerticle.java
```diff
@@ -97,87 +74,29 @@ public class DataImportConsumerVerticle extends AbstractVerticle {
...
+  protected Logger getLogger() {
+    return LOGGER;
   }
 
-  private int getLoadLimit() {
-    return Integer.parseInt(System.getProperty("inventory.kafka.DataImportConsumer.loadLimit", "5"));
-  }
-
-  private int getMaxDistributionNumber() {
-    return Integer.parseInt(System.getProperty("inventory.kafka.DataImportConsumerVerticle.maxDistributionNumber", "100"));
```
pkjacob, line 173: The `getMaxDistributionNumber` method in `KafkaConsumerVerticle` will not work here :( because the system property has "Verticle" in it. In `getConsumerProperty`, we remove "Verticle" :(

src/main/java/org/folio/inventory/support/KafkaConsumerVerticle.java
```diff
@@ -0,0 +1,165 @@
...
+  protected String getCacheEnvVariable(String variableName) {
+    var cacheExpirationTime = getConfig().getString(variableName);
+    if (isBlank(cacheExpirationTime)) {
+      cacheExpirationTime = CACHE_EXPIRATION_DEFAULT;
+    }
+    return cacheExpirationTime;
+  }
+
+  protected int getMaxDistributionNumber() {
+    return getConsumerProperty(MAX_DISTRIBUTION_NUMBER_TEMPLATE, MAX_DISTRIBUTION_NUMBER_DEFAULT);
```
pkjacob, line 128: I think calling getConsumerProperty  will cause prolmens :( 
In `getConsumerProperty`, we remove the word "Verticle" from the class name. However, if you look at the (now removed) getMaxDistributionNumber  methods in `DataImportConsumerVerticle` you can see that the system property used is `inventory.kafka.DataImportConsumer**Verticle**.maxDistributionNumber`

PBobylev: it's true... to make it safer, we'd better pass a specific value each time. Changed my approach. Thank you!

pkjacob (approved):

src/main/java/org/folio/inventory/instanceingress/handler/CreateInstanceIngressEventHandler.java
```diff
@@ -0,0 +1,246 @@
...
+    return isNull(event.getEventPayload())
+      || isNull(event.getEventPayload().getSourceRecordObject())
+      || isNull(event.getEventPayload().getSourceType());
+  }
+
+  private Future<org.folio.Instance> prepareAndExecuteMapping(MappingMetadataDto mappingMetadata,
+                                                              Record targetRecord,
+                                                              InstanceIngressEvent event,
+                                                              String instanceId) {
+    return postSnapshotInSrsAndHandleResponse(targetRecord.getId())
```
KaterynaSenchenko, line 120: Maybe it is better to create snapshot right before saving the record in SRS? This operation logically belongs there and has no relation to the mapping of Instance

PBobylev: That's correct, moved it to be executed just before saving to SRS

KaterynaSenchenko (approved):

PR #733, opened 2024-06-13
MODINV-977 Mark suppress from discovery for SRS holdings record

dmytrokrutii:
## Purpose
When MARC Holdings is edited in the “Inventory“ app and suppressed from discovery, its SRS record "suppressDiscovery" flag is not updated accordingly

## Approach
Check Holdings source - make a call for holdings-sources by sourceId. If source=MARC, call the following endpoint to suppress from discovery the underlying MARC Holdings:
`/source-storage/records/{id}/suppress-from-discovery`

## Related issues
[MODINV-977](https://folio-org.atlassian.net/browse/MODINV-977)


Maksat-Galymzhan (commented):
Please, update News.md file

Maksat-Galymzhan (approved):

RuslanLavrov (approved):

PR #746, opened 2024-07-26
MODINV 1049 - reorder 035 oclc after normalization

JavokhirAbdullayev:
### Purpose
[MODINV-1049](https://folio-org.atlassian.net/browse/MODINV-1049)

### Approach
get index of oclc 035 field and set newly normalized field


src/main/java/org/folio/inventory/dataimport/util/AdditionalFieldsUtil.java
```diff
@@ -582,19 +583,16 @@ private static boolean removeFirstFoundFieldByName(org.marc4j.marc.Record marcRe
   private static void replaceOclc035FieldWithNormalizedData(org.marc4j.marc.Record marcRecord,
                                               DataField dataField) {
     var variableFields = marcRecord.getVariableFields(TAG_035);
+    var index = new AtomicInteger();
```
Aliaksandr-Fedasiuk, line 586: Are you sure we need an AtomicInteger here?

RomanChernetskyi (approved):

RuslanLavrov (approved):

KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/dataimport/util/AdditionalFieldsUtil.java
JavokhirAbdullayev: yes, because inside of for each loop we have to use either final variable or atomic variable

PR #755, opened 2024-08-23
[MODINVSTOR-1243] Implement endpoint to retrieve items from multiple tenants

Saba-Zedginidze-EPAM:
## Purpose
[[MODINVSTOR-1243] Implement endpoint to retrieve items from multiple tenants
](https://folio-org.atlassian.net/browse/MODINVSTOR-1243)

## Approach
- Add new types for receiving request body
- Add new endpoint
- Add endpoint handle to fetch items from different tenants

## Screenshots
Verified and left comment on JIRA ticket: https://folio-org.atlassian.net/browse/MODINVSTOR-1243?focusedCommentId=226464

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -366,7 +366,7 @@
           ]
         },
         {
-          "methods": ["GET"],
+          "methods": ["POST"],
           "pathPattern": "/inventory/tenant-items",
           "permissionsRequired": ["inventory.items.collection.get"],
           "modulePermissions": [
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -367,6 +367,14 @@
         },
         {
           "methods": ["GET"],
+          "pathPattern": "/inventory/tenant-items",
+          "permissionsRequired": ["inventory.items.collection.get"],
+          "modulePermissions": [
+            "inventory-storage.items.collection.get"
+          ]
+        },
+        {
+          "methods": ["GET"],
           "pathPattern": "/inventory/instances/{id}",
           "permissionsRequired": ["inventory.instances.item.get"],
           "modulePermissions": [
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -368,7 +368,7 @@
         {
           "methods": ["POST"],
           "pathPattern": "/inventory/tenant-items",
-          "permissionsRequired": ["inventory.items.collection.get"],
+          "permissionsRequired": ["inventory.tenant-items.collection.get"],
           "modulePermissions": [
             "inventory-storage.items.collection.get"
           ]
@@ -690,6 +690,11 @@
       "description": "Get item collection"
     },
     {
+      "permissionName": "inventory.tenant-items.collection.get",
+      "displayName": "Inventory - get item collection from multiple tenants",
+      "description": "Get item collection from multiple tenants"
+    },
+    {
       "permissionName": "inventory.items.collection.delete",
       "displayName": "Inventory - delete entire item collection",
       "description": "Delete entire item collection"
@@ -836,6 +841,7 @@
       "description": "Entire set of permissions needed to use the inventory",
       "subPermissions": [
         "inventory.items.collection.get",
+        "inventory.tenant-items.collection.get",
         "inventory.items.item.get",
         "inventory.items.item.post",
         "inventory.items.item.put",
--- a/src/main/java/org/folio/inventory/resources/TenantItems.java
+++ b/src/main/java/org/folio/inventory/resources/TenantItems.java
@@ -98,7 +98,7 @@ public class TenantItems {
   }
 
   private List<JsonObject> getItemsWithTenantId(String tenantId, Response response) {
-    if (response.getStatusCode() == 200 && response.hasBody()) {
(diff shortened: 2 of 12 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -365,6 +365,14 @@
             "inventory-storage.instances.item.get"
           ]
         },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/inventory/tenant-items",
+          "permissionsRequired": ["inventory.items.collection.get"],
```
SerhiiNosko, line 371: we need to create new permission for each endpoint instead of reusing existing. Also as I know Eureka platform does not support assigning the same permission for multiple endpoints

SerhiiNosko: resolved

SerhiiNosko (approved):

BKadirkhodjaev (approved):

src/main/java/org/folio/inventory/resources/TenantItems.java
```diff
@@ -0,0 +1,137 @@
...
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.ext.web.handler.BodyHandler;
+
+public class TenantItems {
+
+  private static final Logger log = LogManager.getLogger(MethodHandles.lookup().lookupClass());
```
Aliaksandr-Fedasiuk, line 39: Could you, please, use uppercase for the "log" variable?

src/main/java/org/folio/inventory/resources/TenantItems.java
```diff
@@ -0,0 +1,137 @@
...
+
+import io.vertx.core.http.HttpClient;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.ext.web.handler.BodyHandler;
+
+public class TenantItems {
```
Aliaksandr-Fedasiuk, line 37: Could you add Javadoc to the class?

src/main/java/org/folio/inventory/resources/TenantItems.java
```diff
@@ -0,0 +1,137 @@
...
+    var getByIdsQuery = multipleRecordsCqlQuery(itemIds);
+    var itemsFetched = new CompletableFuture<Response>();
+    itemsStorageClient.getAll(getByIdsQuery, itemsFetched::complete);
+
+    return itemsFetched.thenApplyAsync(response ->
+      getItemsWithTenantId(tenantId, response));
+  }
+
+  private List<JsonObject> getItemsWithTenantId(String tenantId, Response response) {
+    if (response.getStatusCode() != 200 || !response.hasBody()) {
```
Aliaksandr-Fedasiuk, line 101: Please use an already-defined constant instead of a “magic number”.

src/main/java/org/folio/inventory/resources/TenantItems.java
SerhiiNosko: changed

src/main/java/org/folio/inventory/resources/TenantItems.java
SerhiiNosko: added, thanks

src/main/java/org/folio/inventory/resources/TenantItems.java
SerhiiNosko: used HttpStatus enum

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/inventory/resources/TenantItems.java
```diff
@@ -0,0 +1,142 @@
...
+    return itemsFetched.thenApplyAsync(response ->
+      getItemsWithTenantId(tenantId, response));
+  }
+
+  private List<JsonObject> getItemsWithTenantId(String tenantId, Response response) {
+    if (response.getStatusCode() != HttpStatus.SC_OK || !response.hasBody()) {
+      return List.of();
+    }
+    return JsonArrayHelper.toList(response.getJson(), ITEMS_FIELD).stream()
+      .map(item -> item.put(TENANT_ID_FIELD, tenantId))
```
RuslanLavrov, line 110: could you say why "tenantId" field is set to the item json? Looks like it does not comply with the response definition because the item schema does not contain "tenantId" field.

SerhiiNosko: In screen from testing I see that response object has tenantId populated:
image
Maybe Item allows to add properties dynamically

RuslanLavrov: the field "tenantId" successfully has been added to response because in the code we set it to JsonObject representing item and that allows to add any property

SerhiiNosko: do you think we can improve this in some way, for example by adding tenantId to Item schema explicitly?

RuslanLavrov: I suppose yes, or maybe to add a new response schema for this endpoint that would refer to the item schema

PR #757, opened 2024-08-28
MODINV-1069 Fix DataImportConsumerVerticleTest in mod-inventory and Fix NPE in HoldingsItemMatcher

JavokhirAbdullayev:
### Purpose
fix NPE in HoldingsItemMatcher [MODINV-1069](https://folio-org.atlassian.net/browse/MODINV-1069)

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/HoldingLoader.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/HoldingLoader.java
@@ -4,13 +4,11 @@ import static java.lang.String.format;
 import static org.apache.commons.lang3.StringUtils.EMPTY;
 import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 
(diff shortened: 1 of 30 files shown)


KaterynaSenchenko (approved):

src/main/java/org/folio/inventory/dataimport/handlers/matching/loaders/HoldingLoader.java
```diff
@@ -73,6 +72,12 @@ protected String addCqlSubMatchCondition(DataImportEventPayload eventPayload) {
     return cqlSubMatch;
   }
 
+  private static boolean checkIfValueIsNullOrEmpty(String value) {
```
Aliaksandr-Fedasiuk, line 75: @JavokhirAbdullayev , could you use another name for this method like "isNullOrEmpty"?

RomanChernetskyi (approved):

PR #758, opened 2024-08-28
feat(instance-date-type-fields): add new date type fields to Instance schema

SvitlanaKovalova1:
### Purpose: 
[MODINV-1067](https://folio-org.atlassian.net/browse/MODINV-1067) Currently in the Instance schema, there is a field for publication date. However, this data originates from an uncontrolled string in variable fields of the MARC record. The dates in the 008 are more consistently formatted, and are more useful for certain materials like serials. Previously, we implemented sortable properties (publicationPeriod.start/publicationPeriod.end) that are parsed from the publication date field. We need to break this relationship, and instead make the source of that data the elements from the 008. The 008 structure allows for a single position to indicate the type of date that will be represented in the following positions. 

### Approach
1. Add three new properties to the Instance schema:
- dateTypeId
- date1
- date2
2. Update GET /inventory/config/instances/blocked-fields endpoint with new fields


marcjohnson-kint:
@SvitlanaKovalova1 Thank you for the invite to review

I am no longer actively involved in inventory development within FOLIO

What feedback would you like from me regarding this change?

psmagin (approved):

viacheslavkol (approved):

PR #783, opened 2024-10-28
MODINV-1100: Retry instance update on optimistic locking error

mukhiddin-yusuf:
**Purpose:**

Optimistic Locking error is being observed while concurrently handling two or more Marc Bib update events (from the topic srs.marc-bib) which are being consumed at the same time and they are related to the update of the one/same Marc Bib record. Separate instances of consumers tries to handle and then update the instance record by calling the PUT /instance-storage/{instanceID} but receive the OL error.

**Approach:**

- use the same retry approach utilized in other consumers to handle the OL error


src/main/java/org/folio/inventory/dataimport/consumers/MarcBibUpdateKafkaHandler.java
```diff
@@ -49,6 +50,8 @@ public class MarcBibUpdateKafkaHandler implements AsyncRecordHandler<String, Str
   private static final AtomicLong INDEXER = new AtomicLong();
   private static final String MAPPING_RULES_KEY = "MAPPING_RULES";
   private static final String MAPPING_PARAMS_KEY = "MAPPING_PARAMS";
+  private static final String CURRENT_RETRY_NUMBER = "CURRENT_RETRY_NUMBER";
+  private static final int MAX_RETRIES_COUNT = Integer.parseInt(System.getenv().getOrDefault("inventory.di.ol.retry.number", "1"));
```
psmagin, line 54: I think it should be separate var.

PR #794, opened 2024-12-24
MODDATAIMP-1145 (2) Import of MARC records fails with "Invalid token" error. 

JavokhirAbdullayev:
### Purpose 
[MODDATAIMP-1145](https://folio-org.atlassian.net/browse/MODDATAIMP-1145) Import of MARC records fails with "Invalid token" error.

### Todo: 
investigate why userId is being set in payload not in header when _DI_SRS_MARC_AUTHORITY_RECORD_MODIFIED_READY_FOR_POST_PROCESSING_ event is received


Aliaksandr-Fedasiuk (approved):

KaterynaSenchenko (approved):

VRohach (approved):

RuslanLavrov (approved):

src/main/java/org/folio/inventory/dataimport/consumers/DataImportKafkaHandler.java
```diff
@@ -146,7 +146,8 @@ public Future<String> handle(KafkaConsumerRecord<String, String> record) {
       eventPayload.getContext().put(CHUNK_ID_HEADER, chunkId);
       eventPayload.getContext().put(USER_ID_HEADER, userId);
 
-      Context context = EventHandlingUtil.constructContext(eventPayload.getTenant(), eventPayload.getToken(), eventPayload.getOkapiUrl());
+      Context context = EventHandlingUtil.constructContext(eventPayload.getTenant(), eventPayload.getToken(), eventPayload.getOkapiUrl(),
```
RuslanLavrov, line 149: please pass userId during context creation in the OrdersPreloaderHelper as well
https://github.com/folio-org/mod-inventory/blob/cb2b9368c07872adc0c7d5d0048fd99d714481f6/src/main/java/org/folio/inventory/dataimport/handlers/matching/preloaders/OrdersPreloaderHelper.java#L54

JavokhirAbdullayev: thank you Ruslan, nice catch!

JavokhirAbdullayev: but does mod-orders require this header as well? i guess only mod-inventory-storage requires it?

RuslanLavrov: here in this class we need to pass userId just to avoid "invalid token" error during request that is performed in this class. Thank you Javokhir!

PR #798, opened 2025-01-06
[MODINV-1136] Set discoverySuppress instance attribute false by default during mapping from json

RomanChernetskyi:
### Purpose 
The "discoverySuppress" instance attribute is not created in database when instance created by default mapping profile

### Approach
Set discoverySuppress instance attribute false by default during mapping from json

#### Jira
https://folio-org.atlassian.net/browse/MODINV-1136


JavokhirAbdullayev (approved):

JavokhirAbdullayev (approved):
Roman, please set jira link in description or in news

PR #812, opened 2025-02-10
[MODINV-1140] Update MARC-Instance mapping to account for Leader 05 value

RomanChernetskyi:
### Purpose
Update MARC-Instance mapping to account for Leader 05 value
### Approach
If 05 LEADER is 'd’ mark instance and underlying MARC record as deleted during create/update operation
#### Jira
https://folio-org.atlassian.net/browse/MODINV-1140


src/main/java/org/folio/inventory/dataimport/handlers/actions/AbstractInstanceEventHandler.java
```diff
@@ -223,7 +226,21 @@ private Record encodeParsedRecordContent(Record srcRecord) {
     return srcRecord;
   }
 
-  protected void setSuppressFormDiscovery(Record srcRecord, boolean suppressFromDiscovery) {
+  protected void markInstanceAndRecordAsDeletedIfNeeded(Instance instance, Record srsRecord) {
+    Optional<Character> leaderStatus = ParsedRecordUtil.getLeaderStatus(srsRecord.getParsedRecord());
```
Aliaksandr-Fedasiuk, line 230: Could you add more debug messages? Thanks.

Aliaksandr-Fedasiuk (approved):

RuslanLavrov (approved):

PR #813, opened 2025-02-12
MODINV-1138: Adjust mark-deleted endpoint behavior to set the "deleted"

AndreiBordak:


pkjacob (approved):

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -343,8 +343,9 @@ private void softDelete(RoutingContext routingContext) {
   private void updateVisibility(Instance instance, RoutingContext routingContext,  InstanceCollection instanceCollection) {
     instance.setDiscoverySuppress(true);
     instance.setStaffSuppress(true);
+    instance.setDeleted(true);
```
pkjacob, line 346: Add a line in News.md file for this change?

AndreiBordak: done.

PR #817, opened 2025-02-24
[MODSOURCE-867] Allow update of instance without related Marc record

RomanChernetskyi:
### Purpose
Overlay of Instance does not fix record without ACTUAL state
### Approach
Create record during instance update if it does not exist

#### Jira
https://folio-org.atlassian.net/browse/MODSOURCE-867

### DIFF
--- a/src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
+++ b/src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
@@ -40,6 +40,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/inventory/dataimport/handlers/actions/ReplaceInstanceEventHandler.java
```diff
@@ -238,6 +238,11 @@ private void processInstanceUpdate(DataImportEventPayload dataImportEventPayload
               AdditionalFieldsUtil.remove035FieldWhenRecordContainsHrId(targetRecord);
 
               setSuppressFromDiscovery(targetRecord, instance.getDiscoverySuppress());
+              if (targetRecord.getMatchedId() == null) {
```
Aliaksandr-Fedasiuk, line 241: This method does not have enough LOGs. Could you, please, fix this?

Aliaksandr-Fedasiuk (approved):

KaterynaSenchenko (approved):

RuslanLavrov (approved):

PR #818, opened 2025-02-26
MODINV-1139: Make changes to PUT instances endpoint to update MARC with "deleted" markers

AndreiBordak:


src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -211,6 +212,22 @@ private void updateSuppressFromDiscoveryFlag(WebContext wContext, Instance updat
     }
   }
 
+  private void unDeleteInstance(SourceStorageRecordsClient srsClient, Instance updatedInstance) {
```
pkjacob, line 215: Minor - I think it is better to rename this method to `unDeleteSourceStorageRecord` to match with the existing method `deleteSourceStorageRecord`. Also, better to move this method under `deleteSourceStorageRecord` so that `unDetele...` method is after `delete...` method

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -261,13 +278,23 @@ private void updateInstance(Instance instance, RoutingContext rContext, WebConte
...
       FailureResponseConsumer.serverError(rContext.response()));
   }
 
+  private void updateVisibilityFlags(Instance instance, WebContext wContext) {
+    if (isInstanceControlledByRecord(instance)) {
+      var srsClient = getSourceStorageRecordsClient(wContext);
+      if (isTrue(instance.getDeleted())) {
+        deleteSourceStorageRecord(wContext, instance.getId());
+      } else if (isFalse(instance.getDeleted())) {
+        updateSuppressFromDiscoveryFlag(srsClient, instance);
```
pkjacob, line 292: Call to `updateSuppressFromDiscoveryFlag` should be made outside of `if` and `else` blocks.

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -261,13 +278,23 @@ private void updateInstance(Instance instance, RoutingContext rContext, WebConte
...
-          updateSuppressFromDiscoveryFlag(wContext, instance);
-        }
+        updateVisibilityFlags(instance, wContext);
       },
       FailureResponseConsumer.serverError(rContext.response()));
   }
 
+  private void updateVisibilityFlags(Instance instance, WebContext wContext) {
+    if (isInstanceControlledByRecord(instance)) {
+      var srsClient = getSourceStorageRecordsClient(wContext);
```
pkjacob, line 288: `srsClient` is retrieved a this line. However, it is not used when `if` block is executed. (In `if` block, we use `wContext` instead of `srsClient`). 
So, either do one of the following
Update deleteSourceStorageRecord to accept `srsClient` instead of `wContext`
Move this line to the `else` block.

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -261,13 +278,23 @@ private void updateInstance(Instance instance, RoutingContext rContext, WebConte
...
           });
-        if (isInstanceControlledByRecord(instance)) {
-          updateSuppressFromDiscoveryFlag(wContext, instance);
-        }
+        updateVisibilityFlags(instance, wContext);
       },
       FailureResponseConsumer.serverError(rContext.response()));
   }
 
+  private void updateVisibilityFlags(Instance instance, WebContext wContext) {
```
pkjacob, line 286: The changes made in this PR will lead to the unintended emission of Kafka events. I recall us discussing this scenario in one of our daily stand-ups.

Scenario:
A user opens an existing, non-deleted instance for editing, makes some modifications (without changing the "delete" flag), and saves the changes. This triggers `updateVisibilityFlags`. At line 293, an "un-delete" API call is made to SRS, even though the instance is already in an undeleted state. As a result, SRS processes the request and emits a Kafka event.

pkjacob, line 286: There is already an "updateVisibility" method in this class. It will be very confusing to have another method "updateVisibilityFlags" with very similar name.

pkjacob:
NEWS.md file should be updated.

src/test/java/api/InstancesApiExamples.java
```diff
@@ -413,8 +413,8 @@ public void canUpdateAnExistingInstance()
       .put("natureOfContentTermIds",
         new JsonArray().add(ApiTestSuite.getAudiobookNatureOfContentTermId()))
       .put("subjects", new JsonArray().add(
-        new Subject(null,  null, sourceId, typeId)
-      ));
+        new Subject(null,  null, sourceId, typeId)))
```
pkjacob, line 416: Is this test enough for testing all the changes we made in this PR? Is there a way to use mocks to ensure the following? 
`srsClient.postSourceStorageRecordsUnDeleteById` or `srsClient.deleteSourceStorageRecordsById` method is not invoked when PUT /instance/{id} call is made with no change in delete flag.

If there is no way to use mocks, another way is to do this in karate test.

ie, make calls to PUT /instance/{id} API with no change to delete flag.
Then get underlying marc record by pulling GET /id of source-recrod-storage and ensure that the value of MARC field 005 is not updated.

src/main/java/org/folio/inventory/resources/Instances.java
AndreiBordak: suppressFromDiscovery can be updated only if 'deleted' is false, this requirement is reflected in the UI:
image
Should we move it out of the if-else block anyway?

pkjacob: Got it. Thank You.

src/main/java/org/folio/inventory/resources/Instances.java
AndreiBordak: done.

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -261,13 +278,23 @@ private void updateInstance(Instance instance, RoutingContext rContext, WebConte
...
+    if (isInstanceControlledByRecord(instance)) {
+      var srsClient = getSourceStorageRecordsClient(wContext);
+      if (isTrue(instance.getDeleted())) {
+        deleteSourceStorageRecord(wContext, instance.getId());
```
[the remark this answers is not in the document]
AndreiBordak: done.

src/main/java/org/folio/inventory/resources/Instances.java
AndreiBordak: done.

src/main/java/org/folio/inventory/resources/Instances.java
AndreiBordak: done.

AndreiBordak: Done.

src/test/java/api/InstancesApiExamples.java
AndreiBordak: Will do this in Karate test.

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -261,13 +282,28 @@ private void updateInstance(Instance instance, RoutingContext rContext, WebConte
...
+        updateVisibilityFlagsInSrs(existingInstance, updatedInstance, wContext);
       },
       FailureResponseConsumer.serverError(rContext.response()));
   }
 
+  private void updateVisibilityFlagsInSrs(Instance existing, Instance updated, WebContext wContext) {
+    if (!isInstanceControlledByRecord(updated)) {
+      return;
+    }
+    if (isTrue(updated.getDeleted()) && existing.getDeleted() != updated.getDeleted()) {
```
pkjacob, line 294: As `getDeleted()` returns Boolean object, here we have to use `ObjectUtils.notEqual(existing.getDeleted(), updated.getDeleted())` instead of `existing.getDeleted() != updated.getDeleted()`

Same change in line 298 & 301

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -261,13 +282,28 @@ private void updateInstance(Instance instance, RoutingContext rContext, WebConte
...
+      deleteSourceStorageRecord(wContext, updated.getId());
+    } else if (isFalse(updated.getDeleted())) {
+      var srsClient = getSourceStorageRecordsClient(wContext);
+      if (existing.getDiscoverySuppress() != updated.getDiscoverySuppress()) {
```
pkjacob, line 298: Use `ObjectUtils.notEqual` instead of `!=`

src/main/java/org/folio/inventory/resources/Instances.java
```diff
@@ -261,13 +282,28 @@ private void updateInstance(Instance instance, RoutingContext rContext, WebConte
...
+        updateSuppressFromDiscoveryFlag(srsClient, updated);
+      }
+      if (existing.getDeleted() != updated.getDeleted()) {
```
pkjacob, line 301: Use `ObjectUtils.notEqual` instead of `!=`

src/main/java/org/folio/inventory/resources/Instances.java
AndreiBordak: done.

src/main/java/org/folio/inventory/resources/Instances.java
AndreiBordak: done.

src/main/java/org/folio/inventory/resources/Instances.java
AndreiBordak: done.

RomanChernetskyi (approved):

KaterynaSenchenko (approved):

pkjacob:
Build fails. I think you have to update pom.mxl to use the latest version of srs client.

PR #865, opened 2025-09-16
MODINV-1280: Sort bound-with items in their createdDate order

julianladisch:
https://folio-org.atlassian.net/browse/MODINV-1280


nielserik (approved):
Looks good to me.

PR #932, opened 2026-07-31
MODINV-1406: Add customFields to item

elsenhans:
[MODINV-1406](https://folio-org.atlassian.net/browse/MODINV-1406)
**This PR is connected to https://github.com/folio-org/mod-inventory-storage/pull/1251 (which is blocked)**

### Purpose
**Background** 
`mod-inventory-storage` already integrates `folio-custom-fields`: it defines/persists custom field definitions via its own `/custom-fields` interface and accepts a `customFields` value on the stored item record. `mod-inventory` sits in front of storage with its own copy of the item schema and domain model, so without this change any `customFields` value sent by a client would be silently dropped (schema has `additionalProperties: false`) and never returned in responses.

**Acceptance Criteria**
- POSTing/PUTting an item with `customFields` persists and returns that value unchanged via mod-inventory's API
- Items without `customFields` omit the property from the response (no empty/null noise)
- No changes needed to custom field definition management — that remains owned by `mod-inventory-storage`

### Approach
- Add `customFields` property to `ramls/item.json` (referencing a new `ramls/customFields.json` schema, generic object)
- Add `customFields` field + getter/`withCustomFields` to `Item.java`, and ensure it's carried over in `copyWithNewId`/similar copy methods
- Update `ItemRepresentation.java` to include `customFields` in the API response when present
- Update `ItemUtil.java` mapping in both directions: request JSON → Item, Item → stored representation, stored representation → Item
- Unit tests covering: field present/absent in response, round-trip through stored/request representations

### Changes Checklist
- [ ] **API Changes**: Document any API paths, methods, request or response bodies changed, added, or removed.
- [ ] **Database Schema Changes**: Indicate any database schema changes and their impact. Confirm that migration scripts were created.
- [ ] **Interface Version Changes**: Indicate any changes to interface versions.
- [ ] **Interface Dependencies**: Document added or removed dependencies.
- [ ] **Permissions**: Document any changes to permissions.
- [ ] **Logging**: Confirm that logging is appropriately handled.
- [ ] **Unit Testing**: Confirm that changed classes were covered by unit tests.
- [ ] **Integration Testing**: Confirm that changed logic was covered by integration tests.
- [ ] **Manual Testing**: Confirm that changes were tested on local or dev environment.
- [ ] **NEWS**: Confirm that the NEWS file is updated with relevant information about the changes made in this pull request.

### Related Issues
- https://folio-org.atlassian.net/browse/UXPROD-5370
- https://folio-org.atlassian.net/browse/MODINVSTOR-1445
- https://folio-org.atlassian.net/browse/MODINVSTOR-1446
- https://folio-org.atlassian.net/browse/UIIN-3343

### Learning and Resources (if applicable)
_Discuss any research conducted during the development of this pull request. Include links to relevant blog posts, patterns, libraries, or addons that were used to solve the problem._

### Screenshots (if applicable)
_If this pull request involves any visual changes or new features, consider including screenshots or GIFs to illustrate the changes._


ramls/item.json
```diff
@@ -544,6 +544,11 @@
           "readonly": true
         }
       }
+    },
+    "customFields": {
+      "description": "Object that contains custom fields",
+      "type": "object",
+      "$ref": "customFields.json"
```
charismatic-claire, line 551: Just because I'm exploring this: We add a new field to `item` that is `customFields` and we reference it here. And we do this here, since the `ramls` folder basically contains the endpoint definitions, right?

src/main/java/org/folio/inventory/domain/items/Item.java
```diff
@@ -110,6 +111,7 @@ public class Item {
 
   private boolean isBoundWith = false;
   private JsonArray boundWithTitles = null;
+  private JsonObject customFields;
```
charismatic-claire, line 114: There is no right or wrong, I'm just asking questions... Is there any particular reason why we are making this a rather generic `JsonObject` and don't create a `CustomFields` class that we reference here? 
 
Uh, maybe because `customFields` can be any object with arbitrary key value pairs, see `customFileds.json` definition?

src/test/java/org/folio/inventory/resources/ItemRepresentationTest.java
```diff
@@ -50,4 +50,24 @@ public void jsonWithoutNullValuesOnly() {
     assertThat(electronicAccessObject.getJsonObject(0).getValue("uri"), is(testValue));
     assertThat(emptyElectronicAccessObject, is(new JsonArray()));
   }
+
+  @Test
+  public void jsonContainsCustomFieldsWhenPresent() {
+    var customFields = new JsonObject().put("department", "Acquisitions");
+    var item = new Item(UUID.randomUUID().toString(), "123", null,
+        new Status(ItemStatusName.AVAILABLE), null, null, null)
+      .withCustomFields(customFields);
```
charismatic-claire, line 59: That is what makes the difference, right? That is why it contains custom fields in this test, but not in the other, right?

src/test/java/org/folio/inventory/support/ItemUtilTest.java
```diff
@@ -142,4 +142,53 @@ public void shouldNotFailWhenReadOnlyFieldsAreMissing() {
...
+    // given
+    JsonObject customFields = new JsonObject().put("department", "Acquisitions");
+    Item itemWithCustomFields = new Item(UUID.randomUUID().toString(), "2", UUID.randomUUID().toString(),
+      new Status(AVAILABLE), UUID.randomUUID().toString(), UUID.randomUUID().toString(), null)
+      .withCustomFields(customFields);
+    Item itemWithoutCustomFields = new Item(UUID.randomUUID().toString(), "2", UUID.randomUUID().toString(),
+      new Status(AVAILABLE), UUID.randomUUID().toString(), UUID.randomUUID().toString(), null);
+
+    // then
+    assertEquals(customFields, ItemUtil.toStoredItemRepresentation(itemWithCustomFields).getJsonObject("customFields"));
```
charismatic-claire, line 157: Just because I thought about it... We do not need to add the "custom fields" object to any database schema specifically, since `ItemUtil.toStoredItemRepresentation` takes care of it, and when the module actually persists something, the database column is created automatically? Or would we need to change something in `mod-inventory-storage` too? I'm just curious.

charismatic-claire (commented):
I basically approve. I just have some comprehensive questions and one remark about persistence. Cool. Thanks for letting me read this :)

src/test/java/org/folio/inventory/support/ItemUtilTest.java
elsenhans: The DB tables/columns are created in `mod-inventory-storage`:
> This PR is connected to https://github.com/folio-org/mod-inventory-storage/pull/1251 (which is blocked)

ramls/item.json
elsenhans: `item.json` is not itself an endpoint definition — it's a JSON Schema that describes the shape/data model of an **Item resource**. Since we like to save custom fields in an item , we need to extend that schema with a `customFields` property.

src/main/java/org/folio/inventory/domain/items/Item.java
elsenhans: The set of custom field names isn't known at compile time, since `customFields.json` allows `additionalProperties: true`.

src/test/java/org/folio/inventory/resources/ItemRepresentationTest.java
elsenhans: Right! The only difference between the two tests is this one method call:
`.withCustomFields(customFields)`
In the "present" test, it's called. In the "absent" test this call is missing.

alb3rtino (changes requested):
`inventory` interface version should be bumped as the item schema has been updated. Also require a custom-fields supporting `item-storage` interface, see https://github.com/folio-org/mod-inventory-storage/pull/1251#pullrequestreview-4842961899.

src/test/java/org/folio/inventory/support/ItemUtilTest.java
charismatic-claire: Ah alright. Thanks for explaining :)

charismatic-claire (approved):

elsenhans:
@alb3rtino Thanks for the hint!
`inventory` and `item-storage` interface versions has been increased.

